PR #2, opened 2021-12-05
Add note to readme about cmake version to build

StefanH-AT:
This is just to make it more obvious to check the cmake version. RHEL still  likes to prefer cmake 2


MickVermeulen (approved):
LGTM

PR #3, opened 2021-12-05
Note that the zlib -dev package is needed

luxaritas:


aronwk-aaron:
for ubuntu, specifically you need to install `zlib1g-dev`

Hpmason:
I believe g++ is needed as well on top of gcc

XenoTrixx:
> for ubuntu, specifically you need to install `zlib1g-dev`

can confirm zlib1g-dev is needed on ubuntu server 21.10

DarwinAnim8or (approved):

PR #10, opened 2021-12-05
Update Diagnostics.cpp to fix compiler error on MSVC

Jettford:
MSVC is funky and wants Windows.h included before Dbghelp.h or it causes some compiler errors


Jettford:
Fixed in PR #15

PR #11, opened 2021-12-05
change default network version to unmodded client

yuwui:
as most people will not have access to the DLU client it may be better to have the default net version on 171022 to avoid confusion


MickVermeulen:
In addition to this it's probably good to then remove the instructions to change it from the README (https://github.com/DarkflameUniverse/DarkflameServer/blob/34f1d4002b4701d28bfb650322977c62277cdc81/README.md?plain=1#L43)

yuwui:
will do

PR #14, opened 2021-12-05
Set `build.sh` as executable & update `README.md` to acknowledge  `build.sh`'s existence

maxdelayer:
Updated the index of `build.sh` so it is automatically made executable. Given existing documentation says to effectively do all of the steps in `build.sh` without actually *saying* that you can use `build.sh` to simplify this process, I updated the readme as well to explain that explicitly.


RedstoneParadox:
Before this gets merged, I'd just like to point out that `build.sh` only works properly on linux; the build command is `cmake --build .` on windows while build.sh just have `build`; should probably be mentioned in the readme.

luxaritas:
On that note, couldn't --build be used on Linux too for consistency?

averysumner:
Yes, it should always be preferred. It will always use the correct build command for the generator you used.

leolion3:
Cloud add my automated build file aswell if desired https://gist.github.com/leolion3/ccf654ab60c8e110c65ef948da6af461#file-buildserver-sh

MickVermeulen (changes requested):
What is the change that you did to build.sh? Would it be possible to revert that?

leolion3:
I changed the location in the script where mysql is started as before it was being called after executing the MySQL commands which would've not worked. Change history: https://gist.github.com/leolion3/ccf654ab60c8e110c65ef948da6af461/revisions#diff-64290f8d93530c0cfac2b096b5f8a2f47f436c4aea958cfffa7cd4d929755498

maxdelayer:
> What is the change that you did to build.sh? Would it be possible to revert that?

In this pull request I simply made build.sh executable by default, which means that users won't need to make it executable to run it once they've cloned the repo on a linux server. You can revert this by using [git update-index --chmod=-x](https://www.git-scm.com/docs/git-update-index#Documentation/git-update-index.txt---chmod-x)

MickVermeulen:
Ah right, sounds good then.

PR #36, opened 2021-12-06
Add database in Docker for easy deployment

jgkawell:
# Description

This adds an effortless way to deploy the needed database for the server. It wraps up MariaDB in Docker and reads from a `.env` file for configuration. This way you only need Docker available to be able to run the database.


BloodyHellcat:
Can you add a link to the docker container ok the docker hub?
This would prevent setting up a wrong container.

Iterator256:
I think it might be a great idea if the entire setup of DarkflameServer could be deployed with docker-compose. Sadly I don't know a lot of docker but I think it might be possible to create a docker-compose file that starts mariadb, AccountManager, the HTTP server for Brick-By-Brick as well as the main DarkflameServer. Since the DarkflameServer has to be built by the user (for legal reasons if I see that correctly) one would probably have to include a Dockerfile for DarkflameServer which locally builds the docker image and the user would have to supply a `client.rar` file. 

Most of the setup from the README could then be written as a script/in the Dockerfile. 

If this is possible that could allow even very non-technical users to setup the server easily, the instructions would come down to:
1. Setup a UNIX-Like OS with docker on it.
2. Clone the repo
3. Put your `client.rar` file in the same folder
4. Change the password in the `.env` file
5. Edit the .ini files
5. Run `docker-compose up`
6. Go to the AccountManager to set up accounts

(Such a docker based setup could probably be even extended to include features such as automatic backups of the DB every X minutes, which would probably be nice to prevent users from accidentally deleting their server DB)

It would be great if someone with experience in containerizing applications could give their opinion on this. There are probably some details around initialization that require some thought, such as when and how `MasterServer -a` is ran or how to expose the `.ini` files from the local file system to the docker container. 


Or are there any legal issues with providing a script that works with the contents of a `client.rar`? Obviously non of this would ever include  a  `client.rar` file but the install-script/Dockerfile would automatically run commands on the contents of the `.rar` file such as unpacking it with `pkextractor.pyw` and copying files to the `res` directory.

BloodyHellcat:
Technicaly seen it would be possible to drop one big docker container with all in it (client excluded) since you can simply run a hole unix system like with just two commands.

So if anyone would get a unix system container and set everything up, then i could be compressed into an docker image which could be shared and run with again just two commands.

Not me since I haven't been able to get to work till now 😅 but i work with docker at work so I see that it's possible with a good chance.

BloodyHellcat:
> # Description
> 
> This adds an effortless way to deploy the needed database for the server. It wraps up MariaDB in Docker and reads from a `.env` file for configuration. This way you only need Docker available to be able to run the database.

Have you fully been able to setup a functioning server?

VBMCBoy:
I agree with @Iterator256 on creating a bunch of containers for the various parts and then using docker-compose to combine them. If @jgkawell would like me to help, please reach out.

I'll just try the "normal" setup process in the meantime.

Hpmason:
I've been trying to work on the same thing and think it should be possible to build a built docker container for MasterServer and attach the client data as a volume. That way steps are just "1. put these client folders in X folder 2. Run docker-compose up". This should make.it more portable some the user doesn't have to spend the time building the image locally.

VBMCBoy:
Additionally, one could add a script, which checks if the necessary files are unpacked / converted to sqlite and prompt the user to do that or even do it automatically.

VBMCBoy:
Well, you could set the client folder to some environment variable and, when the server starts, check if everything is as it should be, and if not, modify the data accordingly. So one would not need to prepare the data.

VBMCBoy:
Also I'd argue that the average user can still be asked to run `docker-compose build` in addition to `docker-compose up -d`. That just takes a little longer, and still compiles everything on the user's side. Distributing finished images is probably not possible.

jgkawell:
This is a great conversation! I'm actually working on Dockerizing the whole application like you all have suggested. The process will be basically what @Iterator256 suggested in their comment above. It'll take a bit to get working due to the moving parts but I'm hoping to have it working by weeks end.

My goal is to have the user drop an *unpacked* client into a specified folder and then run `docker-compose build` followed by `docker-compose up -d`. That way it's super easy for even the least technical of users.

Also, @BloodyHellcat I have been able to get the server working using this Dockerized MariaDB installation. Not sure what you mean by a link to Docker Hub as the image is specified in the `docker-compose.yaml` file already.

VBMCBoy:
Great! I'm happy to help if you need any. I have some experience with Python and recently started using Docker at work.

TheNoim:
I am also currently working on a simple setup via docker-compose. My goal is it it: You only need to clone the repo and then start docker compose. The only other thing needed is the path to the client. Everything else, including database setup etc should happen automatically.

VBMCBoy:
I've currently held back on doing something like this because I didn't want to do something, that someone else is already doing.

Apart from that, Docker also stops everything together whenever I want it to stop, and starts everything together, when it should start. No more having multiple terminals open or having to start MariaDB with systemd. Very comfortable.

TheNoim:
> I've currently held back on doing something like this because I didn't want to do something, that someone else is already doing.

I mean, there a multiple ways to do the same things. Why not have the options. Even if somebody else is doing the same thing, maybe you could do it in a better way.

VBMCBoy:
Fair enough. And if I want, I could still make pull requests here or in other repos.

BloodyHellcat:
> This is a great conversation! I'm actually working on Dockerizing the whole application like you all have suggested. The process will be basically what @Iterator256 suggested in their comment above. It'll take a bit to get working due to the moving parts but I'm hoping to have it working by weeks end.
> 
> My goal is to have the user drop an _unpacked_ client into a specified folder and then run `docker-compose build` followed by `docker-compose up -d`. That way it's super easy for even the least technical of users.
> 
> Also, @BloodyHellcat I have been able to get the server working using this Dockerized MariaDB installation. Not sure what you mean by a link to Docker Hub as the image is specified in the `docker-compose.yaml` file already.

I'm sorry, since till now I wasn't at home and my PC i havent looked detailed at the files yet.
Thanks for the hint.

BloodyHellcat:
Quick Question, where did you get the Client from?

jgkawell:
> Quick Question, where did you get the Client from?

Can't say that publicly due to legal issues. Simply Google it and you'll find it.

TheNoim:
How do you guys solve the issue with the non headless pkextractor client? I don't know why this isn't headless to begin with. Currently, try to remove every dependence on this tk library.

VBMCBoy:
Oh great. I haven't looked at that because I didn't need it. But from what I could see right now, this will be fun. /s I assume your only (?) option is to remove the tk stuff and ingest the paths e.g. with argparse.
I'm currently working on the (other) python stuff, i.e. "dummy server" and account manager, to start with small steps. ;)

Hpmason:
> I don't know why this isn't headless to begin with. Currently, try to remove every dependence on this tk library.
It wasn't intended to be headless, which I understand https://github.com/lcdr/utils/issues/3

For the Docker build, It's probably best we assume that the user has an unpacked client. For users with a packed client,  we can provide a custom script to unpack all the necessary files/folders.

As for the build process, I'm going to see on my end if I can build a docker image without including any of the client data in the build process, while attaching volume when you start the container with the necessary client data. It may not come out very nice, but if it does I think it could be a better option for some people.

TheNoim:
> For users with a packed client, we can provide a custom script to unpack all the necessary files/folders.

No. This should make it easier for the user. The ideal would be, just setting the client path and docker takes care of everything else. Including unpacking if necessary. 

> As for the build process, I'm going to see on my end if I can build a docker image without including any of the client data in the build process, while attaching volume when you start the container with the necessary client data. It may not come out very nice, but if it does I think it could be a better option for some people.

I mean, this is the easy part. On container startup you can just link all files from a mounted client.

VBMCBoy:
Finally got Brick-Building-Fix and Account Manager working, see https://github.com/jgkawell/DarkflameServer/pull/2

DarwinAnim8or:
@MickVermeulen Could you take a look at this?

jaller200:
Alright, so I'm reading this atm — I've got a `docker-compose` file that sets up MariaDB, phpMyAdmin, Portainer, and builds and runs the MasterServer. The only manual step currently is creating a user in MySQL and importing the SQL script there, provided you have the required server files.

Docker may not be the best tool to unpack the files imho, primarily because that should be a one time thing whereas `docker-compose` will likely be turned on and off multiple times.

I would say it would be better to provide a clear guide / tools on how to manually do that, and then have them map in the proper files to the container.

Let me create a fork myself since I don't have write access here to demonstrate what I've got...

jaller200:
Alright, if you take a look at the changes I've made here (https://github.com/DarkflameUniverse/DarkflameServer/pull/81) you can see what I mean.

jaller200:
Also, right now you will manually need to run `./MasterServer -a` from inside the container, either via Portainer or via `docker exec`. 

The problem with fixing this is that it should only be run once. I'm looking into potentially turning it into a SQL INSERT command, but that requires checking if the encrypt method uses Bcrypt and can return properly. We could also create a standalone script too, or have instructions for how to quickly SSH into the box and run it

jaller200:
I've updated the script tentatively to pull account manager from my fork. Made a couple small changes to the Dockerfile on that one.

VBMCBoy:
I asked a couple oft days ago about the passwords and got told, that they use bcrypt.

If you use the mariadb image @jgkawell used, there is a folder that runs once when the database is created and never again. That would just require mounting the folder with the SQL files to the correct position.

Some other comments on your work:

- I think you can install the C++-Connector as libmysqlcppconn-dev in Ubuntu
- You're setting volumes based on the home directory and the folders in there. In my opinion you should set them relative to the docker-compose file, so one can clone the repo wherever they want.
- For the AccountManager you can copy what I did above, if you like.

Dec 7, 2021 07:25:43 Jonathan Hart ***@***.***>:

MickVermeulen:
Ideally if we're going with Docker I'd prefer the setup that @jaller200 created in #81 as that not only sets up the DB using Docker but everything else as well, actually reducing the amount of build steps. Just doing the database in Docker can lead to more confusion  as it's more optional setup, in my opinion.

jgkawell:
Converted this to draft. I think #90 is a more complete solution. I suggest following over there for continued conversation.

jgkawell:
Closing this. All changes here are incorporated in #90.

PR #48, opened 2021-12-06
Log Zone checksum to help with debugging

Jettford:
This will help people debug mismatched client file issues


DarwinAnim8or:
While the idea here is good, the code could be better. 
You don't need to convert the uint32 to a string here while logging, you would just do %i like you would with printf instead.

Jettford:
Didn't realise you could do that lol.

Jettford:
fixed @DarwinAnim8or

PR #51, opened 2021-12-06
Specify that MariaDB is needed for setup

Ramen2X:
The SQL script in the repo contains syntax that is intended only for MariaDB and does not work with MySQL. This README change will clear this up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ sudo apt install build-essential
 ## Setting up the environment
 
 ### Database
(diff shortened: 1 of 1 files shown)


Pr3roxDLC:
maybe changing the sql syntax so that it works with both migth be the better solution here

DarwinAnim8or:
What Pr3rox said, fixing the sql is a better option

Issue #4, opened 2021-12-05
Syntax errors in 0_initial.sql


QuirkyTurtle101:
(On windows) Attempting to run 0_initial.sql through the MySQL command line throws the following error:

`ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'TABLE accounts (
    id INT NOT NULL AUTO_INCREMENT PRIMARY KEY,
    name VARCHA' at line 1`

Putting the contents of the file through an online MySQL syntax checker throws the same error. I have no experience with MySQL to properly troubleshoot this, but I can only assume it's an issue with the file, rather than on my end, so I'm putting it here.

JoachimFlottorp:
Would also like to note out that MYSQL 8.0.27-0ubuntu0 does not support having a default value set to TEXT
 https://github.com/DarkflameUniverse/DarkflameServer/blob/27dace206b20006a6bfb20624374ed5731179690/migrations/dlu/0_initial.sql#L111

cristobers:
> (On windows) Attempting to run 0_initial.sql through the MySQL command line throws the following error:
> 
> `ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'TABLE accounts ( id INT NOT NULL AUTO_INCREMENT PRIMARY KEY, name VARCHA' at line 1`
> 
> Putting the contents of the file through an online MySQL syntax checker throws the same error. I have no experience with MySQL to properly troubleshoot this, but I can only assume it's an issue with the file, rather than on my end, so I'm putting it here.

Try using MariaDB https://mariadb.org/ instead of MySQL. I also had this issue and switching to MariaDB worked for me.

maxdelayer:
I had the same error, and then I removed the `OR REPLACE` from the first two tables and MySQL stopped complaining.

diminikov:
You can also add IF NOT EXISTS in front of CREATE TABLE to replace the effect of OR REPLACE

averysumner:
As stated above, these SQL files are only fully compatible with MariaDB. For now, opt to use MariaDB rather than MySQL.

Pr3roxDLC:
Maybe it should be changed in the ReadMe to say MariaDB instead of MariaDB or MySQL

PR #68, opened 2021-12-07
Make 0_initial.sql compatible with MySQL

Ramen2X:
Adjusts the syntax of 0_initial.sql to be compatible with MySQL while retaining MariaDB support. Changes `CREATE OR REPLACE TABLE` with `DROP TABLE IF EXISTS` + `CREATE TABLE` and wraps the '' in ugc's filename entry in parentheses.

This should close https://github.com/DarkflameUniverse/DarkflameServer/issues/4.

### DIFF
--- a/migrations/dlu/0_initial.sql
+++ b/migrations/dlu/0_initial.sql
@@ -1,4 +1,5 @@
-CREATE OR REPLACE TABLE accounts (
+DROP TABLE IF EXISTS accounts;
+CREATE TABLE accounts (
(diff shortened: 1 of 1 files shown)


MickVermeulen (approved):
LGTM, I just want to test it at a later time before merging, I'll try that tonight.

XenoTrixx (approved):
Works fine in MariaDB and MySQL.
Tested on the latest build of both.

MickVermeulen:
Thank you for helping with testing @XenoTrixx ! I tested on MySQL and that all works great.

PR #76, opened 2021-12-07
Update README.md

jelimoore:
add note about scripts - the scripts folder is necessary for some function to work - and the worldserver complains about not being able to find certain scripts occasionally

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ certutil -hashfile <file> SHA256
 
 **Setup resource directory**
 * In the `build` directory create a `res` directory if it does not already exist.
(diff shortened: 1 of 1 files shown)


averysumner:
DarkflameServer uses scripts written in C++ that are mapped to the original lua scripts' file names. This isn't needed.

jelimoore:
So how do we fix these entries in the logfiles?

`[06-12-21 21:17:47] [CppScripts]: Attempted to load CppScript for 'scripts\zone\AG\L_ZONE_AG.lua', but returned InvalidScript.`

averysumner:
That is likely a non-essential script as it's not implemented. You would implement a native CppScript (see dScripts) for it, then add an entry to the if/else chain in CppScripts.cpp.

jelimoore:
Gotcha, whenever I see some sort of fail like that in the logs my admin senses start tingling. Thank you for the clarification.

PR #81, opened 2021-12-07
WIP: Add Docker & Docker Compose Support

jaller200:
This request adds both Docker support via the `Dockerfile`, and Docker Compose support for running everything togeter via the `docker-compose.yml` file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,55 @@
+# syntax=docker/dockerfile:1
+
+# GCC Base
+FROM gcc:11.2.0 as build
+
+# Set Build Arguments
+#
+# 171022 = Unmodded Client
+# 171023 = DLU Client
+ARG BUILD_VERSION=171022
+ARG BUILD_ARGS=-j12
+
+# Working Directory
+WORKDIR /opt/darkflame-universe
+
+# Copy Source
+COPY ./ /tmp/darkflame-universe/
+COPY docker-entrypoint.sh /docker-entrypoint.sh
+
+# Install Dependencies
+RUN apt update && \
+    apt remove -y libmysqlcppconn7v5 libmysqlcppconn-dev && \
+    apt install -y cmake sudo zlib1g zlib1g-dev
+
+# Install MySQL C++ Connector
+RUN wget https://dev.mysql.com/get/Downloads/Connector-C++/libmysqlcppconn9_8.0.27-1debian11_amd64.deb -O /tmp/libmysqlcppconn.deb && \
+    dpkg -i /tmp/libmysqlcppconn.deb && \
+    rm /tmp/libmysqlcppconn.deb
+
+# Try to build
+RUN cd /tmp/darkflame-universe/ && \
+    sed -i -e "s/171023/${BUILD_VERSION}/g" CMakeVariables.txt && \
+    cat CMakeVariables.txt | grep VERSION && \
+    mkdir -p build && \
+    cd build && \
+    cmake .. && \
+    make ${BUILD_ARGS}
+
+# Copy back to working directory
+RUN mv /tmp/darkflame-universe/build/AuthServer ./AuthServer && \
+    mv /tmp/darkflame-universe/build/ChatServer ./ChatServer && \
+    mv /tmp/darkflame-universe/build/MasterServer ./MasterServer && \
+    mv /tmp/darkflame-universe/build/WorldServer ./WorldServer
+
+# Expose our ports
+EXPOSE 1001/udp
+EXPOSE 2000/udp
+EXPOSE 2005/udp
+EXPOSE 3000-3200/udp
+
+# Set default environment variables
+ENV SERVER_TYPE=debug
+
+# Start the server
+ENTRYPOINT [ "/docker-entrypoint.sh" ]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,8 +53,26 @@ services:
       - "8080:80"
     restart: always
 
+  # 
+  # Darkflame Universe
+  # The account manager
   #
-  # Darkflame Universe:
+  dlu-account:
+    build: https://github.com/jaller200/AccountManager.git#main
+    container_name: dlu-account
+    depends_on:
+      - mariadb
+    image: darkflame-universe-account:1.0.0
+    networks:
+      - darkflame_net
+    ports:
+      - "5000:5000"
+    volumes:
+      - "~/Desktop/Files/LU/files/account/credentials.py:/app/credentials.py"
+      - "~/Desktop/Files/LU/files/account/resources.py:/app/resources.py"
+
+  #
+  # Darkflame Universe: 
   # The master server
   #
   dlu-master:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,94 @@
+version: "3.9"
+services:
+
+  #
+  # Portainer:
+  # Used for managing Docker containers and administration.
+  #
+  portainer:
+    container_name: portainer
+    image: portainer/portainer:latest
+    networks:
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,55 @@
+# syntax=docker/dockerfile:1
+
+# GCC Base
+FROM gcc:11.2.0 as build
```
MickVermeulen, line 4: I'd definitely like to see a multi stage build for a build image and then a runtime image where you copy over the binaries from the build image to the runtime image after you're done building. GCC is a massive container and its binaries are no longer needed when running DLU.

For ref: https://docs.docker.com/develop/develop-images/multistage-build/

docker-compose.yml
```diff
@@ -0,0 +1,112 @@
...
+    container_name: dlu-account
+    depends_on:
+      - mariadb
+    image: darkflame-universe-account:1.0.0
+    networks:
+      - darkflame_net
+    ports:
+      - "5000:5000"
+    volumes:
+      - "~/Desktop/Files/LU/files/account/credentials.py:/app/credentials.py"
```
MickVermeulen, line 71: Instead of hardcoding some arbitrary path here, I'd prefer just using an empty directory in the DLU repository called `docker-config` or something where people can place all the files related to configuration. A benefit to this is that we make no assumptions about the host system and all the config will neatly be in one place. You can push empty directories to Git by placing a .gitkeep in them.

docker-compose.yml
```diff
@@ -0,0 +1,112 @@
...
+    image: darkflame-universe:1.0.0
+    networks:
+      - darkflame_net
+    ports:
+      - "1001:1001/udp"
+      - "2000:2000/udp"
+      - "2005:2005/udp"
+      - "3000-3300:3000-3300/udp"
+    volumes:
+      - "~/Desktop/Files/LU/files/authconfig.ini:/opt/darkflame-universe/authconfig.ini"
```
MickVermeulen, line 95: Here the same applies as for my previous comment about volumes.

MickVermeulen (changes requested):
Overall I really like this setup. I think the docker-compose can definitely help people set this up. I left a few comments in the source that I think would be nice changes to the docker config. Moreover I think it's good to explain in the readme that DLU can be run with Docker, along with some installation steps and build/run commands.

XenoTrixx:
Is there a reason to start an instance of portainer with the DLU-Server?

I guess most "Gamers" would try to run the container in docker desktop anyway.

docker-compose.yml
```diff
@@ -0,0 +1,112 @@
...
+    environment:
+      - MYSQL_ROOT_PASSWORD=password
+    networks:
+      - darkflame_net
+    ports:
+      - "3306:3306"
+    restart: always
+    volumes:
+      - db_data:/var/lib/mysql
+      - db_conf:/etc/mysql
```
XenoTrixx, line 36: I think it could be a good idea to include a default database and user.
So there is no need to create one for the people that dont know how.

```suggestion
  mariadb:
    command: --default-authentication-plugin=mysql_native_password
    container_name: mariadb
    image: mariadb:latest
    environment:
      - MYSQL_ROOT_PASSWORD: password
      - MYSQL_DATABASE: darkflame
      - MYSQL_USER: darkflame
      - MYSQL_PASSWORD: darkflame_pw
    networks:
      - darkflame_net
    ports:
      - "3306:3306"
    restart: always
    volumes:
      - db_data:/var/lib/mysql
      - db_conf:/etc/mysql
```

XenoTrixx (changes requested):

MickVermeulen:
Agree with both comments raised by @XenoTrixx , let's make it so that as much as possible is by default configured for local play. Anyone that wishes to host something publicly can probably deal with the extra setup steps. Regarding the passwords, though, ideally you'd want those in an .env file instead of in the source: we could have an .env.example that contains these default values that people simply need to copy over to an .env as a setup step, simple copy and paste commands like that aren't really an issue to add I think.

Dockerfile
jaller200: Great idea. I left it there for debugging, but I’ll clean it up and move everything to a small container

docker-compose.yml
jaller200: Sounds good to me as well. What about if I put everything in the resources directory for now?

They can put the `credentials.py` and `resources.py` files in the place too if they want, or under an `account` folder there.

docker-compose.yml
jaller200: That's actually a great idea! Also, I only included Portainer for debugging — most won't care, but it might make it easy if we need someone to SSH into a container, for instance.

Although we can always remove if if you feel it's still unnecessary.

Dockerfile
jaller200: Let me try an alpine container and see if I can get it working in that. Theoretically, it _should_ have all necessary runtime dependencies, but it might require installing some extra stuff related to libstdc++. I'm unsure though.

jaller200:
> I guess most "Gamers" would try to run the container in docker desktop anyway.

That's true. I keep forgetting Docker Desktop is a thing now. The only snag I could see is that I'm unsure if `docker-compose` can be run via the desktop app, but otherwise, we may not need it.

However, it might be nice to just include since it makes it super easy to SSH into the box if something goes completely wrong (for example, I made it to where you can set `SERVER_TYPE=debug` to simply `tail -f /dev/null` so the container will run but not start the server for manual debugging).

docker-compose.yml
jaller200: Updated the file paths

docker-compose.yml
jaller200: Updated the file paths

jaller200:
Also linking TheNoim's PR for reference: https://github.com/DarkflameUniverse/DarkflameServer/pull/90

I'm still unsure about whether I like Docker handling the extraction as opposed to just a script run by the user and mirroring the immutable folders, but if it's working, it seems to be a decent setup as well.

XenoTrixx:
> The only snag I could see is that I'm unsure if `docker-compose` can be run via the desktop app, but otherwise, we may not need it.

Docker Desktop does support docker-compose. =) 
grafik

docker-compose.yml
jgkawell: How about we use the approach of a `.env` file like I did in #36 . We could do passwords, paths, etc. all in that file and since Docker automatically reads a `.env` in the working directory into environment variables we can use all of them in the `docker-compose.yml` without having to edit it.

jaller200:
Ahh, wonderful — you learn something new every day!

I'll continue working on getting the script stabilised this evening after work. In the meantime, Noim is working on their own version that includes support for automatically unpacking the client.

luxaritas:
> Regarding the passwords, though, ideally you'd want those in an .env file instead of in the source: we could have an .env.example that contains these default values that people simply need to copy over to an .env as a setup step, simple copy and paste commands like that aren't really an issue to add I think.

FWIW, my personal preferred workflow is to have a `.env` with sensible default values cheked in, and then use a `.env.local` with any overrides. If there are required values to configure, the `.env` can just leave them blank (if you want an example, you can include it in a comment).

MickVermeulen:
@luxaritas yeah can agree with that, wasn't sure if that was a thing in Docker. Although that is indeed my workflow in other projects. If possible, I'd like to see that too.

jaller200:
Changing this to a draft since I believe https://github.com/DarkflameUniverse/DarkflameServer/pull/90 is a more complete solution. I'll continue working there.

averysumner:
Can we close this as #90 has been completed?

MickVermeulen:
Agreed with Avery, will close this.

PR #87, opened 2021-12-07
Add zone 1700 to list of accessible zones

enteryournamehere:
This removes the need to use the `force` flag when `/testmap`-ing to that zone.
I've also fixed the indentation in this function (so instead of a mix of spaces and tabs it's just spaces).

### DIFF
--- a/dGame/dUtilities/SlashCommandHandler.cpp
+++ b/dGame/dUtilities/SlashCommandHandler.cpp
@@ -1836,51 +1836,52 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 
 bool SlashCommandHandler::CheckIfAccessibleZone(const unsigned int zoneID) {
     switch (zoneID) {
-	case 98:
+        case 98:
         case 1000:
         case 1001:
-            
+
         case 1100:
         case 1101:
-	case 1150:
-	case 1151:
-	case 1152:
-            
+        case 1150:
+        case 1151:
+        case 1152:
+
         case 1200:
         case 1201:
 
-	case 1250:
-	case 1251:
-	case 1260:
-            
+        case 1250:
+        case 1251:
+        case 1260:
+
         case 1300:
-    	case 1350:
-    	case 1351:
-		    
+        case 1350:
+        case 1351:
+
         case 1400:
-	case 1401:
-	case 1450:
-	case 1451:
-            
+        case 1401:
+        case 1450:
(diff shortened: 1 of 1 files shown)


DarwinAnim8or:
The DLU style guide says to use tabs, not spaces.

DarwinAnim8or (changes requested):
Please convert spaces to tabs

enteryournamehere:
Sure, though the only style guide I could find is the [.clang-format](https://github.com/DarkflameUniverse/DarkflameServer/blob/main/.clang-format#L23) file, which specifies that 4 spaces should be used for indentation - for tabs it should have `IndentWidth: 8` and `UseTab: Always` (or `ForIndentation`).

averysumner:
As far as I was aware we were using spaces…

KurtStineAI2:
tabs should be just 4 spaces

luxaritas:
Usage of tabs is specified [in the CONTRIBUTING file](https://github.com/DarkflameUniverse/DarkflameServer/blob/0db954fa72a76b2640dfebcd7987ac3e2ebe0a46/CONTRIBUTING.md#general). Presumably that conflicts with the `.clang-format`

averysumner:
Can you also add 1500 to this?

EmosewaMC:
@averysumner Would it also be a good idea to add comments that say what each zone in the function is?  Or is that defined elsewhere in the files?

enteryournamehere:
> Can you also add 1500 to this?

Sure, done

DarwinAnim8or:
> > Can you also add 1500 to this?
> 
> Sure, done

1500 and 98 are not accessible zones and should not be in this list.
1500 is old starbase, technically present in the pk files but not in the versions list. (new starbase is 1600)
98 is Game Challenge, a leftover entry in the cdclient database that DLU used to insert frostburgh.

Xiphoseer:
Triage: @enteryournamehere Can you address
> 1500 and 98 are not accessible zones and should not be in this list.

Given the name of the PR, it seems like that's the best way forward

enteryournamehere:
> Triage: @enteryournamehere Can you address
> 
> > 1500 and 98 are not accessible zones and should not be in this list.
> 
> Given the name of the PR, it seems like that's the best way forward

Okay, reverted that again... if you guys now agree on using tabs for indentation (I see CONTRIBUTING.md was changed to mention that after the initial comment) I believe this should be good now.

PR #88, opened 2021-12-07
Windows fixes

Jettford:
- "Fixed" segfault
- Fixed logger not actually logging time

### DIFF
--- a/dCommon/dLogger.cpp
+++ b/dCommon/dLogger.cpp
@@ -92,8 +92,8 @@ void dLogger::Log(const char * className, const char * format, ...) {
 
 	va_end(args);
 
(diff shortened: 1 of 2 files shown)


Jettford:
@averysumner Please review

Jettford:
After trying to implement the above suggestion, I ran into some issues which shouldn't be possible.

Basically putting the 2 options in the ifdef above into an if statement, no matter the code path it took would still result in a segfault even with the code never running, I have a feeling this is down to a compiler issue or just Windows itself. 

The best solution would be fixing the ZCompression class to work correctly on Windows but after a few attempts, I can't seem to get it to work. It is possible it is just an issue present with how zlib is built into the server but without being able to compile the server in debug without issues I cannot pinpoint any problem exactly and cannot build a fix.

For now I would like to see this merged just temporarily as I and others cannot come to a solution.

Issue #61, opened 2021-12-06
Chat not working


basedx32:
Tested on a Vultr VPS with all ports open and a home server
OS: Ubuntu 20.04
Database: Mysql 8.0.27 and MariaDB 10.7.1 on docker

Problem: Local chat can not be used, any messages without a slash typed in show a loading button and then a red exclamation mark. Level ups do not show in the chat box either. Only slash commands that do work are /info and /credits, all others do nothing but do not show the red exclamation. Tested using gm levels 0, 1, 9. Others who managed to compile report the same issue but not everyone.

averysumner:
Have you confirmed an instance of ChatServer is running?

Landshark77:
Seeing the same issue on Ubuntu / MariaDB.

This is the output of the ChatServer log file:

```
[06-12-21 21:37:43] [ChatServer]: Starting Chat server...
[06-12-21 21:37:43] [ChatServer]: Version: 1.0
[06-12-21 21:37:43] [ChatServer]: Compiled on: Sun Dec  5 21:43:20 2021
[06-12-21 21:37:43] [dServer]: Server is listening on 192.168.1.35:1501 with encryption: 1
[06-12-21 21:37:43] [Server]: Established connection to master
[06-12-21 21:37:43] [ChatServer]: A server is connecting, awaiting user list.
[06-12-21 21:37:43] [ChatServer]: A server is connecting, awaiting user list.
[06-12-21 21:45:22] [PlayerContainer]: Added user: BooyahMagmaAroma (1152921508901814275), zone: 1000
[06-12-21 21:45:22] [ChatServer]: Asked for ignore list, but is unimplemented right now.
[06-12-21 21:45:32] [ChatServer]: A server is connecting, awaiting user list.
[06-12-21 21:45:38] [PlayerContainer]: Added user: Mindless^MHippo^MBaboon^M (1152921508901814274), zone: 1100
[06-12-21 21:45:38] [ChatServer]: Asked for ignore list, but is unimplemented right now.
[06-12-21 21:51:52] [PlayerContainer]: Removed user: 1152921508901814275
[06-12-21 21:51:55] [PlayerContainer]: Added user: BooyahMagmaAroma (1152921508901814275), zone: 1100
[06-12-21 21:51:55] [ChatServer]: Asked for ignore list, but is unimplemented right now.
[06-12-21 21:57:16] [PlayerContainer]: Removed user: 1152921508901814275
[...]

PR #91, opened 2021-12-07
Pull Request to fix chat messages

JoachimFlottorp:
My knowledge of this is these two variables return false
`std::string playerName = user->GetLastUsedChar()->GetName();`
`bool isMythran = user->GetLastUsedChar()->GetGMLevel() > 0;`

. Which in return makes the message fail to send.

image
Result if `if (!user->GetLastChatMessageApproved() && !isMythran) return;` is commented out!

- [x] Fix User is mythran
- [x] Fix Message is approved

Im not good at c++ at all, so if someone with more knowledge about c++ would like to help. that would be greatly appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,15 +162,6 @@ certutil -hashfile <file> SHA256
 * Copy over or create symlinks from `macros`, `BrickModels`, `chatplus_en_us.txt`, `names`, and `maps` in your client `res` directory to the server `build/res` directory
 * Unzip the navmeshes [here](./resources/navmeshes.zip) and place them in `build/res/maps/navmeshes`
 
-**Fix chatplus_en_us.txt**
-* If the file has been made in windows it has to be converted to support unix. Inside of WSL run this command
-* Move yourself into build/res using cd if you are not yet in there.
-```bash
-cd build/res
-sudo apt install dos2unix
-dos2unix -o chatplus_en_us.txt chatplus_en_us.txt
-```
-
 **Setup locale**
 * In the `build` directory create a `locale` directory if it does not already exist
 * Copy over or create symlinks from `locale.xml` in your client `locale` directory to the `build/locale` directory
--- a/README.md
+++ b/README.md
@@ -162,6 +162,15 @@ certutil -hashfile <file> SHA256
 * Copy over or create symlinks from `macros`, `BrickModels`, `chatplus_en_us.txt`, `names`, and `maps` in your client `res` directory to the server `build/res` directory
 * Unzip the navmeshes [here](./resources/navmeshes.zip) and place them in `build/res/maps/navmeshes`
 
+**Fix chatplus_en_us.txt**
+* If the file has been made in windows it has to be converted to support unix. Inside of WSL run this command
+* Move yourself into build/res using cd if you are not yet in there.
+```bash
+cd build/res
+sudo apt install dos2unix
+dos2unix -o chatplus_en_us.txt chatplus_en_us.txt
+```
+
 **Setup locale**
 * In the `build` directory create a `locale` directory if it does not already exist
 * Copy over or create symlinks from `locale.xml` in your client `locale` directory to the `build/locale` directory
--- a/dChatFilter/dChatFilter.cpp
+++ b/dChatFilter/dChatFilter.cpp
@@ -102,26 +102,14 @@ bool dChatFilter::IsSentenceOkay(const std::string& message, int gmLevel) {
 	if (message.empty()) return true;
 
 	std::stringstream sMessage(message);
-	std::string segment;
+	std::string line;
 	std::regex reg("(!*|\\?*|\\;*|\\.*|\\,*)");
 
-
-		
-
-	#ifdef _win32 
-	while (std::getline(file, line)) {
+	while (std::getline(sMessage, line)) {
 		line.erase(std::remove(line.begin(), line.end(), '\r'), line.end()); //Remove nix line-endings
 		std::transform(line.begin(), line.end(), line.begin(), ::tolower); //Transform to lowercase
-		m_Words.push_back(CalculateHash(line));
-	#else 
-	while (std::getline(sMessage, segment, ' ')) {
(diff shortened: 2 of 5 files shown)


DarwinAnim8or (changes requested):
I'm confused here-- the name being fetched has nothing to do with it. 
The mythran check is to see if you can bypass the spell check / word filter.

You have incorrect spelling in your message shown in the image, which means it's denied. DLU's chat works by checking against a list of approved words, not a list of bad words to be blocked. 

Please clarify what the issue is, and make sure it's not just the spelling errors that are preventing your messages. 
Thank you.

(marking as "requesting changes" seeing as it needs further elaboration and the code itself is just adding two log calls which shouldn't be there in a production enviroment and should be done with a debugger instead in a development enviroment)

JoachimFlottorp:
All those words i sent are verified, they show up as you type and all words can be found in the chatplus_en_us.txt file in build/res

The reason the message is red is because the server did not pass the message through the filter. Due to it being.. Broken.

i've looked further into it and found out that the problem is in file: dChatFilter.cpp `dChatFilter::IsSentenceOkay` which returns false when it checks up the messages against the filter. 
The message get's returned at the `if (!IsInWordlist(hash)`.
Even though just like i said, the message fits the filter, and should get passed through the filter.

The user i am running has a GM level of 9, which should set isMythran to true, however it does not.

DarwinAnim8or:
Note that account and character GM levels are seperate-- the account GM level is the maximum GM level that can be applied to a character owned by that account. 

That means you'd have to do /gmlevel <level> or one of it's alias commands in order to set the GM level for your character, if you have GM enabled then you should see a green "send" button rather than the default one.

Could you verify this first?
I couldn't replicate the bug you're describing, though I have seen that sometimes the client will still show red despite it being okay, but adding and removing a space usually fixes it

averysumner:
There are reports of chat working for other users, so there is definitely just something wrong with your configuration.

JoachimFlottorp:
Alright, i see what you mean by gm level, i was unaware that characters could have a gm level itself. 
However i do not understand how you are able to chat. As it is happening to every server owner that i have talked to. Can you possibly also try to set one of your characters to gm level 0 and try to type yourself. 

I've been looking more on how the filtering is done, and i suspect that how the DCF files are written is "broken" or done in a wrong way. as the word, for example "test" exists in the `chatplus_en_us.txt` but does not exist inside the converted `chatplus_en_us.dcf`. However, don't take my word for it and i could be wrong. i will commit a version with some debugging, this does not have fix anything, but shows more how the hashed data looks like.

averysumner:
Upon further inspection, the issue seems to lie with the line endings in the chatplus file. Look at #61 for more details. If you’d like you can commit a fix for this problem.

JoachimFlottorp:
Alright, lol. Seems like that was a very easy way to fix it. Thanks to @Wincent01 and @averysumner 
I'll commit a change to the README file talking about how to fix this.

JoachimFlottorp:
Open to change if the README does not specify well enough.

averysumner:
I’m wondering if it might be better to change the dcf code to just use windows line endings, probably easier for people setting it up.

Wincent01:
Maybe a better solution would be to accept both line-endings. Something along the lines of this
```c++
...
while (std::getline(file, line)) {
        line.erase(std::remove(line.begin(), line.end(), '\r'), line.end()); //Remove nix line-endings
	std::transform(line.begin(), line.end(), line.begin(), ::tolower); //Transform to lowercase
	m_Words.push_back(CalculateHash(line));
}
```

JoachimFlottorp:
@Wincent01 yeah, sounds like a better way of doing it, since you know how to do it.

dNet/ClientPackets.cpp
```diff
@@ -64,6 +64,8 @@ void ClientPackets::HandleChatMessage(const SystemAddress& sysAddr, Packet* pack
 
 	std::string playerName = user->GetLastUsedChar()->GetName();
 	bool isMythran = user->GetLastUsedChar()->GetGMLevel() > 0;
+	Game::logger->Log("IsUserMythran", "%s", isMythran?"Yes, they are mythran\n":"No, they are not mythran\n");
+	Game::logger->Log("IsMessageApproved", "%s", user->GetLastChatMessageApproved()?"Yes, Their message is approved\n":"No, their message is not approved\n");
```
Wincent01, line 68: Remove these

README.md
```diff
@@ -162,6 +162,15 @@ certutil -hashfile <file> SHA256
...
 
+**Fix chatplus_en_us.txt**
+* If the file has been made in windows it has to be converted to support unix. Inside of WSL run this command
+* Move yourself into build/res using cd if you are not yet in there.
+```bash
+cd build/res
+sudo apt install dos2unix
+dos2unix -o chatplus_en_us.txt chatplus_en_us.txt
+```
+
```
Wincent01, line 173: If we allow both line-endings this is unnecessary

JoachimFlottorp:
Something like this?

dChatFilter/dChatFilter.cpp
```diff
@@ -105,9 +105,22 @@ bool dChatFilter::IsSentenceOkay(const std::string& message, int gmLevel) {
...
+		m_Words.push_back(CalculateHash(line));
+	}
+	#else 
 	while (std::getline(sMessage, segment, ' ')) {
+	#endif
+
+
 		std::transform(segment.begin(), segment.end(), segment.begin(), ::tolower); //Transform to lowercase
-		segment = std::regex_replace(segment, reg, "");
+			segment = std::regex_replace(segment, reg, "");
```
Wincent01: The erase should be unconditional of platform. This will allow use of `\r\n` line-endings on Unix

JoachimFlottorp: New to this pull request on github. Am i the only one that can make commits?

DarwinAnim8or (approved):

Issue #105, opened 2021-12-07
Random Disconnects


IAmVolvic:
Fatal error 11
Stacktrace:
./WorldServer(+0xe4b2f)[0x5622a5d12b2f]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x153c0)[0x7fee3321f3c0]
/lib/x86_64-linux-gnu/libc.so.6(+0x16e052)[0x7fee32e31052]
./WorldServer(+0x1cb8d3)[0x5622a5df98d3]
./WorldServer(+0x1be36c)[0x5622a5dec36c]
./WorldServer(+0x1cffbe)[0x5622a5dfdfbe]
./WorldServer(+0xcd3db)[0x5622a5cfb3db]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3)[0x7fee32cea0b3]
./WorldServer(+0xdc29e)[0x5622a5d0a29e]
double free or corruption (out)

Fatal error 6
Stacktrace:
./WorldServer(+0xe4b2f)[0x5622a5d12b2f]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x153c0)[0x7fee3321f3c0]
/lib/x86_64-linux-gnu/libc.so.6(gsignal+0xcb)[0x7fee32d0918b]
/lib/x86_64-linux-gnu/libc.so.6(abort+0x12b)[0x7fee32ce8859]
/lib/x86_64-linux-gnu/libc.so.6(+0x903ee)[0x7fee32d533ee]
/lib/x86_64-linux-gnu/libc.so.6(+0x9847c)[0x7fee32d5b47c]
/lib/x86_64-linux-gnu/libc.so.6(+0x9a120)[0x7fee32d5d120]
/home/adminpie/Desktop/LegoU/DarkflameServerTest/DarkflameServer/build/_deps/mysql-src/lib64/libmysqlcppconn.so.9(_ZNSs7reserveEm+0x89)[0x7fee33374209]
/lib/x86_64-linux-gnu/libstdc++.so.6(_ZNSs6appendERKSs+0x3b)[0x7fee330f542b]
./WorldServer(+0xe4bb9)[0x5622a5d12bb9]

Fatal error 11
Stacktrace:
./WorldServer(+0xe4b2f)[0x5622a5d12b2f]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x153c0)[0x7fee3321f3c0]
[...]

PR #83, opened 2021-12-07
separate AgSpaceStuff and AgShipShake

red031000:
This should fix the below error on server startup
```
[07-12-21 04:58:49] [CppScripts]: Attempted to load CppScript for 'scripts\ai\AG\L_AG_SHIP_SHAKE.lua', but returned InvalidScript.
```
I have tested myself in game on a new character and verified that the desired functionality is present
The elimination of the error should also help people asking for help (and those providing help) in setting up the server, as they do not go looking for issues in potentially the wrong place (also it's an elimination of an error, which is always nice)

I've also fixed a cmake warning in the PR

PR #115, opened 2021-12-08
Revert "separate AgSpaceStuff and AgShipShake"

red031000:
Reverts DarkflameUniverse/DarkflameServer#83

yeahhhh so much for this

while it did fix the error, it's been causing the crashes in #105 

hopefully a better solution can be found, but until then reversion is necessary


red031000:
already done

PR #117, opened 2021-12-08
fix a cmake warning

red031000:
was bundled in my previous PR, while the code changes didn't work out, this is still probably useful to have


Jettford:
The endif should not even contain anything, The only reason it lets you is for legacy/backwards compatibility reasons 

https://cmake.org/cmake/help/latest/command/endif.html

averysumner:
The entire CMake file needs a rewrite. I’d rather not touch it at all.

PR #132, opened 2021-12-08
Add note to readme for troubleshooting tips with script editing

StefanH-AT:
In my case the client didn't load the fixed, unpacked script and still used the one from the pk. Renaming `res/pack/scripts.pk` to `_scripts.pk` fixed this for me.

Related:
https://www.reddit.com/r/legouniverse/comments/raoj0q/comment/hnony7p/?utm_source=share&utm_medium=web2x&context=3

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,8 @@ The client script for the survival minigame has a bug in it which can cause the 
 * Change `PlayerReady(self)` to `onPlayerReady(self)`
 * Save the file, overriding readonly mode if required
 
+If you still experience the bug, try deleting/renaming `res/pack/scripts.pak`.
+
 ### Brick-By-Brick building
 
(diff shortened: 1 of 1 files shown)


averysumner:
How do you have scripts.pk and a res/scripts folder? Can’t you just run an unpacked client?

Jettford:
> How do you have scripts.pk and a res/scripts folder? Can’t you just run an unpacked client?

Unpacking with lcdrs tool doesn't delete the packs, but the client will default to the pack files.

MickVermeulen:
LGTM, was wondering why people were still experiencing this.

StefanH-AT:
There should definitely a hint about that behavior somewhere, which is why i made this PR

MickVermeulen:
Agreed, given that there were no further comments, I'll merge.

PR #133, opened 2021-12-08
Fix unimplemented timestamp in dLogger::LogBasic on Win

cooltrain7:
Spotted that the timestamp for LogBasic on windows wasn't implemented. Follow up to the same fix in #88 which was for dLogger::Log.

### DIFF
--- a/dCommon/dLogger.cpp
+++ b/dCommon/dLogger.cpp
@@ -44,8 +44,8 @@ void dLogger::LogBasic(const char * format, ...) {
 	vsprintf_s(message, format, args);
 	va_end(args);
 
-	if (m_logToConsole) std::cout << "[" << "time machine broke" << "] " << message;
-	mFile << "[" << "time" << "] " << message;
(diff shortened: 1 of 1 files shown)


averysumner:
This still uses the C API for time. A better fix would be to utilize std::chrono and remove the platform specific code.

cooltrain7:
> This still uses the C API for time. A better fix would be to utilize std::chrono and remove the platform specific code.

This was really just targeted as a typo fix for what would be working code. While doing a refactor would be better in another pull.

averysumner:
I’d rather leave it broken and refactor than continue to mess with this broken C API code.

averysumner:
Furthermore, there was clearly a reason for this being disabled, as we wouldn’t have just left it like this without one.

cooltrain7:
> Furthermore, there was clearly a reason for this being disabled, as we wouldn’t have just left it like this without one.

Give the _exact_ same issue was fixed in #88, just for a diffrent function, I don't think that logic follows. But I'll close this for now none the less.

PR #134, opened 2021-12-08
Update BaseEnemyMech script to change the default faction

### LINKED ISSUE #54: Corrupted Sentries in Return to Venture Explorer not working
I just got to Return to Venture Explorer for the first time (so far everything else has been working great!) and the Corrupted Sentries are neither attacking nor attackable. They show a red (not purple!) health bar and walk around passively, but are not interactable in any way, which keeps me from doing the mission to smash 6 of them. The other enemies, both Stromlings and Spiderlings, are working as expected.
I'm running the server on WSL and the game on Windows, in case that might be relevant.

MickVermeulen:
Fixed #54. As part of the base enemy mech script its faction should be updated to 4
to make sure it's seen as an enemy by the client. The AgDarklingMech script
has been deleted as its functionality was essentially that of BaseEnemyMech
and thus no longer necessary.

### DIFF
--- a/dScripts/AgDarklingMech.cpp
+++ b/dScripts/AgDarklingMech.cpp
@@ -1,9 +1,0 @@
-#include "AgDarklingMech.h"
-#include "DestroyableComponent.h"
-
-void AgDarklingMech::OnStartup(Entity *self) {
-    auto* destroyableComponent = self->GetComponent<DestroyableComponent>();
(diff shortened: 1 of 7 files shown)


MickVermeulen:
Tested this on AG Mechs, AGS Survival mechs, RTVE mechs and AM mechs.

PR #149, opened 2021-12-09
blacklist support added to chat filter

Jettford:
Enabled support for the chat filter to use the chat words file as a blacklist


NealSpellman:
Chat filter is not changing on this repository. We stand by our current filter and if users want to make a change, they do so at their own risk and on their own servers.

Jettford:
This is a configuration option, not changing the norm.

NealSpellman:
Doesn't matter, a config option is still endorsing it. We do not endorse it.

PR #150, opened 2021-12-09
Add ship shake to not log about being missing

Jettford:
This log was causing confusion and issues when in reality there was no problem present.

### DIFF
--- a/dAuthServer/AuthServer.cpp
+++ b/dAuthServer/AuthServer.cpp
@@ -45,6 +45,7 @@ int main(int argc, char** argv) {
 	dConfig config("authconfig.ini");
 	Game::config = &config;
 	Game::logger->SetLogToConsole(bool(std::stoi(config.GetValue("log_to_console"))));
+	Game::logger->SetLogDebugStatements(bool(std::stoi(config.GetValue("log_debug_statements"))));
 
 	//Connect to the MySQL Database
 	std::string mysql_host = config.GetValue("mysql_host");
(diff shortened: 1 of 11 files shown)


averysumner:
Looks good to me.

MickVermeulen:
Not really a fan of this as people will just see invalidscript errors a bit later when going to AG. Either we actually fix the scripts or we just turn off the invalidscript warnings.

Jettford:
I think the main issue of this one is that you see it the second you open the server, later ones people tend to not really care about

Jettford:
Maybe the solution would just be changing the log to make it seem less like an error

DarwinAnim8or:
Hmm, I have my reservations about doing this. It makes sense for this script because it is in fact implemented, but on the logging of invalid scripts should be it's own issue IMO. 

I'd propose that we create a .ini variable to see if we should be logging them, a sort of "verbose logging" mode if you will

Jettford:
Would you like me to do that?

Jettford:
Or was it more of a at some point it should be done?

Jettford:
> Hmm, I have my reservations about doing this. It makes sense for this script because it is in fact implemented, but on the logging of invalid scripts should be it's own issue IMO.
I'd propose that we create a .ini variable to see if we should be logging them, a sort of "verbose logging" mode if you will

@DarwinAnim8or Done the above listed suggestions, take a look at latest commit

DarwinAnim8or (approved):
Have not tested locally but these changes seem sane

Jettford:
@averysumner @MickVermeulen Could you give this a review? Someone added you as reviewers.

MickVermeulen:
LGTM

MickVermeulen:
I'll merge, given that Avery has already approved and Darwin also approved.

PR #151, opened 2021-12-09
Moonbase fix

Jettford:
Gravity in Moonbase was broken at some point during the development of DLU and I don't think that it used to work the way I have just fixed it, can anyone think of any other solutions to the problem?

### DIFF
--- a/dGame/Entity.cpp
+++ b/dGame/Entity.cpp
@@ -1768,6 +1768,7 @@ void Entity::HandleTriggerCommand(std::string id, std::string target, std::strin
 			else if (argArray[0] == "repulse") effectType = 2;
 			else if (argArray[0] == "gravity") effectType = 3;
 			else if (argArray[0] == "friction") effectType = 4;
+
 			phanPhys->SetEffectType(effectType);
 			phanPhys->SetDirectionalMultiplier(std::stof(argArray[1]));
 			if (argArray.size() > 4) {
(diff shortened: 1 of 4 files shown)


MichaelFedora:
This is pretty curious -- why doesn't OnEnter fire alongside OnCreate when a play enters a world by default?

Jettford:
I think it might have happened when the physics engine was changed.

Jettford:
I will take a longer look at this now tbh

averysumner:
It previously worked by just sending the physics effect client sided.

DarwinAnim8or (changes requested):
While this is a solution to the problem, it is not elegant to do collision checks like this in a Component.

I'd much rather fix dPhysics so OnEnter gets fired upon creation inside of a trigger than do this. I'll take a look tomorrow, so no worries ;)

Jettford:
Alright I have taken a further look at dPhysics and I think I may be wrong about the root issue here, I am not really sure what the issue is, the TriggerEvent for OnEnter should be fired every tick as long as they are collided unless I missed something.

Jettford:
```
[10-12-21 20:46:58] [PhantomPhysicsComponent]: Processing events on new objects, x amount: 11
[10-12-21 20:46:58] [PhantomPhysicsComponent]: Character hit: 5652
[10-12-21 20:46:58] [PhantomPhysicsComponent]: Processing events on new objects, x amount: 11
[10-12-21 20:46:58] [PhantomPhysicsComponent]: Character hit: 5652
[10-12-21 20:46:58] [PhantomPhysicsComponent]: Processing events on new objects, x amount: 11
[10-12-21 20:46:58] [PhantomPhysicsComponent]: Character hit: 5652
```
Alright I know that the character is colliding with it and that `TriggerEvent("OnEnter", <character>)` is being called so I am not really certain where the issue is here

Jettford:
Simply constructing the trigger if its target is itself solves the issue.

@DarwinAnim8or if you could review

DarwinAnim8or (approved):

PR #171, opened 2021-12-11
Set chat filter to GM level 8

Brandon-Bullock:


averysumner:
What is this supposed to accomplish? You’re redefining a parameter… are you trying to just disable the chat filter?

Brandon-Bullock:
Yes I was disabling it for a local server.  Accidentally did a merge request with the main repo. Sorry about that.

Brandon-Bullock:
While this is not officially endorsed, if anyone wants to disable the chat filter for their PRIVATE servers then this is the easiest way to do it.

PR #177, opened 2021-12-11
Fix #164: Implement Buccaneer Valiant special ability

### LINKED ISSUE #164: The Broadsider's schooner won't work
### 

### Issue Description

The schooner for the Buccaneer's Broadsider spawns in but does not do any damage to the enemy. Also on the Ubuntu command prompt there is this issue [CppScripts]: Attempted to load CppScript for 'scripts\EquipmentScripts\BuccaneerValiantShip.lua', but returned InvalidScript. 

### Reproduction steps

Get the Broadsider and use the third ability

### Expected Behavior

I expected the schooner to ram the enemy but it just sits there doing nothing 

### Environment

I was running it on Ubuntu.

MickVermeulen:
Adds the ability for the buccaneer valiant to spawn a ship that rams
enemies and smashes them. Next to a script that triggers the ship skill
a few other changes had to be made:
- Force movement behavior server side calculation and sync
- The ship has no physics volume so the FindValidTargets for behaviors
had to be altered to allow ControllablePhysics entities to find entities
within their area. The "target_self" AOE flag has been used to replicate
the old behavior.

Attached is a video of the new behavior. As there were no variables in the BehaviorParameter table wrt the speed of the ship I'm curious to know if the current speed feels right.

Also keen on Wincent and Xipho reviewing the behavior changes I made. I wasn't sure what to test those on. If you have any suggestions let me know.

https://user-images.githubusercontent.com/22911950/145674133-a66d9a93-c079-4307-a3cc-94ab8c262a62.mp4

### DIFF
--- a/dScripts/BuccaneerValiantShip.cpp
+++ b/dScripts/BuccaneerValiantShip.cpp
@@ -1,0 +1,24 @@
+#include "BuccaneerValiantShip.h"
+#include "SkillComponent.h"
+#include "dLogger.h"
+
+void BuccaneerValiantShip::OnStartup(Entity* self) {
+    const auto skill = 982;
+    const auto behavior = 20577;
+    const auto skillCastTimer = 1.0F;
+
+    self->AddCallbackTimer(skillCastTimer, [self]() {
+        auto* skillComponent = self->GetComponent<SkillComponent>();
+        auto* owner = self->GetOwner();
+
+        if (skillComponent != nullptr && owner != nullptr) {
+            skillComponent->CalculateBehavior(skill, behavior, LWOOBJID_EMPTY, true, false, owner->GetObjectID());
+
+            // Kill self if missed
+            const auto selfSmashTimer = 1.1F;
+            self->AddCallbackTimer(selfSmashTimer, [self]() {
+                self->Kill();
+            });
+        }
+    });
+}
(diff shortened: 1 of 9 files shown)


MickVermeulen:
Not sure why the title didn't trigger this but fixes  #164.

MickVermeulen:
So @Wincent01 , or @Xiphoseer , any opinions?

Wincent01:
Looks good to me

Xiphoseer (approved):
Yeah, looks good to me as well. The AoE target logic may be missing some more flags as per <https://docs.lu-dev.net/en/latest/game-mechanics/behaviors/aoe.html> (generated from the DB), but I think this is a net improvement.

dScripts/BuccaneerValiantShip.cpp
```diff
@@ -0,0 +1,24 @@
+#include "BuccaneerValiantShip.h"
+#include "SkillComponent.h"
+#include "dLogger.h"
+
+void BuccaneerValiantShip::OnStartup(Entity* self) {
+    const auto skill = 982;
+    const auto behavior = 20577;
+    const auto skillCastTimer = 1.0F;
+
+    self->AddCallbackTimer(skillCastTimer, [self]() {
```
Wincent01, line 10: Have to capture skill and behavior here if you want to use them within the lambda. The current implementation isn’t standard.

Issue #175, opened 2021-12-11
ENH: Currency exploit


IAmVolvic:
### Is your feature request related to a problem?

The issue with this is anyone can join a server and drop themselves money and hand it out destroying the servers economy

### Describe the solution you'd like

 Essentially verify that if coins are dropped, the same amount is picked.

### Repository breaking implications

Pretty much fix an exploit,

### Describe alternatives you've considered

Disabling trading

Jettford:
Am working on this at the moment.

Jettford:
Done, Patch in #180

Wincent01:
Fixed in #180, closing

PR #180, opened 2021-12-11
Patched coin exploit

Jettford:
Stopped cheaters adding coins to their user that weren't actually present in the world.


Wincent01:
Fixes #175

PR #187, opened 2021-12-12
increased the exchange rate of the nimbus station console to 5 faction tokens for 25 maelstrom infected bricks

Nordegraf:
Currently when you exchange maelstrom infected bricks at the console you can build out of the anvil at nimbus plaza you receive one token per 25 bricks. There is a prompt ingame that says you should get 5 tokens.

I simply increased the corresponding numbers in the NsTokenConsoleServer.cpp from 1 to 5 and tested these changes on my own server running on Manjaro Linux.
factiontokens


averysumner:
Looks good to me, was this just a mistake on our parts @MickVermeulen @Wincent01?

Wincent01:
Looks good, a mistake on our end.

PR #189, opened 2021-12-12
Implement FDB Checksum

Jettford:
- Added config option `check_fdb`
- Added check in WorldServer.cpp
- Added raw MD5 function in MD5.cpp and MD5.h

### DIFF
--- a/dCommon/MD5.cpp
+++ b/dCommon/MD5.cpp
@@ -115,14 +115,6 @@ MD5::MD5(const std::string &text)
     finalize();
 }
 
-// raw md5 construstor
-MD5::MD5(const char * input, size_type length)
-{
-    init();
-    update(input, length);
-    finalize();
-}
-
 //////////////////////////////
 
 void MD5::init()
--- a/dCommon/MD5.cpp
+++ b/dCommon/MD5.cpp
@@ -115,6 +115,14 @@ MD5::MD5(const std::string &text)
     finalize();
 }
 
+// raw md5 construstor
+MD5::MD5(const char * input, size_type length)
+{
+    init();
+    update(input, length);
+    finalize();
+}
+
 //////////////////////////////
 
 void MD5::init()
--- a/dCommon/MD5.h
+++ b/dCommon/MD5.h
@@ -54,7 +54,6 @@ public:
     
     MD5();
     MD5(const std::string& text);
-    MD5(const char * input, size_type length);
     void update(const unsigned char *buf, size_type length);
     void update(const char *buf, size_type length);
     MD5& finalize();
--- a/dCommon/MD5.h
+++ b/dCommon/MD5.h
@@ -54,6 +54,7 @@ public:
     
     MD5();
     MD5(const std::string& text);
+    MD5(const char * input, size_type length);
     void update(const unsigned char *buf, size_type length);
     void update(const char *buf, size_type length);
     MD5& finalize();
--- a/dWorldServer/WorldServer.cpp
+++ b/dWorldServer/WorldServer.cpp
@@ -239,16 +239,19 @@ int main(int argc, char** argv) {
 		// pre calculate the FDB checksum
 		if (bool(std::stoi(Game::config->GetValue("check_fdb")))) {
 				std::ifstream fileStream;
-				fileStream.open ("res/CDServer.fdb", std::ios::binary | std::ios::in);
+				fileStream.open("res/CDServer.fdb", std::ios::binary | std::ios::in);
 				const int bufferSize = 1024;
 				MD5* md5 = new MD5();
 				
+				std::vector<char> fileStreamBuffer = std::vector<char>(bufferSize, 0);
+
 				while (!fileStream.eof()) {
-					char * fileStreamBuffer = new char[bufferSize];
-					fileStream.read(fileStreamBuffer, bufferSize);
-					std::streamsize size = ((fileStream) ? bufferSize : fileStream.gcount());
-					md5->update(fileStreamBuffer, size);
+					fileStreamBuffer.clear();
+					fileStream.read(fileStreamBuffer.data(), bufferSize);
+					md5->update(fileStreamBuffer.data(), fileStream.gcount());
 				}
+
+				fileStreamBuffer.clear();
 
 				const char* nullTerminateBuffer = "\0";
 				md5->update(nullTerminateBuffer, 1); // null terminate the data
--- a/dWorldServer/WorldServer.cpp
+++ b/dWorldServer/WorldServer.cpp
@@ -89,6 +89,7 @@ struct tempSessionInfo {
 std::map<std::string, tempSessionInfo> m_PendingUsers;
 int instanceID = 0;
 int g_CloneID = 0;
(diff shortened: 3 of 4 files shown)


Wincent01 (changes requested):
Reading such a big file in to memory each time someone loads into the world is unnecessary.

Could consider if it’s even worth doing it once, or if the server should just provide a MD5 checksum in the ini file. Loading a 40mb file into memory once will make the process hold on to that memory until termination, almost doubling the normal memory usage of a world server.

MickVermeulen:
Can agree with Wincent here that the current implementation is not ideal, I can see this causing quite a bit of lag. Not sure what you mean by holding on to memory @Wincent01, if we do this once at startup we can just free that memory afterwards again right? Just move this code to the world server init and store the hash for later use as a field.

Wincent01:
Ran into this when working with the locale. Loading that file into memory will make the process request a bunch of memory from the OS. Even if you free the memory, it won't be released back, the process will hold on to it. This it intended to as requesting memory from the OS is a very expensive operation.

MickVermeulen:
In that case, let's do a streaming hash. We have init, update and finalise functions, so a solution like [this](https://stackoverflow.com/questions/10324611/how-to-calculate-the-md5-hash-of-a-large-file-in-c) should be possible. Here it only loads it in kilobyte chunks.

Jettford:
Alright, I will make it calculate it at Zone creation and just have it be stored and I will make it so it loads it in chunks, good idea Mick

Jettford:
@Wincent01 @MickVermeulen It now calculates it at start-up of the world server and does it in 1024 byte chunks.

dCommon/MD5.cpp
```diff
@@ -115,6 +115,14 @@ MD5::MD5(const std::string &text)
     finalize();
 }
 
+// raw md5 construstor
+MD5::MD5(const char * input, size_type length)
```
Wincent01, line 119: Unused, remove this

dCommon/MD5.h
```diff
@@ -54,6 +54,7 @@ class MD5
     
     MD5();
     MD5(const std::string& text);
+    MD5(const char * input, size_type length);
```
Wincent01, line 57: Unused, remove this

dWorldServer/WorldServer.cpp
```diff
@@ -234,6 +235,30 @@ int main(int argc, char** argv) {
 		Game::physicsWorld = &dpWorld::Instance(); //just in case some old code references it
 		dZoneManager::Instance()->Initialize(LWOZONEID(zoneID, instanceID, cloneID));
 		g_CloneID = cloneID;
+
+		// pre calculate the FDB checksum
+		if (bool(std::stoi(Game::config->GetValue("check_fdb")))) {
+				std::ifstream fileStream;
+				fileStream.open ("res/CDServer.fdb", std::ios::binary | std::ios::in);
```
Wincent01, line 242: Remove the space here `open (`

dWorldServer/WorldServer.cpp
```diff
@@ -234,6 +235,30 @@ int main(int argc, char** argv) {
...
+				const int bufferSize = 1024;
+				MD5* md5 = new MD5();
+				
+				while (!fileStream.eof()) {
+					char * fileStreamBuffer = new char[bufferSize];
```
Wincent01, line 247: Align ptr to the left `char* `

Wincent01, line 247: Allocate this on the stack, not on the heap. Currently this will cause a memory leak as it's not deleted. Faster and better to not allocate on the heap at all.

Wincent01 (changes requested):
Revert the changes made to the MD5 files.
Don't allocate unnecessarily on the heap.
Some minor style changes.

dWorldServer/WorldServer.cpp
Jettford: VS won't stop doing this, it is really annoying, sorry

dWorldServer/WorldServer.cpp
Jettford: Same with this

Jettford: I am swapping this out with a std::vector now anyway because this doesn't compile on MSVC

Wincent01: You should be able to just make it a `char buffer[size]`

averysumner: > You should be able to just make it a `char buffer[size]`

Not if size isn’t a constexpr.

Jettford: Yeah, it only works on clang because of a compiler addon (iirc, ask avery about it), but you shouldn't be able to create a char array without the size being defined with a constexpr.

averysumner: > Yeah, it only works on clang because of a compiler addon (iirc, ask avery about it), but you shouldn't be able to create a char array without the size being defined with a constexpr.

Clang has a compiler extension that allows this. A compiler agnostic solution would be over-allocating in the first place or using a char vector.

Jettford: My latest commit makes it a char vector

averysumner (approved):
Looks good to me.

dWorldServer/WorldServer.cpp
```diff
@@ -234,6 +235,33 @@ int main(int argc, char** argv) {
 		Game::physicsWorld = &dpWorld::Instance(); //just in case some old code references it
 		dZoneManager::Instance()->Initialize(LWOZONEID(zoneID, instanceID, cloneID));
 		g_CloneID = cloneID;
+
+		// pre calculate the FDB checksum
+		if (bool(std::stoi(Game::config->GetValue("check_fdb")))) {
```
Wincent01, line 240: This will fail if there is no config value in the ini file, destroying backward comparability. Just compare the raw string with "1".

dWorldServer/WorldServer.cpp
```diff
@@ -838,7 +866,17 @@ void HandlePacket(Packet* packet) {
 		case MSG_WORLD_CLIENT_VALIDATION: {
 			std::string username = PacketUtils::ReadString(0x08, packet, true);
 			std::string sessionKey = PacketUtils::ReadString(74, packet, true);
+			std::string theirFdbChecksum = PacketUtils::ReadString(packet->length - 33, packet, false);
 
+			if (bool(std::stoi(Game::config->GetValue("check_fdb"))) && fdbChecksum != "") { // if fdbChecksum is empty, likely means we are a character server.
```
Wincent01, line 871: Same here

Wincent01 (changes requested):
Fix backwards compatibility

Jettford:
Added and tested the above fixes, all changes are made and is confirmed to work on Windows. Need someone to test on Linux

MickVermeulen:
Can verify tonight.

Wincent01:
Updated some of the style, logging, and made developers able to skip the check. @Jettford if you are OK with these changes we may merge.

Wincent01 (approved):

PR #200, opened 2021-12-12
Add the 'filter_gm_level' config option to the ChatServer

MasterEric:
This PR adds the new configuration option, `filter_gm_level`, to the `chatconfig.ini`. This configuration allows you to set the GM level at which chat filters are applied.

The default value, `1`, only allows forum moderators (GM level 1) and above to speak freely.

Setting the value to `9` restricts all players of GM level 8 and below from speaking freely in chat.

Setting the value to `0` disables chat filters for all players.

TODO:

- [ ] Test the implementation ingame.

### DIFF
--- a/dChatFilter/dChatFilter.cpp
+++ b/dChatFilter/dChatFilter.cpp
@@ -14,7 +14,8 @@
 
 using namespace dChatFilterDCF;
 
-dChatFilter::dChatFilter(const std::string& filepath, bool dontGenerateDCF) {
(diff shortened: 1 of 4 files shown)


NealSpellman:
As we've noted in previous chat filter PRs, we are not enabling a config to easily remove the filter. If a user would like to change how the filter handles, we do not endorse it and will not provide an easy way to do it.

Closing this PR.

PR #205, opened 2021-12-13
Fix proxy items removing each other's skills

### LINKED ISSUE #168: Piercing Patch Hood attack move not appearing on equip
### 

### Issue Description

The hood has a special spear throw attack which doesn't appear in your hotbar until you launch into the world so the current workaround is doing /testmap <world id>. /die or /kill doesn't seem to work.

Video: https://streamable.com/dzhjwo

### Reproduction steps

/gmadditem 15960. Equip/unequip the hat and see if the move appears in your hotbar, if it doesn't try /testmap <world id>.

### Expected Behavior

The spear throw move appears in your hotbar once the item is equipped.

### Environment

Windows Ubuntu

Daystar1998:
Fixed where equipping an item with two proxy items in the same slot would have the second proxy remove the first proxy's skill, which prevented some skills from being shown in the hotbar.

Fixes #168

### DIFF
--- a/dGame/dComponents/InventoryComponent.cpp
+++ b/dGame/dComponents/InventoryComponent.cpp
@@ -631,8 +631,6 @@ void InventoryComponent::LoadXml(tinyxml2::XMLDocument* document)
 			{
 				const auto info = Inventory::FindItemComponent(lot);
 
-				RemoveItemSkills(item->GetLot());
-
 				UpdateSlot(info.equipLocation, { item->GetId(), item->GetLot(), item->GetCount(), item->GetSlot() });
 
 				AddItemSkills(item->GetLot());
@@ -780,7 +778,6 @@ void InventoryComponent::Serialize(RakNet::BitStream* outBitStream, const bool b
 			
 			if (bIsInitialUpdate)
 			{
-				RemoveItemSkills(item.lot);
 				AddItemSkills(item.lot);
 			}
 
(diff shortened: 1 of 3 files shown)


averysumner:
Wait, why do you remove skills then add them again when the player loads? What does that accomplish?

Daystar1998:
Not removing skills first caused weird bugs.

Without removing skills in GameMessageHandler.cpp it would break initially worn equipment.
1. Equip items with skills.
2. Logout and login
3. Switch equipment
4. Switch to initial equipment

The skills will be missing from the hotbar. Every other equipment I switch to worked, just the initial was broken until I logged out and in again

Without the line in Serialize the initially worn equipment was missing skills when I logged in.

I didn't see any immediate issues without it in the LoadXML method, but after the other two I didn't want to take chances.

dGame/dComponents/InventoryComponent.cpp
```diff
@@ -1010,14 +1010,33 @@ void InventoryComponent::EquipItem(Item* item, const bool skipChecks)
...
-	GenerateProxies(item);
+	std::vector<Item*> proxies;
+	
+	std::vector<Item*> generatedProxies = GenerateProxies(item);
+	proxies.insert(proxies.end(), generatedProxies.begin(), generatedProxies.end());
+
+	// Clear skills in slots where the proxies will go
+	for (const auto proxy : proxies) {
+
+		RemoveItemSkills(proxy->GetLot());
```
MickVermeulen, line 1021: Why do you remove them here if they're never equipped? If this is to fix some confusion with the client I'm not a fan of this, that means we're definitely doing something wrong in our core logic, this is just a band-aid.

MickVermeulen (changes requested):
Overall not a fan of this, it seems to overcomplicate the already very complicated item system. I think that item equips may need to be context aware for when a proxy is equipped so that they don't override or proxy items need to be equipped after the main item.

dGame/dComponents/InventoryComponent.cpp
Daystar1998: Because the RemoveItemSkills method finds which slot the proxy item goes in and removes any skills which were previously equipped in that slot. This allows me to skip that same check when calling AddItemSkills on the proxy items so they don't remove each other's skills.

MickVermeulen:
@Wincent01 would you mind taking a look at this?

EmosewaMC:
Addressed by #496

PR #213, opened 2021-12-13
added tracking of Pet Excavator achievement

Nordegraf:
Currently the Pet Excavator Achievement (digging up 6 treasure chests on Pet Cove with your pet) is not tracked.

I added tracking of this achievement in the same function that tracks the progress of the Can You Dig It Mission and renamed it to better reflect its functionality. Because the Achievement is specific to the Pet Cove I check the mapID using the ZoneManager. Please let me know if this is not the intended way of doing this.

I tested this on a clean build server on Manjaro Linux on Pet Cove and on different maps. Receiving the reward and using it works fine.


averysumner:
Presumably if we didn’t include this in this script, some other script or mechanism was supposed to update this… @Wincent01?

Xiphoseer:
I think the fact that the object mentioned in the pet cove achievement is called "Pet Dig with achievements" (7612) and that mission task is handled by a script is a good indication that this might be tye correct way to do it.

MickVermeulen (changes requested):
So tracking this achievement is indeed done by this script, good catch. The current implementation does allow for digging the same treasure twice and still progressing, which is not the right behavior. 

So I checked and there's a different way to check if the treasure is supposed to track this  achievement, it will have the LDF int variable "PetDig" set which will refer to the n-th treasure this is, you can check this out by calling /inspect 2 -s on the treasure and looking at the variables.

Next up, we have to make sure that each treasure is only dug up once. This is achieved using player flags. The dig achievement flags start with `126` and ends with the `PetDig` variable value so you can get these by simply concatenating these two. E.g. for treasure with `PetDig` variable `2`, the respective player flag will be `1262`. Then by checking if the flag is not set you can progress the achievement by 1, after which you need to set the flag to make sure that digging it up again will not progress the achievement any further.

Nordegraf:
> So tracking this achievement is indeed done by this script, good catch. The current implementation does allow for digging the same treasure twice and still progressing, which is not the right behavior.
> 
> So I checked and there's a different way to check if the treasure is supposed to track this achievement, it will have the LDF int variable "PetDig" set which will refer to the n-th treasure this is, you can check this out by calling /inspect 2 -s on the treasure and looking at the variables.
> 
> Next up, we have to make sure that each treasure is only dug up once. This is achieved using player flags. The dig achievement flags start with `126` and ends with the `PetDig` variable value so you can get these by simply concatenating these two. E.g. for treasure with `PetDig` variable `2`, the respective player flag will be `1262`. Then by checking if the flag is not set you can progress the achievement by 1, after which you need to set the flag to make sure that digging it up again will not progress the achievement any further.

Hey Mick, thank you for your message! I'll be looking into the requested changes soon. I'm a bit short on time because of my research atm, so I'll probably take me a few days to get the fixes done.

EmosewaMC:
checking in on this PR - are you planning on finishing this anytime soon?  If not i will gladly finish this one up for ya

Nordegraf:
@MickVermeulen I implemented the tracking using the player flags. I hope the implementation is as expected.

@EmosewaMC Thanks for the offer. I've been setting this off for way too long.

PR #216, opened 2021-12-14
Make loot accurate to the loot drop rates during live. 

Jettford:
This PR is mostly made of repaired code from @averysumner that was never finished and has now been reimplemented.

I have no idea of knowing if these drops are accurate however Krysto (a prominent community member, who knows much more about loot than I) had spent a good amount of time with this build testing the drop rates.

### DIFF
--- a/dGame/dUtilities/Loot.cpp
+++ b/dGame/dUtilities/Loot.cpp
@@ -67,7 +67,7 @@ LootGenerator::LootGenerator() {
         }
 
         // sort in descending order based on randMax
-        std::sort(rarityTable.begin(), rarityTable.end(), [](const RarityTableEntry& x, const RarityTableEntry& y) { return x.randMax < y.randMax; });
+        std::sort(rarityTable.begin(), rarityTable.end(), [](const RarityTableEntry& x, const RarityTableEntry& y) { return x.randMax > y.randMax; });
 
         m_RarityTables.insert({index, rarityTable});
     }
@@ -125,7 +125,7 @@ LootGenerator::LootGenerator() {
 
         // sort by item rarity descending
         std::sort(lootTable.begin(), lootTable.end(), [&](const LootTableEntry& x, const LootTableEntry& y) {
-            return m_ItemRarities[x.itemID] < m_ItemRarities[y.itemID];
+            return m_ItemRarities[x.itemID] > m_ItemRarities[y.itemID];
         });
 
         m_LootTables.insert({index, lootTable});
@@ -191,6 +191,10 @@ std::unordered_map<LOT, int32_t> LootGenerator::RollLootMatrix(Entity* player, u
                         else if (missionComponent->GetMissionState(578) == MissionState::MISSION_STATE_COMPLETE)
                             drop.itemID = 8320; // "Paradox Token"
                     }
+
+                    if (drop.itemID == 13763) {
+                        continue;
+                    } // check if we aren't in faction
 
                     if (drops.find(drop.itemID) == drops.end()) {
                         drops.insert({drop.itemID, 1});
--- a/dGame/dUtilities/Loot.cpp
+++ b/dGame/dUtilities/Loot.cpp
@@ -37,7 +37,7 @@ LootGenerator::LootGenerator() {
 
     for (const uint32_t itemID : uniqueItems) {
         uint32_t itemComponentID = componentsRegistryTable->GetByIDAndType(itemID, COMPONENT_TYPE_ITEM);
-        CDItemComponent item = itemComponentTable->GetItemComponentByID(itemComponentID);
+        const CDItemComponent& item = itemComponentTable->GetItemComponentByID(itemComponentID);
 
         m_ItemRarities.insert({itemID, item.rarity});
     }
@@ -137,12 +137,12 @@ std::unordered_map<LOT, int32_t> LootGenerator::RollLootMatrix(Entity* player, u
 
     std::unordered_map<LOT, int32_t> drops;
 
-    LootMatrix matrix = m_LootMatrices[matrixIndex];
+    const LootMatrix& matrix = m_LootMatrices[matrixIndex];
 
     for (const LootMatrixEntry& entry : matrix) {
         if (GeneralUtils::GenerateRandomNumber<float>(0, 1) < entry.percent) {
-            LootTable lootTable = m_LootTables[entry.lootTableIndex];
-            RarityTable rarityTable = m_RarityTables[entry.rarityTableIndex];
+            const LootTable& lootTable = m_LootTables[entry.lootTableIndex];
+            const RarityTable& rarityTable = m_RarityTables[entry.rarityTableIndex];
 
             uint32_t dropCount = GeneralUtils::GenerateRandomNumber<uint32_t>(entry.minDrop, entry.maxDrop);
             for (uint32_t i = 0; i < dropCount; ++i) {
@@ -150,7 +150,7 @@ std::unordered_map<LOT, int32_t> LootGenerator::RollLootMatrix(Entity* player, u
 
                 float rarityRoll = GeneralUtils::GenerateRandomNumber<float>(0, 1);
 
-                for (RarityTableEntry rarity : rarityTable) {
+                for (const RarityTableEntry& rarity : rarityTable) {
                     if (rarity.randMax >= rarityRoll) {
                         maxRarity = rarity.rarity;
                     } else {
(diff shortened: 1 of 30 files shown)


Jettford:
Tested on Windows 11 (Clang) and Ubuntu 20 (GCC)

averysumner:
I’d like @NealSpellman’s input on these drops before it’s merged.

DarwinAnim8or (approved):
The code looks ok for me.  Didn't try to run anything wrt look accuracy though. 
@averysumner should be the one to have final say here

NealSpellman:
I don't have the time this week to test this, but I'd like someone or Krysto to chime in with data on and confirmations around all loot cases.
- Item packages
    - Packages with predetermined contents
    - Packages with random contents
- Boss loot
- Activity loot
- Team loot
- Wishing wells
- Mission loot (running a mission that requires drops from an enemy to confirm this didn't break)
...etc

This is a really major thing that we don't want broken on main, so someone will need to run this and test a _very wide_ variety of loot sources so we don't break loot for everyone. Preferably also running a simulator on many of the loot pools, such as bosses, to confirm this logic is working right and the odds are roughly the same but there still needs to be a lot of in-game testing in addition.

Jettford:
Alright I will get Krysto to chime in

Jettford:
Also team loot is completely broken but it is on main right now so.... nothing hurt

Jettford:
Alright I have a list of things tested by Krysto accompanied by a statement from him

- Surprise packs
- Consistent packs
- Boss loot
- Mission-specific loot
- Wishing wells

> If there's anything wrong with them, it would require extensive and exhaustive testing beyond the abilities of a single person
at a glance I have found no immediate or glaring issues

NealSpellman:
Still will need more testing than just one person, but it's good the initial results from Krysto are positive. This needs further testing from others and actual drop simulation work on your part to prove this is (roughly) matching what can be calculated from the database.

Verathian2:
Hey all,
I have found on my own server that the drops, especially from gorillas and admirals in GF have felt lower than the original game. Farmed admirals for a couple hours and only got a couple of shirts, pants, etc. (the kind of fillers you would get from dragons in FV). Not that you shouldn't get those things, but that's all I've been seeing.

averysumner:
> Hey all,
> I have found on my own server that the drops, especially from gorillas and admirals in GF have felt lower than the original game. Farmed admirals for a couple hours and only got a couple of shirts, pants, etc. (the kind of fillers you would get from dragons in FV). Not that you shouldn't get those things, but that's all I've been seeing.

It’s been a decade, are you basing this on feeling or video evidence?

averysumner:
> Still will need more testing than just one person, but it's good the initial results from Krysto are positive. This needs further testing from others and actual drop simulation work on your part to prove this is (roughly) matching what can be calculated from the database.

@Jettford maybe you could add a drop simulator command to this?

Jettford:
> @Jettford maybe you could add a drop simulator command to this?

Yeah I can add a command to simulate loot drops. Will add that now.

Jettford:
Implemented a `/rollloot` command to the server (do not run this on a live server, it will basically kill it) the format is as follows

`/rollloot <LootMatrixIndex> <Item ID> <times>`

e.g. 
`/rollloot 218 7570 1000`: Runs 1000 times gathering amount of runs it takes for a Stromling Ape (LMI: 218) to drop a parrot (item id: 7570)

It will provide an output as follows:
```
Ran loot drops looking for 7570, 1000 times. It ran 17467393 times. Averaging out at 17367.392578
```

averysumner:
> Implemented a `/rollloot` command to the server (do not run this on a live server, it will basically kill it) the format is as follows
> 
> `/rollloot <LootMatrixIndex> <Item ID> <times>`
> 
> e.g. `/rollloot 218 7570 1000`: Runs 1000 times gathering amount of runs it takes for a Stromling Ape (LMI: 218) to drop a parrot (item id: 7570)
> 
> It will provide an output as follows:
> 
> ```
> Ran loot drops looking for 7570, 1000 times. It ran 17467393 times. Averaging out at 17367.392578
> ```

At least for the red parrot, that's good convergence on the expected value of 17500 (if my math is correct). I would like some testing with some other objects too (dragons, stromlings, different tiers of enemies, etc).

MickVermeulen (approved):
Agree with Darwin, although I do also agree on this needing more testing.

dGame/dUtilities/Loot.cpp
```diff
@@ -1,384 +1,385 @@
...
-	for (const auto& pair : result) {
-		for (int i = 0; i < pair.second; ++i) {
-			GameMessages::SendDropClientLoot(user, source, pair.first, 0, spawnPosition, 1);
-		}
-	}
+        LootMatrix lootMatrix;
 
-	uint32_t coins = (int)(minCoins + GeneralUtils::GenerateRandomNumber<float>(0, 1) * (maxCoins - minCoins));
+        for (const CDLootMatrix& entry : matrix) {
+            LootMatrixEntry matrixEntry{entry.LootTableIndex, entry.RarityTableIndex, entry.percent, entry.minToDrop, entry.maxToDrop};
```
Wincent01, line 95: This will result in a lot of double storage. Consider using a ptr to the already cached table entry.

Wincent01 (changes requested):
Believe there is still some loot logic to be ironed out. You are able to get large amounts of duplicates from boss enemies like Butterscorch. Got four "Iron Barbute" once, and four "Denim Shirt" along with three "Block Shirt" another time.

NealSpellman:
> Believe there is still some loot logic to be ironed out. You are able to get large amounts of duplicates from boss enemies like Butterscorch. Got four "Iron Barbute" once, and four "Denim Shirt" along with three "Block Shirt" another time.

Yeah, for this stuff I'd suggest watching videos from Live about players smashing dragons and the loot they pick up. This seems broken if they're getting _this many_ duplicates.

Jettford:
> Yeah, for this stuff I'd suggest Jett watches videos from Live about players smashing dragons and the loot they pick up. This seems broken if they're getting this many duplicates.

I don't think this is the case, I think it is mainly down to you are smashing dragons solo

Jettford:
3-4 items seems right considering that is how the loot was kind of supposed to be divided up between teams.

NealSpellman:
> > Yeah, for this stuff I'd suggest Jett watches videos from Live about players smashing dragons and the loot they pick up. This seems broken if they're getting this many duplicates.
> 
> I don't think this is the case, I think it is mainly down to you are smashing dragons solo

For the LMI and activity related to a solo smash, you shouldn't be getting nearly that. Dragons do not drop the same amount of items for solo smashes as they would for higher groups. If this is indeed broken on main right now, then I need data on each of the LMIs that the different varying player numbers have to see if this works as expected.

Jettford:
So are you saying that the loot that should be dropped should be 1/4 of what it is right now?

NealSpellman:
> So are you saying that the loot that should be dropped should be 1/4 of what it is right now?

I'm saying there's 4 different LMIs for a regular dragon on Crux.

`{ "LootMatrixIndex" : 512, "Description" : "Crux Prime: Maelstrom Dragon Invader"},
{ "LootMatrixIndex" : 677, "Description" : "Crux Prime: Maelstrom Dragon Invader, 2 Players"},
{ "LootMatrixIndex" : 678, "Description" : "Crux Prime: Maelstrom Dragon Invader, 3 Players"},
{ "LootMatrixIndex" : 679, "Description" : "Crux Prime: Maelstrom Dragon Invader, 4 Players"},
`

I'm not sure what main branch is doing for it right now / if we take into account player count for regular dragon rewards, but the current results from this ticket don't sound like they're 1 player rewards. If this is broken on main, I'd settle for simulation data confirming all of these are working as intended if you don't intend to fix existing loot bugs.

Jettford:
ah this is interesting, I will take a look at the object data to work out how I can get these values on the server and fix it, but this issue is present on main right now

Jettford:
Alright I have gone through it, it is dropping 1 player loot and it appears to be correct

Jettford:
The only issue is the duplicates so I think your duplicate protection idea is worth looking into

dGame/dUtilities/Loot.cpp
Jettford: I think the speed of not checking through them on add outweighs the few kb of lost memory

averysumner:
This is significantly better than what we have now, I'm merging this.

Issue #178, opened 2021-12-11
BUG: Reading Nexus Tower Plaque crashes worldserver instance


leolion3:
### 

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### Issue Description

Reading the nexus tower plaque in nimbus station crashes the worldserver instance currently running for Nimbus Station. 
Server log (only gibberish):
```
Fatal error -1
Stacktrace:
./WorldServer(+0x7e9db)[0x5609a5ed59db]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x92ae6)[0x7f68e2d04ae6]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x92b21)[0x7f68e2d04b21]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x92d54)[0x7f68e2d04d54]
/LU/LegoUniverseServer/build/_deps/mysql-src/lib64/libmysqlcppconn.so.9(+0x73a7b)[0x7f68e36aea7b]
./WorldServer(+0x9e91b)[0x5609a5ef591b]
./WorldServer(+0x32f686)[0x5609a6186686]
./WorldServer(+0x1fc49b)[0x5609a605349b]
./WorldServer(+0x1f33f6)[0x5609a604a3f6]
./WorldServer(+0x7847d)[0x5609a5ecf47d]
```

### Reproduction steps

Attempt to read the nexus tower plaque in nimbus station

### Expected Behavior

It should show an empty list of participants who helped construct the tower.

### Environment

HiveOS (basically Ubuntu 18.04) on a physical local machine (not a VM) . MySQL is configured correctly with the username/password combos being set, the user having access to the database and the address being set to the MySQL LAN address that MySQL is listening on (not running on localhost).

The server is bound to a DNS address that translates into my server's static IP address.

Wincent01:
Unable to reproduce. Could you go into your CMakeVariables.txt file and enable some debug symbols, or preferably backtrace. Recompile with this and you should achieve meaningful crashlogs, these once are gibberish.

leolion3:
I'm well aware that these are gibberish, I am however not able to compile with the necessary debug flags, because as soon as I enable them I get the error messages

[...]

PR #217, opened 2021-12-14
Temporary fix for WorldServer instance crashing when fetching leaderboard data

leolion3:
When fetching leaderboard data, be it when reading the nexus tower plaque or when playing a minigame, the current WorldServer instance will crash because of an incorrect MariaDB SQL syntax in `dGame/LeaderboardManager.cpp`. While this pull request does not fix the incorrect syntax, it does prevent the crash by catching the exception thrown by the MySQL connector.

The printed exception is
```
You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'leaderboard_vales AS (     SELECT l.time, l.score, UNIX_TIMESTAMP(l.last_played)' at line 1
```

Its what caused the crashes in issues #202 and #178

### DIFF
--- a/dWorldServer/WorldServer.cpp
+++ b/dWorldServer/WorldServer.cpp
@@ -129,7 +129,7 @@ int main(int argc, char** argv) {
 	Game::logger->Log("WorldServer", "Compiled on: %s\n", __TIMESTAMP__);
 
 #ifndef _DEBUG
-	Game::logger->SetLogToConsole(false); //By default, turn it back off if not in debug.
+	Game::logger->SetLogToConsole(true); //By default, turn it back off if not in debug.
 #endif
 
 	//Read our config:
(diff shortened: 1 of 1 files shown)


Jettford:
Try Catches aren't used for performance reasons as far as I know and this isn't a solution. Surely the solution here isn't to slow down the entire network loop and to just fix the SQL?

leolion3:
Of course, but until the SQL syntax is fixed it may as well be a valid solution. Anyway I checked the queries and their SQL syntax is correct, however not for MariaDB. I myself have no experience with MariaDB, I only ever worked with native SQL.

All 4 leaderboard queries cannot be run under MariaDB. I got the newest version 10.6.5 installed.

leolion3:
Edit: The MySQL user one creates for the database has no access to the `dbo.UNIX_TIMESTAMP` procedure, hence the queries cannot be executed as he gets a permission denied. The DLU server thinks that the syntax is incorrect even though it is correct. Permissions may be granted using:
```
GRANT EXECUTE ON PROCEDURE dbo.UNIX_TIMESTAMP TO 'databaseUser'@'localhost'; 
```
or
```
GRANT EXECUTE ON PROCEDURE UNIX_TIMESTAMP TO 'databaseUser'@'localhost'; 
```

Closing pull request

PR #226, opened 2021-12-14
Updated Mail Commands

Verathian2:
- Added 3 new mail commands:
1. setmailsub: Similar to setanntitle with an announcement, this sets the subject of the next mail sent via mailitem or mailitemall.
2. setmailbody: Similar to setannmsg with an announcement, this sets the body of the next mail sent via mailitem or mailitemall.
3. mailitemall: Sends a mail to all characters created on the server (online and offline)
- Added "Help" messages to mail commands when the command is given no arguments.
- Altered the "mail sent" message to be more descriptive when mail is successfully sent.

### DIFF
--- a/dGame/Character.h
+++ b/dGame/Character.h
@@ -364,6 +364,30 @@ public:
      */
 	void SetAnnouncementMessage(const std::string& value) { m_AnnouncementMessage = value; }
 
+    /**
+     * Gets the subject of a mail message that a character made (reserved for GMs)
+     * @return the title of the announcement a character made
+     */
+    const std::string& GetMailSubject() const { return m_MailSubject; }
+
+    /**
+     * Sets the subject of a mail message a character will make (reserved for GMs)
+     * @param value the subject to set
+     */
+    void SetMailSubject(const std::string& value) { m_MailSubject = value; }
+
+    /**
+     * Gets the body of a mail message a character made (reserved for GMs)
+     * @return the body of the mail
+     */
+    const std::string& GetMailBody() const { return m_MailBody; }
+
+    /**
+     * Sets the body of an mail message to make (reserved for GMs)
+     * @param value the body of the mail message
+     */
+    void SetMailBody(const std::string& value) { m_MailBody = value; }
+
 	/**
 	 * Called when the character has loaded into a zone
 	 */
@@ -594,6 +618,16 @@ private:
      * The body of an announcement this character made (reserved for GMs)
      */
 	std::string m_AnnouncementMessage;
+
+    /**
+     * The subject of a mail message this character made (reserved for GMs)
+     */
+    std::string m_MailSubject;
+
+    /**
+     * The body of a mail message this character made (reserved for GMs)
+     */
+    std::string m_MailBody;
 
     /**
      * The spawn position of this character when loading in
--- a/dGame/Character.h
+++ b/dGame/Character.h
@@ -621,13 +621,15 @@ private:
 
     /**
      * The subject of a mail message this character made (reserved for GMs)
+     * Default: "Lost item"
      */
-    std::string m_MailSubject;
+    std::string m_MailSubject = "Lost item";
 
     /**
      * The body of a mail message this character made (reserved for GMs)
+     * Default: "This is a replacement item for one you lost."
      */
-    std::string m_MailBody;
+    std::string m_MailBody = "This is a replacement item for one you lost.";
 
     /**
      * The spawn position of this character when loading in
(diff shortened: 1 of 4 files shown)


DarwinAnim8or (changes requested):
Some things of note here, but the main one I have issue with is the "mailItemAll" command you created. Getting *all* created characters and then looping through them to do an insert + outputting to chat will be very slow. 

Also, to mimick live more, it should only give out the item to people who are currently online, for mythran events + then you don't need to hammer the MySQL database with possibly thousands of queries. 

### The other, smaller issues:

1. Please use `u"your text here"` for string constants, rather than GeneralUtils::ASCIIToUTF16. Strings have a large performance impact and you should use them, and converting strings, as little as possible. 
2. Following the above, use a stringstream to build sentences. Using + "" involved a lot of implicit string copying, which is slow, stringstreams were designed for this in mind and are faster!
3. Line 949 and 955 both use "GetMailSubject", one should be Body
4. Please do not chain if / else statements together. Leave one line of whitespace after closing an else
5. Please do not break up lines like below:
`ChatPackets::SendSystemMessage(
			sysAddr,
			GeneralUtils::ASCIIToUTF16("Sending item to " + playerName + " with subject '" + entity->GetCharacter()->GetMailSubject()
				+ "' and body '" + entity->GetCharacter()->GetMailBody() + "'.")
		);`


I know this sounds like a lot but I do appreciate any effort people put into this! The fact you wrote code for an open source project like this is great :) 

Only reason I wrote so much here is because I know C++ is a language that has lots of quirks and other languages are totally fine with some things that in C++ are *working*; but not the ideal solution. So if you're new I hope you can learn from it!

Verathian2:
Hey Darwin!

Thank you for all the feedback! I will definitely put it to good use in making the code a bit more efficient. Having only really dug into the code the last couple of days, I still have a lot to learn, and appreciate all the feedback!

**mailitemall**
- That sounds good to me! The primary reason I didn't do it for just online players is that the activity table doesn't always seem to track when a player logs off (activity = 1). I often found that activity = 0 (which I presume means online) was sometimes set on characters that I know aren't online. Perhaps this is a bug I will pursue a fix for soon!
- Another reason I was thinking of keeping this one in was that I wanted to mail the beta tester book to all characters on my server as a way to thank them for dealing with the technical difficulties we've faced in the first month of it being online. Perhaps this command could be kept with a lag warning, and a "mailitemallonline" could be created for the GM events some may wish to do? :) I'm fine either way, just let me know what you think! I could always write my own insert to run when the server is "in maintenance" for my next DLU update.

**Other Smaller Issues**
Thank you for all the notes! I'll make some changes, retest, and get another commit in so you can review and make sure the formatting is up to the DLU standard.

Thank you,
Verathian

Verathian2:
Hi Darwin,

I have updated the code reflecting your feedback. I also slightly modified the mail subject and body defaulting, as well as the way they are inserted into the insert query. I also updated mailitemall to only mail players with a max activity value of 0. The activity table as I mentioned before seems to have some offline characters in an online status, but that's a bridge we'll cross in another issue!

Verathian2:
Hello,

I have added one more, very minor adjustment to this request. While not related, it was so minor that I didn't want to create a whole new pull request. This update adjusts the stun time when you are "exiting" a paradox portal in NT. Previously the stun duration kept me standing there for multiple seconds after the animation ended. The adjusted time now lets me move when the animation is over.

Verathian2:
Adjusted the assembly portal as well, which also stunned the player for an excessive amount of time.

averysumner:
> Adjusted the assembly portal as well, which also stunned the player for an excessive amount of time.

Please put changes like this in different PRs.

DarwinAnim8or (changes requested):
Sorry for the delay in this, been busy.

I see the Nt Teleport scripts had a constant modified from 4 to 2, but not in the PlayAnimation call. Could you correct this? 
Also, where did you get the correct animation time from? CDClient?

Verathian:
Hey @DarwinAnim8or , I have actually made quite a few updates to this and run tests on my server (for the mail functions). I am going to close this PR and open a new one, as I've changed my process for coding PRs a fair amount in the last week.

As for the portal update, I have opened https://github.com/DarkflameUniverse/DarkflameServer/pull/289 which contains only these changes (per Avery's request). The animation time of 2 seconds was something I simply tested using trial and error, as the script `l_nt_assemblytube_client.lua` does not contain any timing on the client side, and a paradox portal script does not exist in the same file location. Therefore, I believe that this stun timing is only scripted on the client side.

I first tried 3 seconds, but found I was still stuck for a second (both for the assembly tubes and the paradox portals) after using them. I then reduced the time to 2 seconds, and I was able to move instantly as the animation ended. I modified these variables about 2 weeks ago on my server, and I've only heard good things about the change.

PR #233, opened 2021-12-15
Correct documentation for /instanceinfo

sersorrel:
`/instanceinfo` is currently documented as `/instance-info`, which silently fails.

I've tested that `/instanceinfo` works in-game, and that `/instance-info` does not. I don't know if either command was present on live.

(sidenote: would a PR with documentation for the other commands be accepted? I've written up notes for all currently-implemented commands, and would be happy to contribute them back here, either in the readme or as comments in the source or somewhere else.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,10 +289,10 @@ Here is a summary of the commands available in-game. All commands are prefixed b
   </tr>
   <tr>
     <td>
-      instance-info
+      instanceinfo
     </td>
     <td>
-      /instance-info
+      /instanceinfo
     </td>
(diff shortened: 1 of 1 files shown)


averysumner:
A PR to add docs for other commands would be great, feel free to add to this one.

cooltrain7:
> I've tested that /instanceinfo works in-game, and that /instance-info does not.

Following on with the above you can easily search this repo for said command to view it that way without having to test ingame/clone if you wanted to fix some of the others.

PR #228, opened 2021-12-14
CircleCI

leolion3:
Continuous integration using CircleCI.

PR #90, opened 2021-12-07
Completely automated docker setup

TheNoim:
My final goal is it to have a nearly completely automated setup via docker. The only thing the user should do is specify the client path. Everything else should happen completely automated. From client extraction to database setup and configuration.

PR #248, opened 2021-12-16
Add GitHub Actions Workflow

Xiphoseer:
This PR adds a github actions workflow for a Linux build as well as a `CMakePresets.json` that is defined by cmake and required by the action ([lukka/run-cmake](https://github.com/lukka/run-cmake)) that is used.

The `CMakePresets.json` file sets `jobs` i.e. the `-j` flag to `2` by default. That corresponds to the hardware as described in https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources and cuts the build time in half (about 8 minutes instead of 15).

Warnings end errors are scraped from the log and displayed inline in the GitHub web UI. See <https://github.com/Xiphoseer/DarkflameServer/pull/2/files> for an example. The warnings there are caused by adding `-Wextra`. This may be a good opportunity to tweak the warning settings a bit more to get more feedback on our own code but using different settings for 3rd party code.

At the end of a build, a zip file is uploaded as an artifact, which a maintainer can use to publish a release.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,33 @@
+name: Cmake CI
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+    - name: cmake
+      uses: lukka/run-cmake@v10
+      with:
+        configurePreset: 'default'
+        buildPreset: 'default'
+    - name: artifacts
+      uses: actions/upload-artifact@v2
+      with:
+        name: linux-build
+        path: |
+          build
+          !build/CMakeCache.txt
+          !build/build.ninja
+          !build/_deps
+          !build/cmake_install.cmake
+          !build/*.a
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,26 @@
+{
+  "version": 3,
+  "cmakeMinimumRequired": {
+    "major": 3,
+    "minor": 21,
+    "patch": 0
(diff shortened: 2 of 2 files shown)


Xiphoseer:
CC #228

averysumner:
Is there a possibility to add a Windows build to this with either clang or MSVC++?

Xiphoseer:
> Is there a possibility to add a Windows build to this with either clang or MSVC++?

Yeah, there are windows runners too, afaik the only difference is that every minute counts twice in terms of billing (10x for Mac). We need some way to install dependencies there and someone needs to write the actions yaml. I would suggest to only do windows builds on main though and not PRs.

TheNoim:
I like it. Later I could combine this with #90 and auto build docker images :)

averysumner:
> Yeah, there are windows runners too -- afaik the only difference is that every minute counts twice in terms of billing (10x for Mac). We need some way to install dependencies there and someone needs to write the actions yaml. I would suggest to only do windows builds on main though and not PRs.

Do we have a limited amount of minutes? I'm not entirely sure how Actions works... I know we ran out of artifact storage back when we used Actions on our private repository.

Xiphoseer:
> Do we have a limited amount of minutes? I'm not entirely sure how Actions works... I know we ran out of artifact storage back when we used Actions on our private repository.

According to <https://github.com/pricing> minutes should be free for public repos, but there are some limits for concurrent jobs, queues and artifacts as specified in <https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits>

averysumner:
I'm cool with merging this and having the possibility of adding a Windows build to it later.

averysumner (approved):

PR #251, opened 2021-12-17
Implement CDZoneTable PlayerLoseCoinsOnDeath

### LINKED ISSUE #174: BUG: Dying in some zones shouldn't make you drop coins, but does
### 

### Issue Description

some zones (i.e. WBL worlds) shouldn't punish you with coin loss on death: https://youtu.be/K2s7U9_OMDU?list=PL9EFB8500E625ADC7&t=2960

this is determined by the PlayerLoseCoinsOnDeath values in ZoneTable

currently not implemented

### Reproduction steps

die in portabello (physically, not just spiritually)

### Expected Behavior

no coins drop

### Environment

computer

cooltrain7:
A simple PR that wraps death coin drops with a check to the ZoneTable.PlayerLoseCoinsOnDeath. 
This checks when a player is smashed should they drop coins, and as per the live file this removes coin drops from

- Minigames
- Properties
- Portabello

Fixes #174.


Wincent01:
Would prefer if this boolean was stored somewhere, so it didn't have to be retrieved every time a player was smashed. `dZoneManager` might be a fitting place to put it, as it already has a query retrieving some values from that table.

averysumner:
> Would prefer if this boolean was stored somewhere, so it didn't have to be retrieved every time a player was smashed. `dZoneManager` might be a fitting place to put it, as it already has a query retrieving some values from that table.

Agreed, this is ridiculously inefficient. This value should be cached on startup, please fix this or close the PR.

cooltrain7:
> Would prefer if this boolean was stored somewhere, so it didn't have to be retrieved every time a player was smashed. `dZoneManager` might be a fitting place to put it, as it already has a query retrieving some values from that table.

Sounds good! Would you like it on the `dzoneManager` itself or in the sub `Zone`. I could add it as a private member variable, set it when the manager is initialized and then just add a getter for use?

Wincent01:
> Would you like it on the dZoneManager itself or in the sub Zone

Put is as a private `dZoneManager` member with a getter. `Zone` is mainly supposed to be a representation of the `luz` file, `dZoneManager` is used more for game logic.

cooltrain7:
> `Zone` is mainly supposed to be a representation of the `luz` file, `dZoneManager` is used more for game logic.

Good to know - I'll do that then!

cooltrain7:
Give that a check, should be good now.

averysumner:
It would be preferable to not use a SQL query.

cooltrain7:
> It would be preferable to not use a SQL query.

The query is already being run by the zoneManager on the exact table we want to access from. Seems to be the perfect place for it.

averysumner:
> > It would be preferable to not use a SQL query.
> 
> The query is already being run by the zoneManager on the exact table we want to access from. Seems to be the perfect place for it.

CDZoneTableTable has an optimized function for querying using an std::map stored in memory, it would be preferable to use this instead of a raw SQL query.

cooltrain7:
> > > It would be preferable to not use a SQL query.
> > 
> > 
> > The query is already being run by the zoneManager on the exact table we want to access from. Seems to be the perfect place for it.
> 
> CDZoneTableTable has an optimized function for querying using an std::map stored in memory, it would be preferable to use this instead of a raw SQL query.

I know its outside the scope of this PR but do you want me to replcae this query for the other values? Seeing as we're now doing both a lookup and a ref from the instance.

cooltrain7:
@averysumner Is the last commit what you had in mind?

dZoneManager/dZoneManager.cpp
```diff
@@ -27,19 +27,26 @@ void dZoneManager::Initialize(const LWOZONEID& zoneID) {
 	LOT zoneControlTemplate = 2365;
 
 	std::stringstream query;
-	auto result = CDClientDatabase::ExecuteQuery("SELECT zoneControlTemplate, ghostdistance_min, ghostdistance, PlayerLoseCoinsOnDeath FROM ZoneTable WHERE zoneID = " + std::to_string(zoneID.GetMapID()));
+	auto result = CDClientDatabase::ExecuteQuery("SELECT zoneControlTemplate, ghostdistance_min, ghostdistance FROM ZoneTable WHERE zoneID = " + std::to_string(zoneID.GetMapID()));
 
 	if (!result.eof()) {
 		zoneControlTemplate = result.getIntField("zoneControlTemplate", 2365);
```
averysumner, line 33: Why use CDZoneTableTable::Query only for acquiring the PlayerLoseCoinsOnDeath value? Why not use it for everything?

cooltrain7: Yea I did ask above if I should change both but I wanted to keep this PR on topic. I also noticed that whoever setup the constructor inside `CDZoneTableTable` also set the null value from the DB to be -1, but they're being stored in uints, so its causing a integer underflow. These would also need to be changed to either signed ints, or just change the deafult to 0.

Xiphoseer: IIUC, the c/c++ specifications defines integer underflow on unsigned integers to be modulo width (i.e 32 here), so setting it to -1 is well defined and equal to `0xFFFFFFFF`

c.f. https://stackoverflow.com/a/2760612/4087068

Xiphoseer:
Merging another PR caused a merge conflict, sorry!

dZoneManager/dZoneManager.cpp
cooltrain7: > IIUC, the c/c++ specifications defines integer underflow on unsigned integers to be modulo width (i.e 32 here), so setting it to -1 is well defined and equal to `0xFFFFFFFF`
> 
> c.f. https://stackoverflow.com/a/2760612/4087068

Oh thats good to know - I knew it existed just not the c/c++ specifics for how it was defined.

cooltrain7:
> Merging another PR caused a merge conflict, sorry!

Thats fine I still need to merge the removal of the now needless extra DB query.

dZoneManager/dZoneManager.cpp
```diff
@@ -26,19 +26,20 @@ void dZoneManager::Initialize(const LWOZONEID& zoneID) {
...
+	CDZoneTableTable* zoneTable = CDClientManager::Instance()->GetTable<CDZoneTableTable>("ZoneTable");
+	if (zoneTable != nullptr){
+		const CDZoneTable* zone = zoneTable->Query(zoneID.GetMapID());
+
+		if (zone != nullptr) {
+			zoneControlTemplate = zone->zoneControlTemplate != -1 ? zone->zoneControlTemplate : 2365;
+			const auto min = zone->ghostdistance_min != -1.0f ? zone->ghostdistance_min : 100;
+			const auto max = zone->ghostdistance != -1.0f ? zone->ghostdistance : 100;
+			EntityManager::Instance()->SetGhostDistanceMax(max + min);
+			EntityManager::Instance()->SetGhostDistanceMin(min);
```
Xiphoseer, line 38: This line was set to `max` in the previous and it seems like that may be causing issues. It seems we missed that in review.

EmosewaMC: Addressed in #435

PR #254, opened 2021-12-17
Crux Prime named enemy fix

EmosewaMC:
Fixed an issue where named enemies on crux prime would no longer spawn by un-commenting a line that was commented out.

I am unsure as to why this was commented out in the first place however un-commenting it does not seem to cause short term issues.

### DIFF
--- a/dScripts/BaseRandomServer.cpp
+++ b/dScripts/BaseRandomServer.cpp
@@ -91,7 +91,7 @@ void BaseRandomServer::SetSpawnerNetwork(Entity* self, const std::string& spawne
 
     if (spawnerName == "Named_Enemies")
     {
-        //spawner->Reset();
+        spawner->Reset();
     }
 
     spawner->Activate();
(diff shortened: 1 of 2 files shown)


MickVermeulen:
Probably best if @Wincent01 elaborates here, as he initially designed the script.

starjones101:
I tried this and one actually seemed to respawn within 10-15 minutes. Thanks.

EmosewaMC:
Simply doing a SoftReset() seems to do that trick as well.  Im thinking the issue is with `m_AmountSpawned` since the other two variables should not be changing between spawns.  This does not result in any crashes when spawning a named enemy as well.  I also removed some functions that were either deprecated or not being used anymore as they had no code in them.

Wincent01:
I don't recall why I commented that line. Might've been something along those lines, had not implemented SoftReset yet. I believe we should leave those empty functions there; the script is incomplete, an implementation of those is part of the full script.

EmosewaMC:
Sounds good.  I'll add them back in.  Once they are back in would this change be safe to push to main or does more research need to be done to make sure nothing adverse will happen?

EmosewaMC:
I ran a valgrind with the SoftReset change implemented and I saw no change in memory leaks.  Not sure if this is helpful information but thought i would make sure nothing adverse was leaking.

PR #266, opened 2021-12-19
Uncommented and repaired fix for quickbuilds

Jettford:


averysumner:
I mean I guess this "fixes" it...

Wincent01:
If I recall correctly this had a tendency to make things worse at times.

averysumner:
> If I recall correctly this had a tendency to make things worse at times.

Yeah there’s probably a reason it was commented out…

Jettford:
> If I recall correctly this had a tendency to make things worse at times.

From what I was told it was only bad if it happened while building.

PR #273, opened 2021-12-20
MasterServer now respawns ChatServer on crash

cuzitsjonny:
The code speaks for itself.


anarcal:
I'm pretty sure the WorldServers won't know to reconnect to the newly spawned ChatServer.

IAmVolvic:
As Anarcal said,

This would only spawn a new ChatServer instance, with the current worlds active they have no idea how to reconnect.
Wouldn't be best also to find out why the chat server crashes, and also have a way for it to reconnect in case of a crash.

cuzitsjonny:
Good point. A reconnect loop in WorldServer will be needed to fully implement a ChatServer respawn.

Issue #249, opened 2021-12-16
BUG: ChatServer tries to set wrong Player Ids into `friends`-Table


l-mbert:
### Make sure you've done the following:

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### Issue Description

When I try to add a Friend it throws a Foreign Key Error: 
```
terminating with uncaught exception of type sql::SQLException: Cannot add or update a child row: a foreign key constraint fails (`darkflame`.`friends`, CONSTRAINT `friends_ibfk_1` FOREIGN KEY (`player_id`) REFERENCES `charinfo` (`id`))
```

When I disabled the Foreign Key Checks with `SET FOREIGN_KEY_CHECKS=0` before the Server runs they Query, this is what it writes to the Database: 

![Screenshot 2021-12-16 at 22 41 05](https://user-images.githubusercontent.com/43070365/146453127-5408f01b-2b8a-4872-9669-037131ed6204.png)

## To get the Foreign Key Error:

- Run the latest Version of the Server
- Join the Server with a Friend or a second Client
- Ask this User to be your Friend
- The Log will output the Error written in the Description

## To get the Data to be written to the DB:

- In dChatServer/ChatPacketHandler.cpp in Line 134 add:

```cpp
auto stmtDFKC = Database::CreatePreppedStmt("SET FOREIGN_KEY_CHECKS=0");
stmtDFKC->execute();
delete stmtDFKC;
```

- Build the Server and start it
- Join the Server with a Friend or a second Client
- Ask this User to be your Friend
- The Friend-System works until one changes the World

### Expected Behavior

The Server should write the right User Ids to the Database so that the Friend and Team System will work.

### Environment

MacBook Pro 2019 (Intel Core i9, 16GB Ram, macOS 12.1)
Apple clang version 13.0.0 (clang-1300.0.29.30)

l-mbert:
Looking at the XML Data it seems to be an Id out of the `dest` and then `buff` Blocks.

I added some Logs before that and the Ids should be the right ones. I don't know why it's then setting the wrong Ids.

<img width="981" alt="Screenshot 2021-12-17 at 08 47 34" src="https://user-images.githubusercontent.com/43070365/146508552-13064ddf-4499-411e-9d41-51f909a0b0ac.png">

[...]

PR #276, opened 2021-12-20
ChatServer respawn after crash

cuzitsjonny:
- MasterServer now respawns ChatServer on crash
- WorldServer now attempts to reconnect to ChatServer

### DIFF
--- a/dMasterServer/MasterServer.cpp
+++ b/dMasterServer/MasterServer.cpp
@@ -47,9 +47,12 @@ namespace Game {
 bool shutdownSequenceStarted = false;
 void ShutdownSequence();
 dLogger* SetupLogger();
+void StartAuthServer();
+void StartChatServer();
 void HandlePacket(Packet* packet);
 std::map<uint32_t, std::string> activeSessions;
 bool shouldShutdown = false;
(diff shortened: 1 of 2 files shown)


IAmVolvic:
Has this been tested?

IAmVolvic:
> > Has this been tested?
> 
> Yes. Since you seem to be so worried, why don't you back me up by setting up a second test?

Weirdly while trying to figure out the issue with the ChatServer crashing we ended up fixing it by putting it in a try catch ( Wanted to get the root issue )

The crashing happens because of this error 

" ChatServer_1640005560.log:[20-12-21 10:59:04] [ChatServer] Bruh moment: Duplicate entry '1152921508901818643-1152921508901823303' for key 'friends.PRIMARY' "

DarwinAnim8or:
@IAmVolvic Ah, this may have something to do with the fact that before, the table the chat server uses (friends) was just plain fields, there were no restrictions placed on it whatsoever, so this couldn't happen. 

That explains why it's been crashing so much since FOSS.

l-mbert:
@DarwinAnim8or I might have narrowed it down to #249 if you'd like to take a look at it.

PR #282, opened 2021-12-21
Re-made Loot.h include <vector> to allow for the server to build on GCP

saltshaker:
Commit 0531365 made the server not build on GCP due to removing `#include <vector>` from Loot.h. This just re-adds it.

### DIFF
--- a/dChatFilter/dChatFilter.cpp
+++ b/dChatFilter/dChatFilter.cpp
@@ -8,14 +8,16 @@
 #include <regex>
 
 #include "dCommonVars.h"
-#include "Database.h"
 #include "dLogger.h"
+#include "dConfig.h"
+#include "Database.h"
 #include "Game.h"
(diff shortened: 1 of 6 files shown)


HailStorm32:
Can confirm. Made the following changes after pulling to get it to compile

averysumner:
What environment is "GCP"?

saltshaker:
Google cloud platform, so it's Debian 10.

averysumner:
I'm on Debian Buster compiling on GCC and I have no issues. Are you using clang?

saltshaker:
No I'm using GCC as well. I don't believe clang is installed on my machine.

Wincent01:
This PR does more than add back the include statement. Closing; open a new one with only the intended changes. Don't want a bypass of the chat filter in the history.

Issue #279, opened 2021-12-20
BUG: Client UI doubles currency gains from Achievements


MichaelFedora:
### Make sure you've done the following:

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### Issue Description

For multiple characters now, multiple times, I've been able to get large sums of money on Crux Prime due to killing named enemies for the first time, being excited about having over 200k, and when going back to Nexus Tower it gets lowered by around 50k.

This could be due to a few things:

- Achievements could be giving too much money (double triggering?) and the client isn't keeping proper track
- The server is not storing the right currency amount prior to world transfer and/or loses track in comparison to the client

I can try to do it again today to get some screenshots but it has definitely happened to me more than twice, on different characters.

### Reproduction steps

1. Gain a large amount of money while in one zone -- this can be done via killing Butterscorch and Roo Morg for the first time in Crux Prime
2. See your money is very high
3. Go to another zone (i.e. Nexus Tower)
4. See your money has lowered quite a bit

### Expected Behavior

The money would stay the same, or would be counted right the first time.

### Environment

Windows client / ubuntu server

Wincent01:
This is because the UI coin count double triggers for achievements; the server count is correct.

MichaelFedora:
Is this because something DLU is sending back, or is this not a fixable client-side bug?

EmosewaMC:
@Wincent01 Does this happen for all achievements?  If its an easy fix i'd like to try and give it a go to get to know the files a bit better.

Wincent01:
@lcdr has some insight into how this may be fixed. It has to do with the loot sources variables sent in the pickup currency messages. If that was set correctly the client wouldn't double trigger.

lcdr:
Yeah, in the code every call to `GameMessages::SendSetCurrency` needs to send the correct loot type ID, it appears it currently doesn't do that at all. [See also lu_packets](https://lcdruniverse.org/lu_packets/lu_packets/world/gm/client/struct.SetCurrency.html#structfield.source_type)

EmosewaMC:
I believe I have implemented this properly in #291.  GameMessages::SendSetCurrency now sends the correct loot type id for coins and no longer double ups on achievements.

PR #283, opened 2021-12-21
Fixed secondary coin UI update

EmosewaMC:
Removed a secondary UI coin update when completing achievements.  Fixes issues with #279.  I have tested this on one boss kill in Crux Prime and my coins showed the correct amount without having to transfer zones.


averysumner:
You can't just arbitrarily delete serialized values from a packet. This is going to cause issues.

EmosewaMC:
Ok, ill try to figure out the root cause then and push a PR with a proper fix.

PR #209, opened 2021-12-13
Leaderboards update to reflect live LU

Verathian:
https://github.com/DarkflameUniverse/DarkflameServer/issues/191

Per issue 191, the updates below should correct the leaderboards in DLU to match the functionality of the live game.

Issue #191, opened 2021-12-12
BUG: Leaderboards Not Updating (With Potential Fix)


Verathian:
### 

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### Issue Description

Leaderboards are not updating when a new highscore is achieved, and a highscore already exists.

### Reproduction steps

I have reviewed the LeaderboardManager.cpp code provided by DLU, and believe I have pinpointed the cause of the issue. See the below information/screenshot:

When a highscore already exists for the character, and a new highscore is achieved and confirmed (before this script's execution), the if (any) {...} branch would be executed. Because this script is missing a ";" at the end, however, the script does not execute to completion when "statement->execute();" is called.
![image](https://user-images.githubusercontent.com/46253834/145701912-1c399d82-910a-4d03-9c81-a487cc5fcfab.png)

### Expected Behavior

The leaderboard should update when a new highscore is achieved, not only when a the first score is recorded.

### Environment

Client: Windows 10 + Unpacked Client (unaltered outside of the steps provided by DLU).
Server: Ubuntu 20.04 running the recent DLU update with the corrected chat filter.

MickVermeulen:
Oh good catch, I'll take a look!

MickVermeulen:
Not able to reproduce for the AG race. I think it might be a minigame specific issue, could you elaborate on the minigame that you tried this on?

Verathian:
Hello,

I have taken a more thorough look today into the mini-games, and I believe that the missing ; may not be the root cause of the issue (although it should still be added). It looks like, based on some videos from the original game, that some games are not being scored correctly. As an example, Avant Gardens Survival. During live, you could have the lowest score out of all players, but if your time was longer, you had the highest rank on the leaderboard. Currently in DLU, it is the opposite. Right now the highest score is determining the leaderboard's standings.

I am working on a modified version of LeaderboardManager.cpp and LeaderboardManager.h that will correct these issues. Once I am finished, I will attach the code for a DLU Developer to review and test.

Thanks,
Verathian

Verathian:
**Rewritten LeaderboardManager.cpp**
`#include "LeaderboardManager.h"
#include <utility>
#include "Database.h"
#include "EntityManager.h"
#include "Character.h"
#include "GameMessages.h"
[...]

PR #288, opened 2021-12-22
Leaderboard Updates and Bug Fixes

Verathian:
Reopening https://github.com/DarkflameUniverse/DarkflameServer/pull/209

The following updates have been made to leaderboards:
1. Created a new worldconfig.ini option, "classic_survival_scoring". By default, this will be set to 0 in the config, and will make it so Avant Garden Survival scoring is dictated by score first, then by time. If this option is changed to 1, this order would flip so the scoring is dictated by time, then score (as it was in live LU). This option allows server owners to use the scoring set forth in the original game.
2. Added multiple new queries to ensure the score is gathered correctly from the database when a leaderboard must be populated.
3. Fixed a bug where the "last_played" value in the database was not being updated when the score was updated.

### DIFF
--- a/dGame/LeaderboardManager.cpp
+++ b/dGame/LeaderboardManager.cpp
@@ -3,8 +3,10 @@
 #include "Database.h"
 #include "EntityManager.h"
 #include "Character.h"
+#include "Game.h"
 #include "GameMessages.h"
 #include "dLogger.h"
+#include "dConfig.h"
 
 Leaderboard::Leaderboard(uint32_t gameID, uint32_t infoType, bool weekly, std::vector<LeaderboardEntry> entries,
                          LWOOBJID relatedPlayer, LeaderboardType leaderboardType) {
@@ -90,6 +92,7 @@ void LeaderboardManager::SaveScore(LWOOBJID playerID, uint32_t gameID, uint32_t 
         const auto storedTime = result->getInt(1);
         const auto storedScore = result->getInt(2);
         auto highscore = true;
+        bool classicSurvivalScoring = Game::config->GetValue("classic_survival_scoring") == "1";
 
         switch (leaderboardType) {
             case ShootingGallery:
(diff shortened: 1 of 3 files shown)


Verathian:
https://github.com/DarkflameUniverse/DarkflameServer/issues/191

MickVermeulen (approved):
Nice work, thanks for handing in another PR. Ideally I'd have some more people verify that this PR functions as intended as it changes quite a bit of SQL.

MickVermeulen:
Accidentally closed this, sorry.

HailStorm32:
Other than the one code change I requested, everything else works. Only leaderboards that don't update are shooting gallery and racing, however I believe it's a deeper issue than this pull request.

@MickVermeulen I would say it's ready to be merged if the code change I suggested is added.

MickVermeulen:
Sounds good, appreciate the suggestions. Lmk when all is ready!

EmosewaMC:
Going to review this in the morning and also test on a local branch to ensure changes work well.  If they work fine with the above suggestion then this should be ready to go.

HailStorm32:
Just came across the case where the leaderboard would wrongly update if you got a lower wave and a lower time.

Still trying to find a logic statement to account for that, while also still accounting for the other scenarios

dGame/LeaderboardManager.cpp
```diff
@@ -90,22 +92,36 @@ void LeaderboardManager::SaveScore(LWOOBJID playerID, uint32_t gameID, uint32_t
...
+                        highscore = false;
+                    }
+                }
+                else {
+                    if (score <= storedScore) // Based on score (DLU)
+                        highscore = false;
+                }
+                break;
             case SurvivalNS:
                 if (score < storedScore || time >= storedTime)
```
HailStorm32, line 126: Pretty sure I found the logic to account for all scenarios. 

```suggestion
                if (!(score > storedScore || (time < storedTime && score >= storedScore)))
```

HailStorm32:
@Verathian Are you still on this, or should I fork and resubmit?

PR #289, opened 2021-12-22
Portal Stun Time Reductions

Verathian:
The Nexus Tower Assembly and Paradox portals stunned players for an excessive amount of time after the "loading in" animation was complete. This corrects the time so the player can move once the animation ends.

### DIFF
--- a/dScripts/NtAssemblyTubeServer.cpp
+++ b/dScripts/NtAssemblyTubeServer.cpp
@@ -97,7 +97,7 @@ void NtAssemblyTubeServer::TeleportPlayer(Entity* self, Entity* player)
 
     GameMessages::SendPlayAnimation(player, u"tube-resurrect", 4.0f);
 
-    const auto animTime = 4;
+    const auto animTime = 2;
 
     const auto playerID = player->GetObjectID();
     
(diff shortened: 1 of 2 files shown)


averysumner:
Does this match the actual values from the scripts?

Verathian:
I was able to find a script on the client side for the assembly tube (res\scripts\02_client\map\nt\l_nt_assemblytube_client.lua), but there is no reference in the code to a stun time. I was not able to find a script for the paradox portals. 

This is a change I made to my personal server, and nobody has run into any issues. This has to do with how long the character is stunned on the server side of things.

Jettford:
The paradox portal scripts are based upon server scripts that aren't public, that's why you couldn't find it.

Verathian:
> The paradox portal scripts are based upon server scripts that aren't public, that's why you couldn't find it.

That makes the most sense to me. They would have been stored in the old LU server code.

MichaelFedora:
As someone who plays on Verathian's server, it feels really good. I can't move until exactly when the animation is done with the Paradox portal, which is perfect.

Jettford:
@averysumner These changes seem accurate, I would merge imo

MickVermeulen (approved):
Given the positive feedback on this PR, I'm fine with merging it.

PR #291, opened 2021-12-22
Implemented sourceType for SendSetCurrency

EmosewaMC:
In the function GameMessages::SendSetCurrency, the functionality that sends the correct source type to the client was not implemented and was simply set to 0.  This PR implements that functionality and changes the Character::SetCoins function to always SendSetCurrency where as before this would only happen if the coin source was not an achievement.


averysumner:
I know our codebase is messy, but there's no need to hardcode magic numbers into it more than we already have. Please make an enum for the loot source type.

EmosewaMC:
@averysumner That is what I did in [this commit](https://github.com/DarkflameUniverse/DarkflameServer/pull/291/commits/97f8e9e9fc62763bfd60df7a4dc43c32c917845b) no?

dCommon/dCommonVars.h
```diff
@@ -469,6 +469,31 @@ enum eRebuildState : uint32_t {
 	REBUILD_INCOMPLETE
 };
 
+enum eCoinSourceType : int32_t {
```
lcdr, line 472: Note that this enum is used not only for coins, but also for items (e.g. for achievements it delays the reward pickup animation until after the achievement complete animation has completed), so this should be called `eLootSourceType`. This is also what the game calls it internally. Same goes for the prefixes of the enum variants.

dGame/dMission/Mission.cpp
```diff
@@ -481,8 +481,10 @@ void Mission::YieldRewards() {
         inventoryComponent->AddItem(pair.first, count);
     }
 
-    if (info->reward_currency > 0) {
-        character->SetCoins(character->GetCoins() + info->reward_currency, info->isMission);
+    if (info->reward_currency > 0 && info->isMission) {
```
lcdr, line 484: You could just do a nested if clause and first check currency, then isMission, and assign the loot source type accordingly and only have one SetCoins call. Your way works too, but is a bit unusual.

lcdr:
Looks good overall, left some minor notes. Also, I don't work on the codebase itself, so a quick look by a maintainer is probably still a good idea.

dGame/dMission/Mission.cpp
EmosewaMC: I see what you mean.  I [re-did my statements](https://github.com/DarkflameUniverse/DarkflameServer/pull/291/commits/e84811bdcee9218b4c88c5ad6a3a4e63298cebcf) there to have only one SetCoins statement.

dCommon/dCommonVars.h
EmosewaMC: Ah gotcha.  [I have updated the enum name and prefixes](https://github.com/DarkflameUniverse/DarkflameServer/pull/291/commits/6ff5ae78aa0f27b8e13bb65eea75085fe6c6669c) to be correct.

EmosewaMC:
@averysumner @DarwinAnim8or if either of you two could review these changes to make sure everything looks good to you that would be great!

EmosewaMC:
also tagging @Wincent01 for review since I forgot to in my original comment.

dGame/dMission/Mission.cpp
```diff
@@ -473,16 +473,20 @@ void Mission::YieldRewards() {
...
-        // Sanitfy check, 6 is the max any mission yields
+        // Sanity check, 6 is the max any mission yields
         if (count > 6) {
             count = 0;
         }
 
         inventoryComponent->AddItem(pair.first, count);
     }
-
+
```
Xiphoseer, line 483:
```suggestion

```

Xiphoseer (approved):
Looks reasonable to me

dGame/dMission/Mission.cpp
```diff
@@ -473,16 +473,20 @@ void Mission::YieldRewards() {
...
         inventoryComponent->AddItem(pair.first, count);
     }
-
+    
     if (info->reward_currency > 0) {
-        character->SetCoins(character->GetCoins() + info->reward_currency, info->isMission);
+        int32_t lootSource = LOOT_SOURCE_ACHIEVEMENT;
+        if(info->isMission) {
+            lootSource = LOOT_SOURCE_MISSION;
+        }
```
Xiphoseer, line 488:
```suggestion
        int32_t lootSource = info->isMission ? LOOT_SOURCE_MISSION : LOOT_SOURCE_ACHIEVEMENT;
```

EmosewaMC:
@Xiphoseer Added both changes.

dGame/Character.h
```diff
@@ -309,12 +309,11 @@ class Character {
...
     /**
-     * Updates the current amount of coins of the character by a specified amount, for achievements this is not sent
-     * as it's tracked by the client
+     * Updates the current amount of coins of the character by a specified amount
      * @param newCoins the amount of coins to update by
-     * @param message whether to notify the client of the change
+     * @param coinSource The source of the loot
      */
-    void SetCoins(int64_t newCoins, bool message = true);
+    void SetCoins(int64_t newCoins, int32_t coinSource);
```
Xiphoseer, line 316:
```suggestion
    void SetCoins(int64_t newCoins, eLootSourceType coinSource);
```

dGame/Character.cpp
```diff
@@ -567,18 +567,15 @@ const NiPoint3& Character::GetRespawnPoint(LWOMAPID map) const
 	return pair->second;
 }
 
-void Character::SetCoins(int64_t newCoins, const bool message) {
+void Character::SetCoins(int64_t newCoins, int32_t lootSource) {
```
Xiphoseer, line 570:
```suggestion
void Character::SetCoins(int64_t newCoins, eLootSourceType lootSource) {
```

dGame/dGameMessages/GameMessages.h
```diff
@@ -78,7 +78,7 @@ namespace GameMessages {
 	void SendPlayFXEffect(const LWOOBJID& entity, int32_t effectID, const std::u16string& effectType, const std::string& name, LWOOBJID secondary = LWOOBJID_EMPTY, float priority = 1, float scale = 1, bool serialize = true);
 	void SendStopFXEffect(Entity* entity, bool killImmediate, std::string name);
 	void SendBroadcastTextToChatbox(Entity* entity, const SystemAddress& sysAddr, const std::u16string& attrs, const std::u16string& wsText);
-	void SendSetCurrency(Entity* entity, int64_t currency, int lootType, const LWOOBJID& sourceID, const LOT& sourceLOT, int sourceTradeID, bool overrideCurrent);
+	void SendSetCurrency(Entity* entity, int64_t currency, int lootType, const LWOOBJID& sourceID, const LOT& sourceLOT, int sourceTradeID, bool overrideCurrent, int32_t sourceType);
```
Xiphoseer, line 81:
```suggestion
	void SendSetCurrency(Entity* entity, int64_t currency, int lootType, const LWOOBJID& sourceID, const LOT& sourceLOT, int sourceTradeID, bool overrideCurrent, eLootSourceType sourceType);
```

dGame/dGameMessages/GameMessages.cpp
```diff
@@ -706,7 +706,7 @@ void GameMessages::SendBroadcastTextToChatbox(Entity* entity, const SystemAddres
 	SEND_PACKET_BROADCAST
 }
 
-void GameMessages::SendSetCurrency(Entity* entity, int64_t currency, int lootType, const LWOOBJID& sourceID, const LOT& sourceLOT, int sourceTradeID, bool overrideCurrent) {
+void GameMessages::SendSetCurrency(Entity* entity, int64_t currency, int lootType, const LWOOBJID& sourceID, const LOT& sourceLOT, int sourceTradeID, bool overrideCurrent, int32_t sourceType) {
```
Xiphoseer, line 709:
```suggestion
void GameMessages::SendSetCurrency(Entity* entity, int64_t currency, int lootType, const LWOOBJID& sourceID, const LOT& sourceLOT, int sourceTradeID, bool overrideCurrent, eLootSourceType sourceType) {
```

Xiphoseer (commented):
I think these changes are technically part of what @averysumner requested above. You should be able to just click "commit" on all 4 and then I'll merge this, but this isn't a "request changes" review on purpose because it's a really minor difference.

dGame/Character.h
EmosewaMC: ahh I understand.  since eLootSourceType is defined in dCommonVars.h as a int32_t, i can use eLootSourceType as the variable type since its basically a nickname.  Is this the correct understanding or no?

dGame/dMission/Mission.cpp
```diff
@@ -482,7 +482,8 @@ void Mission::YieldRewards() {
     }
 
     if (info->reward_currency > 0) {
-        character->SetCoins(character->GetCoins() + info->reward_currency, info->isMission);
+        int32_t lootSource = info->isMission ? LOOT_SOURCE_MISSION : LOOT_SOURCE_ACHIEVEMENT;
```
Xiphoseer:
```suggestion
        eLootSourceType lootSource = info->isMission ? LOOT_SOURCE_MISSION : LOOT_SOURCE_ACHIEVEMENT;
```
Missed one, sorry. Fortunately, there was a warning in the automatic checks. Unfortunately, I don't think you can see that 🤔

EmosewaMC: Noticed this one while I was compiling the changes.  Already pushed the change

dGame/Character.h
Xiphoseer: Yeah basically. With the added benefit that the compiler warns you when you try to set any integer to it that isn't annotated the same way.

Issue #268, opened 2021-12-19
BUG: normal green space jacket is turned into dark green space jacket when leaving char create


Terrev:
### Make sure you've done the following:

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### Issue Description

in char create:

![image](https://user-images.githubusercontent.com/21133460/146666143-fb36a17d-4acc-4827-86de-36ee37855ee9.png)

result upon saving minifig:

![image](https://user-images.githubusercontent.com/21133460/146666153-71336b33-d333-49f9-844d-589de9e704f1.png)

![image](https://user-images.githubusercontent.com/21133460/146666167-20139a71-dfb7-4e85-85b7-c4846b046d5f.png)

seems the server gives you the wrong item in this case, probably just a typo in that massive switch statement that handles it or whatever

it's possible there's more like this, I just stumbled into finding this one by chance

### Reproduction steps

choose a normal green space jacket in cc

### Expected Behavior

you get an item that matches that color+decal combo

### Environment

shoulder and back pain

Terrev:
update - definitely more than just the space jacket, but don't have time to look into specifics atm unfortunately

![image](https://user-images.githubusercontent.com/21133460/146726202-be848279-a56c-4d5f-8a38-e5d73de93e44.png)

![image](https://user-images.githubusercontent.com/21133460/146726217-4bd6c36a-ead3-4b0f-ad23-6605c645bc38.png)

![image](https://user-images.githubusercontent.com/21133460/146726353-665f261f-7988-4fa3-8caa-d6e82177ae1f.png)

![image](https://user-images.githubusercontent.com/21133460/146726371-73204875-e0ce-4ccd-8aa6-c6760c44911e.png)

Racater:
Can be closed

PR #293, opened 2021-12-22
Fix issue #268 : wrong shirt colors in the character selection

Racater:
This pull request fixes the issue #268 .
Now the characters will get the correct shirt color for shirts such as the space suit from the server following character creation (dark green and medium blue).
The error was probably due to a wrong copy and paste in the code.
Has been tested locally.

### DIFF
--- a/dGame/UserManager.cpp
+++ b/dGame/UserManager.cpp
@@ -566,10 +566,10 @@ void UserManager::LoginCharacter(const SystemAddress& sysAddr, uint32_t playerID
     }
 }
 
-uint32_t getShirtColorId(uint32_t color) {
+uint32_t GetShirtColorId(uint32_t color) {
 	
 	// get the index of the color in shirtColorVector
-	auto colorId = std::find (shirtColorVector.begin(), shirtColorVector.end(), color);
+	auto colorId = std::find(shirtColorVector.begin(), shirtColorVector.end(), color);
 	return color = std::distance(shirtColorVector.begin(), colorId);
 }
 
--- a/dGame/UserManager.cpp
+++ b/dGame/UserManager.cpp
@@ -582,7 +582,7 @@ uint32_t FindCharShirtID(uint32_t shirtColor, uint32_t shirtStyle) {
 	}
 
 	case 3: {
-		shirtID = shirtStyle >= 35 ? 5808 : SHIRT_DARK_GREEN;
+		shirtID = shirtStyle >= 35 ? 5748 : SHIRT_DARK_GREEN;
 		break;
 	}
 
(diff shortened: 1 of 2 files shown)


averysumner:
Looks good to me.

Jettford:
Maybe this should be replaced with a CDClient query such as this implementation of it. https://github.com/UchuServer/Uchu/blob/92894ace6f0f83ff1c9725b2bde58dfa4e883481/Uchu.Char/Handlers/CharacterHandler.cs#L97-L110

DarwinAnim8or (approved):

DarwinAnim8or:
@Racater hadn't seen your latest commit before reviewing, have you tested this works?

Racater:
@Jettford following your comment, I rewrote and simplified the code for the shirt at character creation. This way we remove this list of hardcoded values.

Racater:
This has been tested to work on my local server with all edge cases I could think of (first shirt style, last shirt style, different colors, etc.)

dGame/UserManager.cpp
```diff
@@ -566,114 +566,34 @@ void UserManager::LoginCharacter(const SystemAddress& sysAddr, uint32_t playerID
...
-		break;
-	}
-	}
-
-	// Initialize another variable for the shirt color
-	uint32_t editedShirtColor = shirtID;
-
-	// This will be the final shirt ID
-	uint32_t shirtIDFinal;
+uint32_t getShirtColorId(uint32_t color) {
```
MickVermeulen, line 569: This should be CamelCase

dGame/UserManager.cpp
```diff
@@ -566,114 +566,34 @@ void UserManager::LoginCharacter(const SystemAddress& sysAddr, uint32_t playerID
...
+	
+	// get the index of the color in shirtColorVector
+	auto colorId = std::find (shirtColorVector.begin(), shirtColorVector.end(), color);
```
MickVermeulen, line 572: `std::find (` should be `std::find(`

MickVermeulen (changes requested):
Have some code style comments.

dGame/UserManager.cpp
Racater: done

dGame/UserManager.cpp
Racater: fixed

MickVermeulen (approved):
Yeah LGTM, @DarwinAnim8or you assigned yourself for review, do you still plan on doing that or was it a misclick (given that you already approved before :P)

DarwinAnim8or (approved):

Issue #286, opened 2021-12-21
Anvil of armor and etc no longer work


MustacheMan2:
### Make sure you've done the following:

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### Issue Description

Anvil of armor and similar items no longer drop their powerups, broken from a recent commit, not sure which one though.

### Reproduction steps

place a cauldron of life, anvil of armor, etc. no power ups will drop.

### Expected Behavior

they should be dropping the powerups they say they will drop.

### Environment

ubuntu server

metric-ambient:
Me and my friend also having this bug since updating last night.

FutronBob:
Can confirm, been having this bug for the last couple days.  My guess would be from the loot table changes.

averysumner:
Has this persisted?

PR #294, opened 2021-12-22
Fix loot crash

Wincent01:
When applied this commit fixes a crash with the loot system due to a missing nullptr check.


BeeatZz:
hey sorry to bother but how do i apply this? i know next to nothing :(

Xiphoseer:
I think this PR might be causing #286, as this fix assumes that no object without a missions component will ever roll loot.

Issue #278, opened 2021-12-20
Boogie Down Achievement in Nimbus Station


stephanosstam:
### Make sure you've done the following:

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### Issue Description

Dancing with Kant Dance, Sofie Cusion or Peppy Slapbiscuit in NS doesn't register to the achievement Boogie Down.

### Reproduction steps

1.Go near to one of the before mentioned npcs
2.Do any dance emote

### Expected Behavior

Dancing with the given nps should register to the Boogie down achievement

### Environment

Wsl 1 using Elocore's video tutorial

godismyjudge95:
Can confirm this is bugged.

PR #299, opened 2021-12-23
Boogie Down Achievement + MissionTask Error Fix

Verathian:
I have included the following two fixes together, as they are both related to emotes and mission/achievement progression.

**Boogie Down Achievement Fix:** https://github.com/DarkflameUniverse/DarkflameServer/issues/278
- Fixed the "Boogie Down" achievement by adding a custom script "NsDanceNpcs" in a similar fashion to "SbLupTeleport". 
- This has been tested on a test server and completes the correct step in the achievement depending on which NPC is danced with.

**MissionTask Error Correction**
- Corrected the error "[MissionTask] Failed to find associated entity (0)!" by altering "GameMessages.cpp", which was popping up in the console any time an emote was used that is the condition for a mission/achievement the player is on, but the targetID of the entity was not found. The root cause was that the Progress function for the mission component was still being called, even when a target was not found (and is required).
- An example of this is when the player used a valid dance emote for the Boogie Down achievement, but the targetID of three entities you need to dance near were not found. This error would also occur for other missions/achievements.
- To validate this did not break missions where an emote is required for an NPC, I tested these changes with the mission "Cheer for the Championships!" (MissionID 2077), which requires you to cheer next to Nexus Naomi. The achievement completed successfully.

### DIFF
--- a/dGame/dGameMessages/GameMessages.cpp
+++ b/dGame/dGameMessages/GameMessages.cpp
@@ -5018,7 +5018,7 @@ void GameMessages::HandlePlayEmote(RakNet::BitStream* inStream, Entity* entity) 
 
 	//if (missionComponent) { // Altered to prevent "[MissionTask] Failed to find associated entity (0)!" every time a player emoted when not near the target for an active mission/achievement
 	if (missionComponent && targetID != LWOOBJID_EMPTY) {
-			missionComponent->Progress(MissionTaskType::MISSION_TASK_TYPE_EMOTE, emoteID, targetID);
+		missionComponent->Progress(MissionTaskType::MISSION_TASK_TYPE_EMOTE, emoteID, targetID);
 	}
 	
 	if (targetID != LWOOBJID_EMPTY)
--- a/dGame/dGameMessages/GameMessages.cpp
+++ b/dGame/dGameMessages/GameMessages.cpp
@@ -5014,9 +5014,11 @@ void GameMessages::HandlePlayEmote(RakNet::BitStream* inStream, Entity* entity) 
 	if (emoteID == 0) return;
 	std::string sAnimationName = "deaded"; //Default name in case we fail to get the emote
 
-	MissionComponent* mission = static_cast<MissionComponent*>(entity->GetComponent(COMPONENT_TYPE_MISSION));
-	if (mission) {
-		mission->Progress(MissionTaskType::MISSION_TASK_TYPE_EMOTE, emoteID, targetID);
+	MissionComponent* missionComponent = static_cast<MissionComponent*>(entity->GetComponent(COMPONENT_TYPE_MISSION));
+
+	//if (missionComponent) { // Altered to prevent "[MissionTask] Failed to find associated entity (0)!" every time a player emoted when not near the target for an active mission/achievement
+	if (missionComponent && targetID != LWOOBJID_EMPTY) {
+			missionComponent->Progress(MissionTaskType::MISSION_TASK_TYPE_EMOTE, emoteID, targetID);
 	}
 	
 	if (targetID != LWOOBJID_EMPTY)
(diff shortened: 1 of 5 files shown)


dGame/dGameMessages/GameMessages.cpp
```diff
@@ -5014,9 +5014,11 @@ void GameMessages::HandlePlayEmote(RakNet::BitStream* inStream, Entity* entity)
 	if (emoteID == 0) return;
 	std::string sAnimationName = "deaded"; //Default name in case we fail to get the emote
 
-	MissionComponent* mission = static_cast<MissionComponent*>(entity->GetComponent(COMPONENT_TYPE_MISSION));
-	if (mission) {
-		mission->Progress(MissionTaskType::MISSION_TASK_TYPE_EMOTE, emoteID, targetID);
+	MissionComponent* missionComponent = static_cast<MissionComponent*>(entity->GetComponent(COMPONENT_TYPE_MISSION));
```
MickVermeulen, line 5017: You can use `entity->GetComponent<MissionComponent>();` here.

MickVermeulen (changes requested):
I don't think this is the route to go down, I think we might be implementing something wrong with regards to UseEmoteTasks. I recommend going through the CDClient for mission 595, checking what is required there and what variables are available and then comparing that against our implementation, we might just be missing something critical there along the way.

EmosewaMC:
> I don't think this is the route to go down, I think we might be implementing something wrong with regards to UseEmoteTasks. I recommend going through the CDClient for mission 595, checking what is required there and what variables are available and then comparing that against our implementation, we might just be missing something critical there along the way.

Can confirm, when the target Entity is being read from the inStream, the entity being read has an LWOOBJID of zero instead of the intended value.

EmosewaMC:
From my own testing I have found that some npcs in the world flat out dont return as a valid target when an emote is used near them.  This may be a client sided issue but I can't confirm yet

EmosewaMC:
From looking around the files it appears that the other npc's with emote missions, namely the ones in forbidden valley, also have their missions hard coded.  @MickVermeulen In short there currently is an issue with how the client sends the LWOOBJID to the server and for entities like the concert ones and the forbidden valley gate ones, the server does not even receive a LWOOBJID.  Personally this implementation would be a good fix until a better one is found
Console logs for an Entity properly being the target of an emote:
`[10-01-22 20:02:27] [GameMessages] Emote (354) (288300748454049731)`
`[10-01-22 20:02:27] [GameMessages] Emote target found (1)`
Console logs for concert and forbidden valley gate npcs:
`[10-01-22 20:00:57] [GameMessages] Emote (372) (0)`
Since no LWOOBJID is found the game does not progress any emote tasks sent to GameMessages.  The forbidden valley gate guards work only because they are forced progress.  

One change I would want to see in this code before it is merged however is to not use the forceProgress if possible.  This should be using the missionComponent->Progress() method as this is cleaner than a forced progression.
@Verathian This is an example of what the progress would look like for this script:
`missionComponent->Progress(MissionTaskType::MISSION_TASK_TYPE_EMOTE, <The emote that was used>, self->GetObjectID());
This would allow less magic numbers in the code.

MickVermeulen:
@EmosewaMC thanks for the thorough rundown! I still think it's not worth the fix if we end up needing to fix it later; the less obvious the issue is during gameplay the less likely a proper fix will actually be made.

EmosewaMC:
@MickVermeulen For sure.  I prefer to have working, modular code instead of bodges.  Nothing much can be done then until we figure out exactly why the server just doesnt recieve object ids from certain npcs.

EmosewaMC:
After further investigation, this is not the fix that is needed for the mission.  The issue is somewhere else, possibly in the client, but a script like this is not the correct solution in the end.  Closing.

Issue #232, opened 2021-12-15
BUG: Inventor Beehive attack doesn't account kills to the player


DerPannda:
### 

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### Issue Description

The Inventor Improved Launcher 3 "Surprise Beehive" Attack causes the enemies to not drop loot for the player, if the attack itself kills the entity.

### Reproduction steps

1. Attack enemy with the "Surprise Beehive 2" of the Inventor Improved Launcher.
2. Wait for the enemy to die to the damage over time dealt by the attack
3. No loot will drop from the enemy despite being killed by an ability of the player.

### Expected Behavior

Instead of dropping nothing at all, the enemy is expected to drop loot as it was killed by an ability of the player.
Other abilities seem to be working fine such as Engineer Turret and Fortified Fortress or the Forcefield of the Innovator.

### Environment

Ubuntu

Terrev:
Unfortunately I don't have footage to back this up, but I'm pretty sure this was a bug during live, for what it's worth - at the very least, there was *some* problem with rank 3 beehive that made it useless, and I'm like 90% sure this was it (either that or it wasn't doing damage at all... but this sounds more familiar)

lcdr:
Update: It seems this skill uses a buff to actually deal the damage in the end, and the issue is with [BuffComponent.cpp not passing on the buff source to the skill system](https://github.com/DarkflameUniverse/DarkflameServer/blob/main/dGame/dComponents/BuffComponent.cpp#L74)  when executing the buff's behavior, and thus losing the info about the player being the one who should be credited.

averysumner:
Fixed in #306.

PR #306, opened 2021-12-24
Damage Buff source fix - inventor beehive bug

m888r:
Description:
- Adds an overload for `HandleUnmanaged` that includes the source
- Modifies damage buffs in `BuffComponent.cpp` to also include the source

Motivation and context:
Addresses issue #232 where the surprise beehive inventor skill does not credit kills to the player.

Tested by killing an enemy with the ability and seeing that it now drops loot, which it did not do previously.
[https://streamable.com/btz2vs](https://streamable.com/btz2vs)

### DIFF
--- a/dGame/dComponents/SkillComponent.cpp
+++ b/dGame/dComponents/SkillComponent.cpp
@@ -481,31 +481,12 @@ void SkillComponent::SyncProjectileCalculation(const ProjectileSyncEntry& entry)
 	delete bitStream;
 }
 
-void SkillComponent::HandleUnmanaged(const uint32_t behaviorId, const LWOOBJID target)
-{
-	auto* context = new BehaviorContext(target);
-
-	context->unmanaged = true;
-	context->caster = target;
-
-	auto* behavior = Behavior::CreateBehavior(behaviorId);
-
-	auto* bitStream = new RakNet::BitStream();
-
-	behavior->Handle(context, bitStream, { target });
-
-	delete bitStream;
-
-	delete context;
-}
-
 void SkillComponent::HandleUnmanaged(const uint32_t behaviorId, const LWOOBJID target, LWOOBJID source)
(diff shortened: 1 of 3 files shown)


dGame/dComponents/SkillComponent.h
```diff
@@ -161,6 +161,14 @@ class SkillComponent : public Component {
      */
     static void HandleUnmanaged(uint32_t behaviorId, LWOOBJID target);
```
lcdr, line 162: No need to duplicate the entire function, just change this to `static void HandleUnmanaged(uint32_t behaviorId, LWOOBJID target, LWOOBJID originator = LWOOBJID_EMPTY);`

dGame/dComponents/SkillComponent.cpp
```diff
@@ -499,6 +499,25 @@ void SkillComponent::HandleUnmanaged(const uint32_t behaviorId, const LWOOBJID t
 	delete context;
 }
 
+void SkillComponent::HandleUnmanaged(const uint32_t behaviorId, const LWOOBJID target, LWOOBJID source)
+{
+	auto* context = new BehaviorContext(target);
```
lcdr, line 504: The constructor already sets the originator, just use `new BehaviorContext(originator)` and don't set the originator afterwards.

m888r:
fixed both please re review when possible

lcdr:
Looks good to me 👍 . I'm not involved with the codebase though, so before merging one of the maintainers (probably @Wincent01) should still check if everything's in accordance with the coding style and if `originator` or `source` is preferred as a name.

m888r:
Thanks, and yeah I couldn't quite figure out which one is correct to use, so that would be very appreciated!

averysumner:
Works on my machine and works on a larger server, merging.

PR #309, opened 2021-12-26
Implemented a master API service

Jettford:
Implemented a master API server so that external services can access instances, players and manage those instances and players. This API has room for more additions but I feel this is a good base with some basic stuff needed for the AccountManager refactor.


Jettford:
The above has not been tested on Linux or macOS

DarwinAnim8or (changes requested):
From what I can tell-- this has 0 checks to see if someone has access to this data. Anyone could request it.

For a good API, you'd have to have some kind of authentication so that nobodies can't get data. Also I'm not sure sending player names + what world they're on is very GDPR-friendly, which is why I scrapped it from DLU.

Another thing that I had told you is that Master is not responsible for keeping track of who is where, it only needs to know how full an instance is, and do other instance management-related tasks.

I'd recommend removing a list of player names for these reasons, and of course add some form of authentication. Could be an auto-generated api key on first boot?

Jettford:
I didn't put any API key sort of checks in this because it should only ever be accessed from localhost for something like an account manager but alright, I will add an auth system

As for the player names not being GDPR friendly, alright I will make it so you supply the name and master sends the info to every instance and they forward it to the player.

Jettford:
Note:

After further speaking to Darwin over discord we decided that I would restrict the API to localhost operation and remove the master server player tracking to make use of the one that already exists on the chat server.

DarwinAnim8or (changes requested):
This is looking good.

I'm only requesting changes here because I noticed that the api_port doesn't have a fallback value, you should probably add that in the code rather than just expecting the variable has been defined in people's .ini files. (backwards compat and all that)

Jettford:
> This is looking good.
I'm only requesting changes here because I noticed that the api_port doesn't have a fallback value, you should probably add that in the code rather than just expecting the variable has been defined in people's .ini files. (backwards compat and all that)

I just set it to not start because if someone has said fall back port occupied for something the master server won't start.

Jettford:
This is also unfinished, still need to make some changes

Jettford:
all finished and solved the merge conflict

Jettford:
Adding a few more routes, then will be ready for review

Jettford:
Found more issues :cry:

Will fix it up later

DarwinAnim8or (changes requested):
As sent in private conversation with the author:

- api_port should have a fallback or the API shouldn't be started in the _Server.cpp files
- Duplicate code for announcements present

PR #313, opened 2021-12-27
Relax #294

Xiphoseer:
This PR relaxes the check added in #294 to see if that resolves #289 or anything else. I haven't tested this though,


Wincent01:
This is not what causes those items not to drop anything. It is caused by the loot rework removing an override of the loot calculation methods.

Xiphoseer:
Feel free to close it, if it isn't otherwise useful then 😊

PR #314, opened 2021-12-28
GM Mail Command Additions and Edits

Verathian:
This is a re-opened PR for PR https://github.com/DarkflameUniverse/DarkflameServer/pull/226. This version has been through much more testing, validation, and has some improvements over the prior PR.

**Additions:**
- `/setmailsender`: Sets the mail "sender" line of the next GM mail sent. Maximum 20 characters.
- `/setmailsub`: Sets the mail "subject" line of the next GM mail sent.
- `/setmailbody`: Sets the mail "body" of the next GM mail sent.
-  `/mailitemall`: Mails an item to all online or all existing characters. Sending to online will be based on the activity log table (which doesn't seem to be accurate every time, but is the only way to get all players cross-world), which attempts to replicate the GM mail events near the end of LU. Sending to all players will send to every character that exists in the charinfo table. This version of the command could be slightly taxing on the database, so it shouldn't be done often (as it inserts into the "mail" table once for every character).

**Modified Commands:**
- `/mailitem`: This command has been modified to allow the custom sender, subject, body to work. It has also been given a syntax helper when the proper number of arguments are not given, the ability to send more than 1 of an item, and more feedback when an item is successfully sent.

**Philosophy for Changes:**
The changes above are quality of life improvements, intended to make the life of a server operator easier, and to expand on the features of the existing mail function. These commands allowed me to send out a custom message on my server with candy canes for a Frostburg Item Vendor I added to Nimbus Plaza (as a holiday event). All players have received the items with no issue, and the `/mailitemall all ... ...` command that was used did not cause any strain on the server, and made my life 10x easier for distributing items.

Let me know if there are any concerns or questions.
Tagging @DarwinAnim8or , since you were assisting me with the original request.

### DIFF
--- a/dGame/dUtilities/SlashCommandHandler.cpp
+++ b/dGame/dUtilities/SlashCommandHandler.cpp
@@ -868,7 +868,7 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 
 	if (chatCommand == "mailitem" && entity->GetGMLevel() >= GAME_MASTER_LEVEL_MODERATOR) {
 		// Display help for the user
-		if (args.size() <= 1) {
+		if (args.size() < 2) {
 			ChatPackets::SendSystemMessage(sysAddr, u"Mails an item to a specified character");
 			ChatPackets::SendSystemMessage(sysAddr, u"/mailitem <name> <itemid> <count>");
 			return;
@@ -913,6 +913,12 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 			}
 
 			count = std::stoi(args[2]);
+			
+			if (count < 1)
+			{
+				ChatPackets::SendSystemMessage(sysAddr, u"Cannot send less than 1 of an item. Defaulting to 1.");
+				count = 1;
+			}
 		}
 
 		uint64_t currentTime = time(NULL);
@@ -985,6 +991,12 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 			}
 			
 			count = std::stoi(args[2]);
+
+			if (count < 1)
+			{
+				ChatPackets::SendSystemMessage(sysAddr, u"Cannot send less than 1 of an item. Defaulting to 1.");
+				count = 1;
+			}
 		}
 
 		// Mail each character from the player query.
--- a/dGame/dUtilities/SlashCommandHandler.cpp
+++ b/dGame/dUtilities/SlashCommandHandler.cpp
@@ -828,6 +828,7 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 		}
 
 		entity->GetCharacter()->SetMailSender(ss.str());
+		ChatPackets::SendSystemMessage(sysAddr, u"Custom mail sender set.");
 		return;
 	}
 
@@ -844,6 +845,7 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 			ss << string << " ";
 
 		entity->GetCharacter()->SetMailSubject(ss.str());
+		ChatPackets::SendSystemMessage(sysAddr, u"Custom mail subject set.");
 		return;
 	}
 
@@ -860,6 +862,7 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 			ss << string << " ";
 
 		entity->GetCharacter()->SetMailBody(ss.str());
+		ChatPackets::SendSystemMessage(sysAddr, u"Custom mail body set.");
 		return;
 	}
 
--- a/dGame/dUtilities/SlashCommandHandler.cpp
+++ b/dGame/dUtilities/SlashCommandHandler.cpp
@@ -810,7 +810,67 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
         }
     }
 
-	if (chatCommand == "mailitem" && entity->GetGMLevel() >= GAME_MASTER_LEVEL_MODERATOR && args.size() >= 2) {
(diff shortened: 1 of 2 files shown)


dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -810,7 +810,70 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
...
+			ChatPackets::SendSystemMessage(sysAddr, u"Custom sender length is too long. A custom sender name must be 20 characters or less.");
+			return;
+		}
+
+		entity->GetCharacter()->SetMailSender(ss.str());
+		ChatPackets::SendSystemMessage(sysAddr, u"Custom mail sender set.");
+		return;
+	}
+
+	if (chatCommand == "setmailsub" && entity->GetGMLevel() >= GAME_MASTER_LEVEL_DEVELOPER) {
```
MickVermeulen, line 835: I'd prefer the use of an `else if ` chain here

MickVermeulen (changes requested):
Overall I think this looks quite good, have some style changes.

dGame/dUtilities/SlashCommandHandler.cpp
Verathian: Hey Mick!

Thanks for the feedback. Just to make sure, you would like me to do an else if chain for the `setmail____` commands? I only ask because this would be inconsistent with what has been used for other commands in the slashcommandhandler (such as `/setanntitle` and `/setannmsg`, which came by default with the original DLU release. I'm willing to do it, I just wanted to bring up that this particular file usually just uses if statements without and if and else if chains.

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -840,23 +903,135 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
...
+		int count = 1;
+
+		if (args.size() == 3) {
+			if (!GeneralUtils::TryParse(args[2], count))
+			{
+				ChatPackets::SendSystemMessage(sysAddr, u"Invalid item count.");
+				return;
+			}
+
+			count = std::stoi(args[2]);
```
Marcono1234, line 915: Why do you parse the string here again? `count` should already contain the parsed value from the `TryParse` call above, shouldn't it? Same applies to the "mailitemall" command below.
(I am just curious.)

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -840,23 +903,135 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
...
+		if (!GeneralUtils::TryParse(args[1], lot))
+		{
+			ChatPackets::SendSystemMessage(sysAddr, u"Invalid item lot.");
+			return;
+		}
+
+		int count = 1;
+
+		if (args.size() == 3) {
+			if (!GeneralUtils::TryParse(args[1], count))
```
Marcono1234, line 987: Should probably parse `args[2]`, right? `args[1]` is the item.

dGame/dUtilities/SlashCommandHandler.cpp
MickVermeulen: Ah right. Yeah using if-else chains is just neater as we'd never expect the chat command variable to change during execution, leading us to having no use for multiple if statements. You don't have to change the old ones, but it's better to do this going forward imo.

Verathian: Hi Mick,

I have updated all the mail commands to be in an else if chain.

dGame/dUtilities/SlashCommandHandler.cpp
Verathian: You are correct, this should be validating the count. I have updated this is my recent push.

Verathian:
Update: I have updated the permissions in my recent push show the "setmail___" commands match the "mailitem" commands. There is no reason they shouldn't match in permission levels.

dGame/dUtilities/SlashCommandHandler.cpp
MickVermeulen: Good catch, @Verathian did you do this on purpose?

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -840,23 +903,135 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
...
+
 		
-		ChatPackets::SendSystemMessage(sysAddr, u"Mail sent");
+		sql::ResultSet* res = stmt->executeQuery();
+
+		uint32_t lot;
+
+		// Validate the given LOT.
+		if (!GeneralUtils::TryParse(args[1], lot))
+		{
```
MickVermeulen, line 979: Maybe I missed this the first time around, but this should be `if ( ... ) {`, e.g. the bracket on the same line.

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -840,23 +903,135 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
...
+			ChatPackets::SendSystemMessage(sysAddr, u"Invalid item lot.");
+			return;
+		}
+
+		int count = 1;
+
+		if (args.size() == 3) {
+			if (!GeneralUtils::TryParse(args[2], count))
+			{
```
MickVermeulen, line 988: Same here with the same line brackets.

MickVermeulen (changes requested):
Have some style changes (sorry if I missed those the first time around), and also a question that was asked before.

simondumalski:
Been using these changes on a server for a couple weeks now and have experienced no issues whatsoever.

aronwk-aaron:
Could also just use Nexus Dashboard to send mod mail, as there is a feature for this with a nice interface

PR #329, opened 2021-12-31
More windows fixes

Jettford:
- Added World, Chat and Auth servers as deps to Master
- Added "windows-default" CMake preset so that CMake didn't error out on Windows
- Added NOMINMAX define above cpplinq include


Jettford:
Could someone give this a test on Linux?

PR #330, opened 2021-12-31
Fix double storage in LootGenerator

### LINKED ISSUE #274: ENH: Fix double storage in LootGenerator

LootGenerator caches quite a bit of data on startup leading to significantly more memory usage on the WorldServer.

Possibly use references to the existing tables in memory to reduce this memory usage.

_No response_

_No response_

### Additional context

_No response_

TurtleTheSeaHobo:
Fixes #274

**Summary**
Modified `LootGenerator` to lookup necessary tables and data on calling
`RollLootMatrix`, rather than caching them internally, as this avoids
large amounts of data duplication and an avenue of cache incoherency.

**Functionality**
`LootGenerator` is functionally equivalent to before the fix. The private-side layout has changed significantly (all private members removed, new private methods added), however the public interface remains the same.

**Testing**
Tested change and found no significant change in performance in terms of CPU load, though profiled memory usage for `LootGenerator` is effectively zero (profiler still caught transient vector allocs in functions, but these were present pre-fix as well).

In testing, change does not appear to have altered loot rolling/dropping behavior at all. None of the changes have modified the rolling algorithm or underlying loot tables. Only the method of data access has been changed.

**Notes**
By querying the loot tables on a per-call basis, this change also ensures that the loot rolling mechanics always reflect the underlying tables in the database. Though the tables are currently constant, the caching adds a bit of pesky indirection and would lead to unintuitive behavior if the loot tables ever were to change at runtime (perhaps needed by a future feature request). 

Please let me know of any concerns with this pull request and I'd be happy to resolve them. This is my first serious PR to a big repo, so I want to do it right.

### DIFF
--- a/dChatServer/ChatPacketHandler.cpp
+++ b/dChatServer/ChatPacketHandler.cpp
@@ -351,6 +351,13 @@ void ChatPacketHandler::HandleTeamInvite(Packet* packet)
 		return;
 	}
 
+	if (team->memberIDs.size() > 3) {
+		// no more teams greater than 4
+
+		Game::logger->Log("ChatPacketHandler", "Someone tried to invite a 5th player to a team");
+		return;
+	}
+
 	SendTeamInvite(other, player);
 
 	Game::logger->Log("ChatPacketHandler", "Got team invite: %llu -> %s\n", playerID, invitedPlayer.c_str());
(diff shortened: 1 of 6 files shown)


Jettford:
This does remove the issue of the cache storing the double (or more) entries but removing it is not the way to do that. This slows down the server a lot as every time there is a loot drop it runs through a lot more data. The solution would be to when the data is cached check if any items exist that have identical values, that is all that you need to do.

TurtleTheSeaHobo:
I see. Your certainly right about the increased load on drops. It didn't seem to hit performance very noticeably in my testing (which, granted, was just me and few friends running around smashing every destructible we could find), but I could see that potentially scaling up poorly. I'll look into the de-duplication approach as you suggest.

Since this entails an entirely different change to the code, should I just close this PR?

Issue #328, opened 2021-12-31
BUG: Enemies become invincible from Samurai buff


IAmVolvic:
### Make sure you've done the following:

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### Issue Description

When having a full-set Samurai gear no matter the level, the buff " When you reach 0 Armor,  nearby characters will be shielded from 15 damage for a short time " Issue is it doesn't wear off so the enemies become invincible. 

This is an issue when you're just playing the game in general and you can no longer kill anything near you

### Reproduction steps

Equip a set of 4 piece samurai set lose some armor next to an enemy and wait till the buff wears off and then you can no longer kill the enemy

### Expected Behavior

When buff wears off the enemy/enemies buff should continue to take damage ( or how live did it ) not 100% sure

### Environment

Local Server 

Distributor ID: Ubuntu
Description: Ubuntu 20.04.3 LTS
Release: 20.04
Codename: focal

Mysql: Ver 8.0.27-0ubuntu0.20.04.1 for Linux on x86_64 ((Ubuntu))

Ram: 128

Supermicro 8-Bay 2xE5-2670v3

heatraydeathgun:
We've also been seeing this. Nearby enemies receive permanent (so far in testing) invincibility from the Samurai 4-item passive triggering, while intended function is to grant only surrounding players protection from X amount of damage for a short time after reaching 0 armor.

metric-ambient:
My friend experience this with the Samurai gear as well.

MichaelFedora:
duplicate of https://github.com/DarkflameUniverse/DarkflameServer/issues/211 I believe

averysumner:
Duplicate of #211.

PR #339, opened 2022-01-02
Fix Area of Effect Behavior Targets

CodeAX2:
**Description:**
This fixes how entities are targeted during Area of Effect Behaviors. Each AOE Behavior has fields set in the CDClient database to determine who to target. The fields are read in on initialization and are used to filter targets. This fixes issues such as #328 

**Motivation/Context:**
Passive abilities, such as the Samurai's invulnerability at 0 armor, and other in-game effects are handled as an area of effect. The effect targets specific entities based on fields outlined in the CDClient Database. Some behaviors target enemies (such as the buildable sirens in Gnarled Forest), while others target friends (such as the aforementioned Samurai ability). However, there was no logical distinction, and so abilities that should target friends were targeting enemies.

**Types of Changes:**
This is a simple change to the definitions in the area of effect behavior script to read fields from the database related to targeting, as well as changes to destroyable components to allow for validating a target based on these fields.

**Testing:**
This has been tested using multiple players in a variety of situations and behaviors, such as the Sirens and Samurai abilities. Sirens should target enemies (which they do), whereas Samurais should target friends (which they also do)

**Screenshots:**
Below are two screenshots showing the basic functionality.

(1) The siren successfully targets enemies (stromlings) and not friends (players)
Lego Universe Screenshot 2022 01 02 - 14 22 53 95

(2) The Samurai ability does not target enemies (stromlings), but is in effect (via the glowing orange around the player)
Lego Universe Screenshot 2022 01 02 - 14 23 22 60

### DIFF
--- a/dGame/dComponents/DestroyableComponent.cpp
+++ b/dGame/dComponents/DestroyableComponent.cpp
@@ -498,7 +498,7 @@ Entity* DestroyableComponent::GetKiller() const
 	return EntityManager::Instance()->GetEntity(m_KillerID);
 }
 
-bool DestroyableComponent::CheckValidity(const LWOOBJID target, const bool ignoreFactions) const
+bool DestroyableComponent::CheckValidity(const LWOOBJID target, const bool ignoreFactions, const bool targetEnemy, const bool targetFriend) const
 {
 	auto* entity = EntityManager::Instance()->GetEntity(target);
 
@@ -537,15 +537,19 @@ bool DestroyableComponent::CheckValidity(const LWOOBJID target, const bool ignor
 
 	auto candidateList = destroyable->GetFactionIDs();
 
+
+	bool isEnemy = false;
+
 	for (auto value : candidateList)
 	{
 		if (std::find(enemyList.begin(), enemyList.end(), value) != enemyList.end())
 		{
-			return true;
+			isEnemy = true;
+			break;
 		}
 	}
 
-	return false;
+	return (isEnemy && targetEnemy) || (!isEnemy && targetFriend);
 }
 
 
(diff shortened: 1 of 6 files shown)


averysumner:
Looks good to me, requested review from people who know more about behaviors than me.

Xiphoseer (approved):
Looks good to me

Jettford (approved):

dGame/dComponents/DestroyableComponent.cpp
```diff
@@ -533,19 +532,12 @@ bool DestroyableComponent::CheckValidity(const LWOOBJID target, const bool ignor
...
-		{
-			return true;
-		}
-	}
+	// Get if the target entity is an enemy
+	bool isEnemy = IsEnemy(targetEntity);
 
-	return false;
+	// Return true if the target type matches what we are targeting
+	// Friends are entities who are not enemies
```
Xiphoseer, line 539: This isn't actually true. There is a separate column in the Factions table that we currently don't use: `friendList`

I'm not certain about the semantics, but I'd suggest you call `DestroyableComponent::IsFriend` as well and we can fix its implementation in a follow up PR.

Xiphoseer (changes requested):
While we're still waiting for a review from @Wincent01 who built this implementation, there was another question raised internally about our `IsFriend` being wrong, which you don't have to fix in this PR but it would be nice to prepare for that at least.

dGame/dComponents/DestroyableComponent.cpp
CodeAX2: Okay I see. Originally I just had the one call to IsEnemy since also calling IsFriend takes a bit more time, but since IsFriend is planned to be changed in the future, I have updated the code to determine IsEnemy and IsFriend separately.

Xiphoseer (approved):

IAmMajo:
I have had this on my server with around 40 players since January 5th and it works perfectly.

PR #344, opened 2022-01-04
Fix Various Issues With Powerup Spawners

CodeAX2:
**Description:**
Powerup spawners have been fixed to properly drop imagination, armor, etc. once more. Their timings have also been fixed to better mirror the intended behavior.

**Motivation/Context:**
Powerup spawners (such as the anvil of armor, fountain of imagination, etc.) did not spawn any powerups. They would also disappear after 1 minute, rather than after their cycles of spawning. This PR fixes both issues, as now they spawn their proper powerups, and follow the intended behavior, of cycling their spawning for a set number of seconds for a set number of cycles, and then destroying themselves after another delay once the final spawning has finished.

**Types of Changes:**
Basic changes to code that is relevant. Including...
- Fix to order of updating timers in entities to ensure new timers are not updated in the current tick, and no timers are skipped over
- Change to spawn behavior to only apply the 60-second limit if the entity is not a powerup spawner
- A fix to loot dropping to properly consider missions only if they are available
- A change to powerup spawners to directly set the LOTs dropped

*Testing:*
In-game testing of the 3 power-up spawners, including timing between spawning powerups and self-destruction, as well as testing that the regular loot tables still act properly.

*Screenshots:*
(1) Fountain of Imagination Dropping Powerups:
Lego Universe Screenshot 2022 01 03 - 20 08 11 42

(2) Correct Amount of Imagination Acquired (Difference of +5)
Lego Universe Screenshot 2022 01 03 - 20 08 19 06

(3) Regular Loot Dropping for Enemies
Lego Universe Screenshot 2022 01 03 - 20 08 31 51


dGame/dBehaviors/SpawnBehavior.cpp
```diff
@@ -74,9 +76,25 @@ void SpawnBehavior::Handle(BehaviorContext* context, RakNet::BitStream* bitStrea
 		context->RegisterEndBehavior(this, branch, entity->GetObjectID());
 	}
 
-	entity->AddCallbackTimer(60, [entity] () {
-		entity->Smash();
-	});
+	// Determine if the entity spawns powerups
+	bool isPowerupSpawner = false;
```
Xiphoseer, line 80: The logic here is special-casing powerup spawners but nothing else. That indicates to me that the "real" fix is elsewhere.

dGame/dUtilities/Loot.cpp
```diff
@@ -137,10 +137,6 @@ std::unordered_map<LOT, int32_t> LootGenerator::RollLootMatrix(Entity* player, u
 
     std::unordered_map<LOT, int32_t> drops;
 
-    if (missionComponent == nullptr) {
-        return drops;
-    }
-
```
Xiphoseer, line 143: Does it work without this change, given the comment by @Wincent01 on #313?

Xiphoseer (commented):
Added some questions

dGame/dUtilities/Loot.cpp
CodeAX2: I'll have to test without it. It seems to me that either way it is probably safer to have the checks for mission components being null to only affect the parts of loot that require it. Through testing I did confirm that this part did not affect the drops for powerup spawners, but figured I would keep it since it seemed like a safer option to me.

dGame/dBehaviors/SpawnBehavior.cpp
CodeAX2: Right, this is a bit case specific. Though it seems that the rest of the code within this function was still vital to having powerup spawners working. Without catching the edge case, the callback timer will always be added to the powerup spawner, thus destroying it after 60 seconds. 
Are there any cases where the callback is fully required? I assume probably to ensure objects are destroyed at some point, but this is a simple 60 seconds for any spawn behavior. 
The way I see it, either the handle() method needs more information to better determine how long until an object should be destroyed, or perhaps the callback should be cancelled from the powerup spawner itself. Though I don't see any way to stop a specific callback, since no handle is returned or stored, only all current callbacks on an entity can be cancelled (which is obviously problematic)

Hopefully that makes sense!

Xiphoseer: Yeah it totally makes sense, which is why I'd want to steer this review to figuring that out. We'll probably have to wait for @Wincent01 to give you feedback here, because he and Mick built most of the scripts.

metric-ambient:
I tested this on my server and it fixes the fountain, anvil, and cauldron, but does not fix the Hat of Sprouting Imagination.

EmosewaMC:
This has been resolved and works in live.  Closing.

PR #231, opened 2021-12-14
Aarch64 port

jumpy-badger:
Summary of changes:
* added aarch64 defines and checks to raknet library
* replaced sql connector with mariadb connector cpp
* * Windows will use the prebuilt binaries from the website but can build it from source if desired (with msvc anyway).
* * Some changes to code (like setting the environment variable) were so that installing mariadb connector manually didn't have to be done
* change in dGame/LeaderboardManager.cpp was because of a compiler error on my system.
* A few changes because mariadb connector doesn't have asStdString() as part of it's SQLString class

Tested and compiled on Windows 10 x64 and Raspberry Pi 4 8gb (Pi OS 64 bit)

PR #345, opened 2022-01-04
Improve CI

Xiphoseer:
Changes in this PR:

- Fix a cmake warning
- Add a CMakeLists for thirdparty
- Use `add_library` for sqlite
- Move some compiler flags closer to where they matter


Xiphoseer:
This PR should warn about some unused `std::async` results which were previously hidden and that's actually correct as:
> If the std::future obtained from std::async is not moved from or bound to a reference, the destructor of the std::future will block at the end of the full expression until the asynchronous operation completes.

c.f. https://en.cppreference.com/w/cpp/thread/async#Notes

Xiphoseer:
Additionally: Can someone try to compile this on windows and/or Mac?

MickVermeulen:
Can try on a mac, no earlier than Friday, though.

MickVermeulen (commented):
Alright so this all looks good to me but I'd prefer merging #231 first and then updating this PR to work with that, given that the ARM port has been in the running for quite a while.

Xiphoseer:
> Alright so this all looks good to me but I'd prefer merging #231 first and then updating this PR to work with that, given that the ARM port has been in the running for quite a while.

Sure, I'm fine with that

Jettford:
@Xiphoseer Some merge conflicts have popped up, appears to be because of your windows-ci changes

Xiphoseer:
> @Xiphoseer Some merge conflicts have popped up, appears to be because of your windows-ci changes

Yeah, I know, just haven't gotten around to fixing it

Jettford:
Interesting the Ubuntu and Windows CI build appears to have failed with no seeming error?

EmosewaMC:
> This PR should warn about some unused `std::async` results which were previously hidden

Should we add an assignment to these asyncs so they actually run on a different thread as opposed to now where they basically run synchronous with the main thread?

aronwk-aaron:
@Xiphoseer this looks good to merge, are there any blockers for this?

EmosewaMC:
> @Xiphoseer this looks good to merge, are there any blockers for this?

This is waiting on #231 and it is also not complete.

Xiphoseer:
> @Xiphoseer this looks good to merge, are there any blockers for this?

Micks comment still applies, but unless there are merge conflict, there's mothing specifically blocking from my side.

EmosewaMC:
> 

This appears to be a bug with GCC.  with MSVC these threads properly run async.  Pushing a fix to main.

codeshaunted:
> > If the std::future obtained from std::async is not moved from or bound to a reference, the destructor of the std::future will block at the end of the full expression until the asynchronous operation completes.
> 
> 
> 
> This appears to be a bug with GCC.  with MSVC these threads properly run async.  Pushing a fix to main.

I'm guessing it's a non-standard implementation by MSVC. This note is from cppreference, not specifically pertaining to GCC.

EmosewaMC:
> > > If the std::future obtained from std::async is not moved from or bound to a reference, the destructor of the std::future will block at the end of the full expression until the asynchronous operation completes.
> > 
> > 
> > This appears to be a bug with GCC.  with MSVC these threads properly run async.  Pushing a fix to main.
> 
> I'm guessing it's a non-standard implementation by MSVC. This note is from cppreference, not specifically pertaining to GCC.

Ah my mistake. Still something should be fixed so as to not halt threads that can cause the client to disconnect.

codeshaunted:
> > > > If the std::future obtained from std::async is not moved from or bound to a reference, the destructor of the std::future will block at the end of the full expression until the asynchronous operation completes.
> 
> > > 
> 
> > > 
> 
> > > This appears to be a bug with GCC.  with MSVC these threads properly run async.  Pushing a fix to main.
> 
> > 
> 
> > I'm guessing it's a non-standard implementation by MSVC. This note is from cppreference, not specifically pertaining to GCC.
> 
> 
> 
> Ah my mistake. Still something should be fixed so as to not halt threads that can cause the client to disconnect.

Yeah for sure, we don't want to block the main thread.

dGame/dUtilities/Mail.cpp
```diff
@@ -397,7 +398,8 @@ void Mail::HandleMailRead(RakNet::BitStream* packet, const SystemAddress& sysAdd
 }
 
 void Mail::HandleNotificationRequest(const SystemAddress& sysAddr, uint32_t objectID) {
-	std::async(std::launch::async, [&]() {
+	// We need to assign the result of this async to a variable so the thread here isnt blocking.
+	auto futureValue = std::async(std::launch::async, [&]() {
```
Xiphoseer, line 402: FYI: I don't think this variable improves things much. The issue is that `std::async  returns a `std::future` whose destructor will wait for the thread/task to complete according to the box in <https://en.cppreference.com/w/cpp/thread/future/%7Efuture> (seems to be a C++ 14 thing?)

Xiphoseer (commented):
FWIW I think it'd be more appropriate not to have code changes in this PR because it adds more stuff to review.

dGame/dUtilities/Mail.cpp
EmosewaMC: thats fair.  I didnt mean to send those to this branch for what its worth.  Meant for this to go into its own branch.  Ill address that tomorrow.  Ill also look up a way to make this properly async on all platforms.

EmosewaMC:
Tested that servers on Windows AArch64 and WSL still build and boot correctly on a fresh build and they still do build and allow users to connect and login which should be enough testing.  Merging

PR #347, opened 2022-01-04
Remove Friend Fix

### LINKED ISSUE #267: BUG: Remove friend doesn't remove friend

### Issue Description

Using remove friend does nothing. No friend is removed or any changes to the database.

### Reproduction steps

- Get a friend or use a second account
- Add them as a friend
- Try to remove them

### Expected Behavior

The "friend" is removed.

### Environment

WSL 1 - Ubu 20.04

EmosewaMC:
Fixes #267 
The function for remove friend is working as intended, however I noticed that the function was not properly getting the LWOOBJID for one of the queries, and as such was unable to perform the necessary work to properly remove a friend.

First, the character name that was being fetched from the packet was using the incorrect offset and was changed to be the proper offset to get the name correctly. Then the conversion to a LWOOBJID was corrected from ClearBit to SetBit to get the correct ID for querying the table. I have also cleaned up the function a bit to reduce unnecessary work.

I have tested this in my personal instance and this does remove multiple friends from the client, database and the friends client.


Xiphoseer:
This might conflict with #311

EmosewaMC:
> This might conflict with #311

Looks like it does however, the lines it conflicts with can be safely overwritten.  The lines they are changing are simply info for querying the database.  Will test their changes on my end to make sure there are no conflicts @Xiphoseer

EmosewaMC:
when #311 is merged I will update this PR with the proper changes.

EmosewaMC:
@Xiphoseer since #311 has been put to the side for now could we get this one merged?  If needed I can test again to confirm this works but it should still be functional without issues

PR #354, opened 2022-01-07
Sanitize SQL Queries to CDClient.sqlite

TheMatt2:
## Description
Unlike the connection to the main mysql database, many sql queries to the sqlite database are created by
directly concatenating stings. This poses a security concern, source of crashes, and general instability.

Specifically, the usage the `/lookup` command could crash the server from poorly formed user input.

The sanitation issue is solved by creating a new function:

```CDClientDatabase::ExecuteQueryWithArgs()```

https://github.com/TheMatt2/DarkflameServer/blob/e5f7d164cbf958cabd6f6f7da9cf56356e565c18/dDatabase/CDClientDatabase.h#L50

This function makes use of sanitation facilities in the third party Sqlite3 library to sanitize the values.
This is a template style function to make it as easy to use as possible for developers.

All other changes are adding the `CDClientDatabase::ExecuteQueryWithArgs()` function to replace unsafe usages of creating sqlite queries.

For the three cases where a string was inputted into the sqlite query, this has been tested by running the old code and new code side by side, and confirming the generated query match as expected.
Those three cases are as follows:
https://github.com/TheMatt2/DarkflameServer/blob/e5f7d164cbf958cabd6f6f7da9cf56356e565c18/dGame/dUtilities/SlashCommandHandler.cpp#L1248
https://github.com/TheMatt2/DarkflameServer/blob/e5f7d164cbf958cabd6f6f7da9cf56356e565c18/dGame/dBehaviors/Behavior.cpp#L355
https://github.com/TheMatt2/DarkflameServer/blob/e5f7d164cbf958cabd6f6f7da9cf56356e565c18/dGame/dComponents/RenderComponent.cpp#L202

For other cases, where the sql query concatenated a number, which doesn't directly pose a security risk, they were tested by the same side by side test on sample of the changes. The rest were checked by eye to see they looked right.

All tests were performed using a build on Ubuntu.

## Other Changes
Also fixes a validation issue with the `/runmacro` command. Now it checks the input data to prevent a user from running a script that is outside of the `macro` folder.

https://github.com/TheMatt2/DarkflameServer/blob/e5f7d164cbf958cabd6f6f7da9cf56356e565c18/dGame/dUtilities/SlashCommandHandler.cpp#L615

This also adds a limit to the number of results `/lookup` can show. Without it, it is possible for the Client to be lagged out for a long period of time, as the server sends chat messages to it. Lookup results are arbitrarily capped at 50.

Note this does _not_ fix # 350 
https://github.com/TheMatt2/DarkflameServer/blob/e5f7d164cbf958cabd6f6f7da9cf56356e565c18/dGame/dUtilities/SlashCommandHandler.cpp#L1248

## Points of Contention
As one may not, all strings passed to `CDClientDatabase::ExecuteQueryWithArgs()` are c-style strings. This is because passing a `std::string` will silently fail, and the SQL query that is generated will be corrupted. It would be nice to add a compile time check to prevent this, if someone knows how.

I know the PR points say to either make functional changes or whitespace changes, not both. However, it seems there will be some whitespace changes due to the way the editor operates. Apologize.

I would another set of eyes on this, and make sure that none of the changes somehow break the SQL query in a way I haven't noticed.

### CODE REVIEW

**Review by Xiphoseer — CHANGES_REQUESTED** [review]
I don't think it's a good idea to merge this as-is. You used [`CppSQLite3Buffer`](https://github.com/DarkflameUniverse/DarkflameServer/blob/a9860388b2b34d4a52c63c673d0e4a94ec089ebc/thirdparty/SQLite/CppSQLite3.h#L67) (which seems to be *just* a wrapper around [`sqlite3_vmprintf`](https://www.sqlite.org/c3ref/mprintf.html)) over c++ stringstreams. I can't see any improvement in security there.

On the other hand, there's [`CppSQLite3Statement`](https://github.com/DarkflameUniverse/DarkflameServer/blob/a9860388b2b34d4a52c63c673d0e4a94ec089ebc/thirdparty/SQLite/CppSQLite3.h#L231) which wraps the [Prepared Statement Object](https://www.sqlite.org/c3ref/stmt.html) API
with its `bind` functions. This would solve the issue properly IMO.

**Comment by TheMatt2** [comment]
I think using `CppSQLite3Buffer` is better than c++ stringstreams because then the SQL query is known safe. This is particularly important for the queries that include adding strings, as it prevents the server from crashing.

Now, specifically for adding only numbers to a query, I agree that in practice there is know difference.
But the point is using the buffer _known_ safe, instead of relying on the programmer to use stringstream correctly to avoid an issue.

That said, I think you are right [`CppSQLite3Statement`](https://github.com/DarkflameUniverse/DarkflameServer/blob/a9860388b2b34d4a52c63c673d0e4a94ec089ebc/thirdparty/SQLite/CppSQLite3.h#L231) offers a better solution. It would avoid a templated function and prevent the issue with `std::string` handling.

This also means the sqlite queries can share a format with the mysql queries.

**Comment by Xiphoseer** [comment]
Right, I only skimmed over the definition of sqlite_vmprintf and seemed to have missed the most important point, sorry.

So this function generates a "known safe" statement as you call it, if it's used with numbers as you mentioned and if (and only if) `%Q` is used for string parameters (or %q within quotes).

In that case you should cleary indicate that in the header file where your wrapper function is defined. Otherwise, it kind of defeats the purpose by risking that further updates just use `%s` because they don't look at the implementation details and that still "works".

Additionally, using varargs in C++ seems like a bit of a footgun, especially if as you say it silently fails on `std::string`s.

I guess for the future it'd be nice to have been given some reason how to check why you might be right in claiming safety for CppSqlite3Buffer in this PR, especially given the total lack of documentation on that class in its header file.

**Comment by Xiphoseer** [comment]
Fo clarify, I got the impression that the sqlite mprintf functions differed from C in that they automatically allocated memory (hence the m) and not that

> plus some additional non-standard formats (%q, %Q, %w, and %z)

would be a security feature

**Comment by TheMatt2** [comment]
Apologize not explaining clearly why `CppSQLite3Buffer` represented a way of providing security.
The _main_ concern is this would prevent crashes that would be difficult to solve. With the current implementation by running the command such as `/lookup ')` will cause an error on processing the SQL statement, and crash the server.

Regardless, [`CppSQLite3Statement`](https://github.com/DarkflameUniverse/DarkflameServer/blob/a9860388b2b34d4a52c63c673d0e4a94ec089ebc/thirdparty/SQLite/CppSQLite3.h#L231). Seems to provide a safer solution and avoids the "footgun" while properly validating input.

**Comment by TheMatt2** [comment]
Sorry. Please don't review yet.
Still working out some kinks :-)

**Comment by TheMatt2** [comment]
Fixed a bug in the corrections.
Turns out the Sqlite3 results object holds a reference to the prepared statement. So if the prepared statement leaves scope while the results are still being used, the server would crash. whoops

That should all be worked out now.

**Comment by TheMatt2** [comment]
@Xiphoseer 
Is this ready to merge?

File: dDatabase/CDClientDatabase.h [hunk 1]
```diff
@@ -40,4 +40,10 @@ namespace CDClientDatabase {
      */
     CppSQLite3Query ExecuteQuery(const std::string& query);
     
+    //! Queries the CDClient and parses arguments
+    /*!
+      \param query The query with formatted arguments
+      \return prepared SQLite Statement
+    */
+    CppSQLite3Statement CreatePreppedStmt(const std::string& query);
```
Xiphoseer, line 48: I'd personally prefer `CreatePreparedStmt`

Xiphoseer (approved):
LGTM, would like one additional sign-off though

dDatabase/CDClientDatabase.h
TheMatt2: My thinking was to use the same name as the corresponding mysql function
https://github.com/DarkflameUniverse/DarkflameServer/blob/3de393250340ad27aa2dcac39afe9ed93c114959/dDatabase/Database.h#L27

Do you want me to change this?

Xiphoseer: No, that's a pretty good reason to keep it like this

TheMatt2 (commented):
Cleaned up whitespace

Now all added code uses tabs and fixes some of the existing inconsistency.

dDatabase/Tables/CDBehaviorParameterTable.cpp
```diff
@@ -59,11 +59,11 @@ float CDBehaviorParameterTable::GetEntry(const uint32_t behaviorID, const std::s
 	}
 
 #ifndef CDCLIENT_CACHE_ALL
-	std::stringstream query;
+	auto query = CDClientDatabase::CreatePreppedStmt(
+        "SELECT parameterID, value FROM BehaviorParameter WHERE behaviorID = ?;");
+    query.bind(1, (int) behaviorID);
```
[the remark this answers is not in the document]
TheMatt2: Updated

averysumner:
Why are we sanitizing queries to the CDClient? Generally that file will be read once into an vectors or maps and then never used again… besides, only the server ever interacts with CDClient. Additionally, please don’t make multiple unrelated changes in a single pull request.

TheMatt2:
The point is remove and prevent SQL injection.

While the security risk of SQL injection is limited, CDClient is simply data that is already in the client, these can still crash the server.
This is being fixed by properly sanitizing the SQL statements.

While, admittingly, concatting a number is probably safe, it is my understanding that it is good practice to always use a sanitization function when making SQL statements.

Apologize on the combining with whitespace changes. I'll refrain from doing that in the future.

Xiphoseer:
FWIW, I agree with this. Would be nice to have this extra layer of redundancy afainst manipulations to the CDServer by SQL injection.

TheMatt2:
Merged and retested.

EmosewaMC:
This has merge conflicts - Could you please resolve them?

aronwk-aaron:
Resolved conflicts

EmosewaMC (approved):

EmosewaMC:
Will remind once again to disable auto linting in the future and also to only make changes the PR is focussed around.  The runmacro changes could have easily been their own PR and been merged in 1 hour but instead have been lumped in with this and took a long time to merge.  

Merging because this all looks fine to me after a review.

PR #356, opened 2022-01-08
Added UNIQUE to charInfo table for names

### LINKED ISSUE #321: ENH: Add UNIQUE constraint to `charinfo.name` database column

It looks like the `charinfo.name` database column currently has no `UNIQUE` constraint:
https://github.com/DarkflameUniverse/DarkflameServer/blob/0a7d9c46eba551b5e2f908caf9c689f4139bffb4/migrations/dlu/0_initial.sql#L18

But uniqueness seems to be enforced in the code here:
https://github.com/DarkflameUniverse/DarkflameServer/blob/0a7d9c46eba551b5e2f908caf9c689f4139bffb4/dGame/UserManager.cpp#L263-L266

Would it make sense to add a `UNIQUE` constraint, or is there something I am overlooking?

_No response_

_No response_

### Additional context

_No response_

EmosewaMC:
Added UNIQUE for charInfo table for names.  Tested creating characters with the same name and had no issues.  Fixes #321 suggestion

### DIFF
--- a/migrations/dlu/0_initial.sql
+++ b/migrations/dlu/0_initial.sql
@@ -15,7 +15,7 @@ DROP TABLE IF EXISTS charinfo;
 CREATE TABLE charinfo (
     id BIGINT NOT NULL PRIMARY KEY,
     account_id INT NOT NULL REFERENCES accounts(id),
-    name VARCHAR(35) NOT NULL,
+    name VARCHAR(35) NOT NULL UNIQUE,
     pending_name VARCHAR(35) NOT NULL,
     needs_rename BOOLEAN NOT NULL DEFAULT FALSE,
     prop_clone_id BIGINT UNSIGNED AUTO_INCREMENT UNIQUE,
--- a/migrations/dlu/0_initial.sql
+++ b/migrations/dlu/0_initial.sql
@@ -15,7 +15,7 @@ DROP TABLE IF EXISTS charinfo;
 CREATE TABLE charinfo (
     id BIGINT NOT NULL PRIMARY KEY,
     account_id INT NOT NULL REFERENCES accounts(id),
-    name VARCHAR(35) NOT NULL UNIQUE,
+    name VARCHAR(35) NOT NULL,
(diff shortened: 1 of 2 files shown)


cooltrain7:
I wonder if it would be a good idea to add a DB template check during startup to validate each of the tables/columns/constraints and update them accordingly (or at a minimum log that changes are needed). Otherwise you're relying on people to spot commits that effect the DB and apply them. Not just for this commit but for others in the future that will add new columns.

Xiphoseer:
Yeah, technically, this should be a new migration file `1_unique_charinfo_names.sql` (or similar) with
```sql
ALTER TABLE charinfo ADD UNIQUE (name); 
```

EmosewaMC:
> Yeah, technically, this should be a new migration file `1_unique_charinfo_names.sql` (or similar) with
> 
> ```sql
> ALTER TABLE charinfo ADD UNIQUE (name); 
> ```

Gotcha.  Will update the PR (and the README?) to reflect this change.

Jettford:
> I wonder if it would be a good idea to add a DB template check during startup to validate each of the tables/columns/constraints and update them accordingly (or at a minimum log that changes are needed). Otherwise you're relying on people to spot commits that effect the DB and apply them. Not just for this commit but for others in the future that will add new columns.

@cooltrain7 Maybe instead of a schema checker maybe there should be a migration runner added to the server, so instead of checking the schema just run migrations that aren't listed as merged in a new table.

Marcono1234:
Thanks a lot @EmosewaMC for addressing the issue I raised!

In case you are not aware of it, GitHub supports certain [keywords for pull requests](https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests). This way GitHub automatically links pull request and issue, and closes the issue when the pull request is merged (maintainers can also manually add the link if necessary). It should also be possible to edit the description of your pull request retroactively and add the line "Resolves #321" to it.

PR #357, opened 2022-01-08
Fixed an issue where repeatable missions were not rewarding the items they should be rewarding

EmosewaMC:
Small change that fixes an off by one error for repeatable mission item rewards

Also: Clarified variable names for Wisp Lee script to match what they are in the NpcWispServer.cpp file

### DIFF
--- a/dGame/dMission/Mission.cpp
+++ b/dGame/dMission/Mission.cpp
@@ -438,10 +438,13 @@ void Mission::YieldRewards() {
         items.emplace_back(info->reward_item4_repeatable, info->reward_item4_repeat_count);
 
         for (const auto& pair : items) {
+            // Some missions reward zero of an item and so they must be allowed through this clause,
+            // hence pair.second < 0 instead of pair.second <= 0.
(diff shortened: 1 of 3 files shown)


dGame/dMission/Mission.cpp
```diff
@@ -438,7 +438,7 @@ void Mission::YieldRewards() {
         items.emplace_back(info->reward_item4_repeatable, info->reward_item4_repeat_count);
 
         for (const auto& pair : items) {
-            if (pair.second <= 0 || (m_Reward > 0 && pair.first != m_Reward)) {
+            if (pair.second < 0 || (m_Reward > 0 && pair.first != m_Reward)) {
```
Xiphoseer, line 441: Can you add a comment (to the source) explaining why this condition is correct?

EmosewaMC: Sure

EmosewaMC:
This has had no issues on my end from any tests.  Should be ready to merge @Xiphoseer

Xiphoseer (approved):
LGTM

PR #358, opened 2022-01-08
Added Frostburgh into Starbase Interface

PyEddy:
This adds a way for players to launch themselves to Frostburgh without needing to TP (assuming the user has fixed Frostburgh on their CDServer.sqlite), it adds an entry to the LUP interface (in Starbase 3001).

### DIFF
--- a/dScripts/SbLupTeleport.cpp
+++ b/dScripts/SbLupTeleport.cpp
@@ -116,6 +116,28 @@ void SbLupTeleport::OnStartup(Entity* self)
         choiceOptions->PushBackValue(ntArgs);
     }
 
+    {
+        AMFArrayValue* nsArgs = new AMFArrayValue();
+
+        AMFStringValue* image = new AMFStringValue();
+        image->SetStringValue("textures/ui/zone_thumnails/Deep_Freeze.dds");
+        nsArgs->InsertValue("image", image);
+
+        AMFStringValue* caption = new AMFStringValue();
+        caption->SetStringValue("%[ZoneTable_98_DisplayDescription]");
+        nsArgs->InsertValue("caption", caption);
+
+        AMFStringValue* identifier = new AMFStringValue();
+        identifier->SetStringValue("zoneID_98");
+        nsArgs->InsertValue("identifier", identifier);
+
+        AMFStringValue* tooltipText = new AMFStringValue();
+        tooltipText->SetStringValue("%[ZoneTable_98_summary]");
+        nsArgs->InsertValue("tooltipText", tooltipText);
+
+        choiceOptions->PushBackValue(nsArgs);
+    }
+
     args.InsertValue("options", choiceOptions);
 }
 


Jettford:
This is a bad idea for numerous reasons, it isn't live accurate, Frostburg's missions are pulled out in 1.10.64, the launch pad is in the wrong place (you can't leave), racing doesn't work and it requires extra edits to not just put people into the void.

PyEddy:
> This is a bad idea for numerous reasons, it isn't live accurate, Frostburg's missions are pulled out in 1.10.64, the launch pad is in the wrong place (you can't leave), racing doesn't work and it requires extra edits to not just put people into the void.

You can leave, I tested it, just place your rocket from models in your backpack onto the launchpad and you will be sent into Nimbus Station

But yeah in terms of missions, that's a problem

Jettford:
> You can leave, I tested it, just place your rocket from models in your backpack onto the launchpad and you will be sent into Nimbus Station

You can leave if you have configured the world which no one without development knowledge can do

PyEddy:
I was discussing on discord of potentially a way this could be toggled on depending on the player, if they have access to Frostburgh and want to get there without needing to TP then they could toggle something in one of the config files to enable this addition. It would be off by default tho as not everyone has access to Frostburgh

Jettford:
and this reference the Deep Freeze thumbnail not a frost burgh one.

PyEddy:
Yeah, there was no Frostburgh thumbnail found in the client so I used DeepFreeze's thumbnail as its similar (they are both snow maps)

Jettford:
> I was discussing on discord of potentially a way this could be toggled on depending on the player, if they have access to Frostburgh and want to get there without needing to TP then they could toggle something in one of the config files to enable this addition. It would be off by default tho as not everyone has access to Frostburgh

I don't understand why adding this as an extra config option to be completely honest. If you wanted to add a feature to allow adding maps to rocket launchers but I see no point in hardcoding a Frostburgh one that uses the wrong thumbnail

Jettford:
> Yeah, there was no Frostburgh thumbnail found in the client so I used DeepFreeze's thumbnail as its similar (they are both snow maps)

Frostburgh is quite different

PyEddy:
> > Yeah, there was no Frostburgh thumbnail found in the client so I used DeepFreeze's thumbnail as its similar (they are both snow maps)
> 
> Frostburgh is quite different

Yeah I am aware, couldn't find its thumbnail in the client tho, might try looking deeper, I just went with the Deep Freeze thumbnail as its snow so I didn't think it would be a problem

Jettford:
> Yeah I am aware, couldn't find its thumbnail in the client tho, might try looking deeper, I just went with the Deep Freeze thumbnail as its snow so I didn't think it would be a problem

There isn't one in the client afaik.

PyEddy:
> > Yeah I am aware, couldn't find its thumbnail in the client tho, might try looking deeper, I just went with the Deep Freeze thumbnail as its snow so I didn't think it would be a problem
> 
> There isn't one in the client afaik.

Yeah, well atleast the code for adding Frostburgh so that its accessible to everyone on the server is there now, I kinda expected to receive these comments after I sent the PR so probably it wasn't a really good idea to send one, well if someone manages to add in Frostburgh into their server, this code could make it simple for their friends to go to Frostburgh

Jettford:
And also, that locale reference isn't present in the 1.10.64 client

PyEddy:
> And also, that locale reference isn't present in the 1.10.64 client

Yep, that is true

PyEddy:
This would be good for those who have DLU clients, not sure about people using lcdr's client

A lot of problems would occur

Jettford:
> This would be good for those who have DLU clients, not sure about people using lcdr's client
A lot of problems would occur

The DLU Client was never made public and code in the server shouldn't be assuming that it is

PyEddy:
> The DLU Client was never made public and code in the server shouldn't be assuming that it is

Yep, I tested this on both clients however (I will say my lcdr client is modified tho)

PyEddy:
I think a toggle option in the config files would be a good idea, if someone finds a way to add Frostburgh to their server, they could toggle this addition

Jettford:
> Yep, I tested this on both clients however (I will say my lcdr client is modified tho)

lcdrs unpacked client by default does not contain the required resources for this to work

It will display
```
%[ZoneTable_98_DisplayDescription]
%[ZoneTable_98_summary]
```
with an image of DeepFreeze above it and clicking on it will send them to a world they can't get out of

PyEddy:
> > Yep, I tested this on both clients however (I will say my lcdr client is modified tho)
> 
> lcdrs unpacked client by default does not contain the required resources for this to work
> 
> It will display
> 
> ```
> %[ZoneTable_98_DisplayDescription]
> %[ZoneTable_98_summary]
> ```
> 
> with an image of DeepFreeze above it and clicking on it will send them to a world they can't get out of

Indeed, this is why an option to toggle this entry on or off would be a good idea, if someone doesn't have it (they haven't went fixed it in their client), it would be off by default, if they do find a way to access Frostburgh, they can toggle this on

Jettford:
I would leave it to modders to add this if they go through the effort of re adding frostburgh as adding numerous config keys for this small feature is a bad idea because it would be at least 3. e.g. frostburgh_rocket_enabled, frostburgh_rocket_zone_id, frostburgh_rocket_thumbnail_path

PyEddy:
> I would leave it to modders to add this if they go through the effort of re adding frostburgh as adding numerous config keys for this small feature is a bad idea because it would be at least 3. e.g. frostburgh_rocket_enabled, frostburgh_rocket_zone_id, frostburgh_rocket_thumbnail_path

Yeah I understand, I could see you using this on your server lol

Me and lcdr were talking about this on discord, he wanted to add Frostburgh into the LUP interface and I did that, maybe he might extend further upon this

Jettford:
> Me and lcdr were talking about this on discord, he wanted to add Frostburgh into the LUP interface and I did that, maybe he might extend further upon this

I doubt he meant he wanted to push this upstream, if I am wrong I apologise but a significantly better solution is needed than this.

PyEddy:
> > Me and lcdr were talking about this on discord, he wanted to add Frostburgh into the LUP interface and I did that, maybe he might extend further upon this
> 
> I doubt he meant he wanted to push this upstream, if I am wrong I apologise but a significantly better solution is needed than this.

Oh agreed, we do need a better solution, that's kind of why I pitched the idea of toggling it on the config files
I don't know if he wanted this upstream either, it was a discussion between me, Wincent and lcdr where lcdr brought up the idea of adding an entry to the LUP interface and Wincent brought up the idea of adding it to the property interface

I just programmed the addition to the LUP interface, I could DM you the convo we had on discord if you want

averysumner:
I see no reason to add this, here’s a few reasons:
1. It deviates from live LEGO Universe significantly.
2. Current Frostburgh is broken and missing all missions, assuming you do get it working via FDB editing or asset swapping.
3. None of the scripts for Frostburgh are implemented, and the ones that are are broken.

For these reasons I’m closing this PR.

PR #359, opened 2022-01-09
Add handling of SIGTERM

AxtonF:
supervisord, systemd services and the kill command by default send SIGTERM signals to shutdown proccesses. The Master Server should handle the SIGTERM signal the same way as SIGINT to ensure a gracefull shutdown on SIGTERM signal.

I tested this on my ubuntu 20.04 VPS and it works as intended.

### DIFF
--- a/dMasterServer/MasterServer.cpp
+++ b/dMasterServer/MasterServer.cpp
@@ -62,6 +62,7 @@ int main(int argc, char** argv) {
 	//Triggers the shutdown sequence at application exit
 	std::atexit(ShutdownSequence);
 	signal(SIGINT, [](int) { ShutdownSequence(); });
+	signal(SIGTERM, [](int) { ShutdownSequence(); });
 
 	//Create all the objects we need to run our service:
 	Game::logger = SetupLogger();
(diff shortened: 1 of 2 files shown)


Jettford:
There are similar signal things in, ChatServer, WorldServer and AuthServer. Would this fix be needed there too?

AxtonF:
It depends. I think theoretically yes and I should probably add that to the pull request but its not that important since you wouldn't go around killing random world servers. But killing the MasterServer is actually a common occurrence if you use a supervisor or just the normal kill command.

Jettford:
Ah ok, wasn't certain what SIGTERM was.

AxtonF:
It's basically just a signal from the OS to the programm that it should shut down now. And SIGTERM ist the standard signal that tells programms to shut down. But since shutting down all the other servers is the responsibility of the Master Server it's not important for them to react differently. I added the handling to the WorldServer anyways, but I could not find any signal handling so I guess it wasn't important because they don't have to do anything special to shutdown? Feel free to correct me on that.

Xiphoseer (approved):

PR #365, opened 2022-01-10
Change the win delay to increase time to get BONS loot

Jettford:
The time was previously 5 seconds, 30 seconds is what it was in live.


averysumner:
Looks good to me.

PR #366, opened 2022-01-10
Implementing and Fixing All Racing Achievements

### LINKED ISSUE #218: Some racing missions not working
### 

### Issue Description

I've tried many times, they just don't work:

"No Wreckin'"
 
"Dragonscourge"

### Reproduction steps

take these missions when you get to them, race. Counters for the dragon eggs wont go up, and missions like No Wreckin' will not complete.

### Expected Behavior

missions should be completable as normal

### Environment

ubuntu server

EmosewaMC:
Previously, most racing achievements either did not progress correctly or did not progress at all.  This PR fixes all of these tasks and allows them all to progress properly.  Bugs unrelated to the achievements functionality will not be in this PR (Ex. Lap times are not properly tracked for the first and possibly second lap making some lap times only possible on lap 3).  

A new enum file has been added that contains named taskparam1 to reduce the amount of magic numbers that are in the original implementation.  

This PR also fixes general grammatical errors found while working on this PR and also has alphabetized includes in files I was working in.
Fixes #218

### DIFF
--- a/dGame/dMission/RacingTaskParam.h
+++ b/dGame/dMission/RacingTaskParam.h
@@ -1,0 +1,20 @@
+#pragma once
+
+#include <cstdint>
+
+enum class RacingTaskParam : int64_t {
+    RACING_TASK_PARAM_FINISH_WITH_PLACMENT = 1,             //<! A task param for finishing with a specific placement.
+    RACING_TASK_PARAM_LAP_TIME = 2,                         //<! A task param for finishing with a specific lap time.
+    RACING_TASK_PARAM_TOTAL_TRACK_TIME = 3,                 //<! A task param for finishing with a specific track time.
+    RACING_TASK_PARAM_COMPLETE_ANY_RACING_TASK = 4,         //<! A task param for completing a racing task.
+    RACING_TASK_PARAM_COMPLETE_TRACK_TASKS = 5,             //<! A task param for completing a task for a specific track.
+    RACING_TASK_PARAM_MODULAR_BUILDING = 6,                 //<! A task param for modular building with racing builds.
+    RACING_TASK_PARAM_SAFE_DRIVER = 10,                     //<! A task param for completing a race without smashing.
+    RACING_TASK_PARAM_SMASHABLES = 11,                      //<! A task param for smashing entities during a race.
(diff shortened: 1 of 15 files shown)


dGame/dMission/RacingTaskParam.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include <cstdint>
+
+enum class RacingTaskParam : int64_t {
```
Jettford, line 5: This enum shouldn't be its own file, should be in one of the racing components imo

Jettford: or dCommonVars

Wincent01: I believe putting it into its own file is fine, would prefer that rather than clod up dCommonVars even more

averysumner: Whatever you do, do not put it in dCommonVars.

Jettford: Lol, I just thought that was where all enums were slapped

dGame/dMission/RacingTaskParam.h
```diff
@@ -0,0 +1,20 @@
...
+    RACING_TASK_PARAM_FINISH_WITH_PLACMENT = 1,             //<! A task param for finishing with a specific placement.
```
Marcono1234, line 6: Looks like there is a typo (first 'e' of 'placement' is missing): <code>RACING_TASK_PARAM_FINISH_WITH_PLAC<ins>E</ins>MENT</code>

dGame/dMission/RacingTaskParam.h
EmosewaMC: Since MissionTaskType is in its own file I think this fits well in its own file because these values are only used for racing.  Have updated the enum to be a int32_t as I recall that enums shouldnt be of size 64.

dGame/dMission/RacingTaskParam.h
EmosewaMC: Fixed.  Good spot!

EmosewaMC:
I have tested this rigorously and have not had any issues or hard crashes.  Whenever there is some time for reviews @Wincent01 I know you wrote the racing control components so when you have time to review this that'd be great!  Any other reviews are also appreciated.

Jettford (approved):

PR #370, opened 2022-01-12
Fixed a bug where Chaos Cleaner would accidentally add a Maelstrom Dagger to your inventory

EmosewaMC:
This mission does take the item away but gives it back to you for some reason.  This has been fixed in this PR.


Jettford:
This seems correct, I am going to merge

PR #372, opened 2022-01-13
Fixed an issue where Gate Crasher task was progressed when eaten by a shark

EmosewaMC:
Fixed an issue where getting eaten by a shark would accidentally progress the task for entering the Ravencloud gate's left hand side.  

This creates an issue where you are now unable to get the achievement Gate Crasher however this is due to a larger issue that is not easily fixed.

Tested getting killed by a shark on Ubuntu and behavior is as expected; Shark mission progresses and Gate Crasher does not


Jettford:
Both of these mission IDs appear to be invalid missions?

PR #375, opened 2022-01-13
Added Bat Lord Full Set Bonus

EmosewaMC:
Previously the Bat Lord full set bonus was not implemented at all.  This PR implements the functionality for the full set to properly restore 3 hp on a kill.


Jettford (approved):

Jettford:
This looks fine to me but Wincent should definitely look at it considering iirc (Sorry if wrong) this is his system

PR #379, opened 2022-01-14
Add instructions on solving missing DLL issue

Jettford:


averysumner:
Better solution would be to copy the dlls into the build folder.

Jettford:
That is a good point to be honest.

TheMatt2:
The file can be moved programmatically by adding a line in CMakeLists.txt

```
229     endif(UNIX)
230    if(WIN32)
231    link_directories(${mysql_SOURCE_DIR}/lib64/vs14)
+++    file(COPY ${mysql_SOURCE_DIR}/lib64/mysqlcppconn-9-vs14.dll DESTINATION ${PROJECT_BINARY_DIR})
232    endif(WIN32)
```

Jettford:
You need to also move the lib*.dll's too

TheMatt2:
Interesting. It seemed work anyway.

There is probably a way to use `file(GLOB...)` to move all of them at once, otherwise they would all need to be listed.
Or just list them.

Perhaps there should also be a check to move the `.pdb` files if it is a debug build?

Jettford:
PDBs are referenced by absolute paths, they shouldn't be moved

averysumner:
zlib’s dll should also be moved (which notably is named different is the build is debug. Best solution would probably be to recursively add all .dll or .so files to the build folder.

TheMatt2:
> PDBs are referenced by absolute paths, they shouldn't be moved

I'm a bit confused how this could be the case. Specifically for the PDB files for mysql's connector, they are downloaded from mysql website, not compiled.

So how could they be referenced by absolute path?

I assume they are meant to be in the same folder as the corresponding DLL, though I don't know how to really test this.

Jettford:
Hmm, I have never thought about that. I just know inside the PE files the PDB is always defined by its absolute compile path

averysumner:
> Hmm, I have never thought about that. I just know inside the PE files the PDB is always defined by its absolute compile path

You would likely need to manually re-reference the pdb file if you ever wanted to use it. I don't see it being worth it to move the PDBs by default.

Xiphoseer:
There's a cmake `install` directive + command specifically for moving binaries and libraries where they need to be. Doing that as part of build isn't really the best option IMO

Xiphoseer:
Suggestion: Merge this and open an issue about automatic installation?

aronwk-aaron:
Taking Xipho's suggestion and merging this doc change, then making an issue to fix the cmake issue

Issue #181, opened 2021-12-11
Banana bunches break in the trees


MustacheMan2:
### 

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### Issue Description

When you hit a banana tree, the bunch breaks while up top, this can cause bananas to remain out of reach.

https://user-images.githubusercontent.com/48499798/145685988-5cfcb6ca-b364-4959-8a18-7320139f8a47.mp4

### Reproduction steps

walk up to a banana tree, and hit the trunk.

### Expected Behavior

The bunch should fall from the tree still intact, and then will drop the items when broken apart.

https://youtu.be/5OrRMmr4aUU?t=174

### Environment

ubuntu server

NealSpellman:
This is another won't fix on our part, I think. @Wincent01 If you could comment, I don't exactly recall what the problem was initially that lead us to choosing to not make it accurate. Was it some AI/physics component for the banana falling? Something like that.

Wincent01:
The bananas are special in that they use rigidbody physics. To make them bounce around on the floor correctly we'd need to do an entire physics simulation, which for a couple bananas ain't worth it.

DarwinAnim8or:
@Wincent01 I think that we could solve this issue by spawning the bananas slightly lower maybe? (so that they don't end up on top of the tree I mean)

averysumner:
I mean if someone wants to PR in full dynamics simulation I’m not stopping them… :P

DarwinAnim8or:
Reopen this issue if issue persists after applying the latest update (#383 has been merged to main)

PR #383, opened 2022-01-16
Changed banana drop spawn Y height to prevent them from getting stuck on the trees

EmosewaMC:
Changed the Bananas drop Y spawn height to be 10 Y units lower to prevent bananas from getting stuck on the tree.  Tested on various banana trees in my local instance and had zero issues.

### DIFF
--- a/dScripts/GfBanana.cpp
+++ b/dScripts/GfBanana.cpp
@@ -64,8 +64,13 @@ void GfBanana::OnHit(Entity* self, Entity* attacker)
 
 		return;
 	}
-
-	bananaEntity->SetPosition(bananaEntity->GetPosition() - NiPoint3::UNIT_Y * 5);
+	
+	// Get the banana cluster position
+	auto entityPosition = bananaEntity->GetPosition();
+	// Lower by 10 units to prevent bananas from spawning on top of the tree
+	entityPosition.SetY(entityPosition.GetY() - 10);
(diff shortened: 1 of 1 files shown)


cooltrain7:
Pretty sure you only need to update the value on the original line. Also looks like 8 units is the lowest you can have it in order for all the drops to fall.

```	
// Lower by 8 units to prevent drops from spawning on top of the tree
bananaEntity->SetPosition(bananaEntity->GetPosition() - NiPoint3::UNIT_Y * 8);
```

Might also want to attach 181 as the linked issue.

EmosewaMC:
I am not a maintainer and cannot link issues.

Made this change

aronwk-aaron:
#181 ping issue

cooltrain7:
> I am not a maintainer and cannot link issues.

Add `Fixes 181` to your original post and it should attach it. Just something for the future.

DarwinAnim8or:
Erm, 8 means that it takes away less height than before, not more. Meaning now the bananas will spawn higher than they do currently. 

Please try it with 12 instead

cooltrain7:
I suggested 8 as it seemed to be a good position that wasn't too low, yet kept them from getting stuck.
TreeDrops

DarwinAnim8or:
> 

Ah yeah, this seems good. Merging then

Issue #377, opened 2022-01-13
BUILD: Docker setup throws Errors on Windows and Linux


DasRaschloch:
### Make sure you've done the following:

- [X] I have read the [installation guide](https://github.com/DarkflameUniverse/DarkflameServer/blob/main/README.md).

### Platform

Other

### Architecture

ARM

### Error Logs

There seems to be an issue with the installation of the Darkflame Docker as the command "docker compose up -d --build" throws several Warnings/Errors. After finished installation the command "docker compose exec darkflame /app/MasterServer -a" then throws the Error:

Error: no such table: ActivityRewards
Fatal error -1
Stacktrace:
/app/MasterServer[0x414163]
/usr/local/lib64/libstdc++.so.6(+0xb10ea)[0x7f15364ce0ea]
/usr/local/lib64/libstdc++.so.6(+0xb1155)[0x7f15364ce155]
/usr/local/lib64/libstdc++.so.6(+0xb13e8)[0x7f15364ce3e8]
/app/MasterServer[0x40a80c]
/app/MasterServer[0x43fc97]
/app/MasterServer[0x416864]
/app/MasterServer[0x41c81f]
/app/MasterServer[0x4171f6]
/app/MasterServer[0x40bc6d]

After failing multiple times on Linux i tried using Docker Desktop on Win10 and got the same errors.

I followed both the README.md, the docker.md and the docker_Windows.md with no luck.

I listed the Errors/Warnings below in Chronological Order. 
<details>

```
/build/thirdparty/libbcrypt/src/wrapper.c: In function 'crypt_gensalt_ra':
/build/thirdparty/libbcrypt/src/wrapper.c:257:26: warning: implicit declaration of function 'strdup'; did you mean 'strcmp'? [-Wimplicit-function-declaration]
  257 |                 retval = strdup(retval);
      |                          ^~~~~~
      |                          strcmp
/build/thirdparty/libbcrypt/src/wrapper.c:257:24: warning: assignment to 'char *' from 'int' makes pointer from integer without a cast [-Wint-conversion]
  257 |                 retval = strdup(retval);
      |                        ^
-----------------------------------------------------------------------------------------------------------
/build/thirdparty/raknet/Source/RakNetTypes.cpp: In member function 'void SystemAddress::SetBinaryAddress(const char*)':
[...]

Issue #162, opened 2021-12-10
Error: no such table: ActivityRewards when running ./MasterServer -a


IanSJones:
### 

- [X] I have read the [installation guide](https://github.com/DarkflameUniverse/DarkflameServer/blob/main/README.md).

### Platform

WSL2

### Architecture

x86

### Error Logs

<details>

0_initial.sql has been applied but there is no reference to table ActivityRewards in there.

root@GBR00033WLKNO:/home/ijones/DarkflameServer/build# ./MasterServer -a
Error: no such table: ActivityRewards
Fatal error -1
Stacktrace:
./MasterServer(+0x1b513)[0x7f7f8a355513]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0xa5966)[0x7f7f89755966]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0xa59d1)[0x7f7f897559d1]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0xa5c65)[0x7f7f89755c65]
./MasterServer(+0x1275e)[0x7f7f8a34c75e]
./MasterServer(+0x461a7)[0x7f7f8a3801a7]
./MasterServer(+0x1d1d4)[0x7f7f8a3571d4]
./MasterServer(+0x2463f)[0x7f7f8a35e63f]
./MasterServer(+0x1f8a6)[0x7f7f8a3598a6]
./MasterServer(+0x13ba7)[0x7f7f8a34dba7]
root@GBR00033WLKNO:/home/ijones/DarkflameServer/build#

</details>

Jettford:
You failed to follow the setup, read the readme again.

Maybe take a closer look at the Client database section.

IanSJones:
MariaDB [lego]> show tables;
+---------------------+
| Tables_in_lego      |
+---------------------+
| accounts            |
| activity_log        |
| bug_reports         |
| charinfo            |
| charxml             |
| command_log         |
| friends             |
| leaderboard         |
| mail                |
| object_id_tracker   |
| pet_names           |
| play_keys           |
| properties          |
| properties_contents |
| servers             |
| ugc                 |
+---------------------+
16 rows in set (0.003 sec)


OK, I will re-read - thanks.

IanSJones:
"Run each SQL file in the order at which they appear here on the database"

^ There is only one file ^  by the way,  :o)

IanSJones:
[...]

PR #385, opened 2022-01-17
Clearer Error Message on Misconfiguration

TheMatt2:
Previously, if a user didn't run the LCDR script to generate a CDServer.sqlite file `./MasterServer` will crash with this error message.

```
Error: no such table: ActivityRewards
Fatal error -1
Stacktrace:
/app/MasterServer[0x414163]
/usr/local/lib64/libstdc++.so.6(+0xb10ea)[0x7f15364ce0ea]
/usr/local/lib64/libstdc++.so.6(+0xb1155)[0x7f15364ce155]
/usr/local/lib64/libstdc++.so.6(+0xb13e8)[0x7f15364ce3e8]
/app/MasterServer[0x40a80c]
/app/MasterServer[0x43fc97]
/app/MasterServer[0x416864]
/app/MasterServer[0x41c81f]
/app/MasterServer[0x4171f6]
/app/MasterServer[0x40bc6d]
```

This issue is caused by a configuration problem, yet causes a traceback and the cryptic message "Error: no such table ActivityRewards"

This message does not tell the user they need to generate the file CDServer.sqlite and causes confusion.
See #162 and #377

This PR clarifies the error message so a user will now see.

```
$ ./MasterServer -a
[17-01-22 19:32:36] [WorldServer] ./res/CDServer.sqlite does not exist
```

This shorter error message makes it clearer that this is a configuration issue and hopefully will lead the user to discover the instructions in https://github.com/DarkflameUniverse/DarkflameServer#resources to generate the file.

Also, if the file CDServer.sqlite exists, but is empty or corrupted, a more clear message is displayed. This could occur if the current version of code is run, which will create a blank CDServer.sqlite file when the crash occurs.

```
./MasterServer -a
Error: no such table: ActivityRewards
[17-01-22 20:04:39] [WorldServer] Failed to initialize CDServer SQLite Database
[17-01-22 20:04:39] [WorldServer] May be caused by corrupted file: ./res/CDServer.sqlite 
[17-01-22 20:04:39] [WorldServer] Error: SQLITE_ERROR[1]: no such table: ActivityRewards
[17-01-22 20:04:39] [WorldServer] Error Code: 1
```

This suppresses the traceback, since it is a configuration issue, and reminds the user this is likely related to the CDServer.sqlite file.

### DIFF
--- a/dMasterServer/MasterServer.cpp
+++ b/dMasterServer/MasterServer.cpp
@@ -5,6 +5,7 @@
 #include <map>
 #include <string>
 #include <thread>
+#include <filesystem>
 
 #ifdef _WIN32
 #include <bcrypt/BCrypt.hpp>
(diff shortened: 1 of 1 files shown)


Xiphoseer:
FYI, this seems to cause a linker error on GCC 8, where you need to pass `-lstdc++fs` to link the `std::filesystem` apis

PR #386, opened 2022-01-18
Added joined a faction flag

EmosewaMC:
Added logic to set the player flag for joining a faction as was done in live.  Tested on Ubuntu and had zero issues.

### DIFF
--- a/dGame/Character.cpp
+++ b/dGame/Character.cpp
@@ -451,6 +451,13 @@ bool Character::GetPlayerFlag(const uint32_t flagId) const {
 	return false; //by def, return false.
 }
 
+void Character::SetRetroactiveFlags() {
+	// Retroactive check for if player has joined a faction to set their 'joined a faction' flag to true.
+	if (GetPlayerFlag(ePlayerFlags::VENTURE_FACTION) || GetPlayerFlag(ePlayerFlags::ASSEMBLY_FACTION) || GetPlayerFlag(ePlayerFlags::PARADOX_FACTION) || GetPlayerFlag(ePlayerFlags::SENTINEL_FACTION)) {
+		SetPlayerFlag(ePlayerFlags::JOINED_A_FACTION, true);
+	}
+}
+
 void Character::SaveXmlRespawnCheckpoints() 
 {
     //Export our respawn points:
--- a/dGame/Character.h
+++ b/dGame/Character.h
@@ -413,6 +413,12 @@ public:
 	 */
 	void SendMuteNotice() const;
 
+     /**
+      * Sets any flags that are meant to have been set that may not have been set due to them being
+      * missing in a previous patch.
+      */
+     void SetRetroactiveFlags();
+     
     /**
      * Get the equipped items for this character, only used for character creation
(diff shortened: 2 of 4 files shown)


Jettford:
Is this in the packet captures?

Jettford (approved):

averysumner:
I was the one who translated this script from Lua and I feel like I would've added that if it was there... It's possible I missed it. Do you have any proof it was there?

averysumner:
Found the script, I thought it was going to be in 02_server but it wasn't: https://explorer.lu-dev.net/scripts/ai/ns/l_ns_get_faction_mission_server.lua

I don't see this flag getting set, but I also don't see these faction flags getting manually set, so it's likely I figured out the flags on my own. This is probably fine to merge.

Jettford:
Before this is merged, would it be worth adding a simple check during Player creation on world connect to set this flag if they are in a faction, as a like legacy check?

EmosewaMC:
I was thinking about that Jett.  I think that would be a good thing for retroactive flag setting

EmosewaMC:
@Jettford I added a retroactive flag check to the world server for when the level load is completed by user to set their joined a faction flag to true.  @averysumner if you could double check if this is an ok change here or if I should put this elsewhere that'd be great.

Jettford:
You are putting `== true` after each GetFlag, you don't need to do this, it already returns a bool.

EmosewaMC:
@Jettford Noted for future implementations of booleans in statements.  Implemented that change

cooltrain7:
Might be a good idea to move it to a function? I would add it as a character function `Character::SetRetroactiveFlags` or something of the like,  Keeps it in one place, and creates a place if others need to set missed flags in the future.

Jettford:
I don't really see the need, it will never need to be called from anywhere else and isn't a large amount of code

cooltrain7:
> I don't really see the need, it will never need to be called from anywhere else and isn't a large amount of code

Yea its just and idea, just feels a little odd to have faction flag being set from a spot halfway down WorldServer.

Jettford:
WorldServer is literally one massive function ¯\_(ツ)_/¯

DarwinAnim8or:
> Might be a good idea to move it to a function? I would add it as a character function `Character::SetRetroactiveFlags` or something of the like, Keeps it in one place, and creates a place if others need to set missed flags in the future.
> 
> Oh I also don't know what the sentiment would be to change out magic numbers for use of enums? I.e `flagID = 46;` for `flagID = VENTURE_FACTION;`? Though that could always be another PR that changes it across the board (where applicable).

Please do this instead, HandlePacket shouldn't be responsible for player's save data.

EmosewaMC:
@DarwinAnim8or moved the retroactive flag check to its own function in character.cpp.  Tested and had no issues.

EmosewaMC:
> Oh I also don't know what the sentiment would be to change out magic numbers for use of enums? I.e `flagID = 46;` for `flagID = VENTURE_FACTION;`? Though that could always be another PR that changes it across the board (where applicable).
Separate PR imo

PR #391, opened 2022-01-18
Remove ability for gmlevel 0 to use /playanim

averysumner:
This PR removes the ability for players under GAME_MASTER_LEVEL_DEVELOPER to use /playanim. Normally I would just push this change, but there has been controversy in various circles about whether or not this should be removed. To me, the choice is clear: there is no reason normal users should have access to a debug command, and it should be removed. I'd like others' opinions on this.

### DIFF
--- a/dGame/dUtilities/SlashCommandHandler.cpp
+++ b/dGame/dUtilities/SlashCommandHandler.cpp
@@ -375,11 +375,6 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 	}
 
     if (user->GetMaxGMLevel() == 0 || entity->GetGMLevel() >= 0) {
-		if ((chatCommand == "playanimation" || chatCommand == "playanim") && args.size() == 1) {
-			std::u16string anim = GeneralUtils::ASCIIToUTF16(args[0], args[0].size());
-			GameMessages::SendPlayAnimation(entity, anim);
-		}
-
 		if (chatCommand == "die") {
 			entity->Smash(entity->GetObjectID());
 		}
@@ -457,6 +452,11 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 		ChatPackets::SendSystemMessage(sysAddr, GeneralUtils::ASCIIToUTF16(lowerName) + u" set to " + (GeneralUtils::to_u16string(minifigItemId)));
 
 		GameMessages::SendToggleGMInvis(entity->GetObjectID(), false, UNASSIGNED_SYSTEM_ADDRESS); // need to retoggle because it gets reenabled on creation of new character
+	}
+	
+	if ((chatCommand == "playanimation" || chatCommand == "playanim") && args.size() == 1 && entity->GetGMLevel() >= GAME_MASTER_LEVEL_DEVELOPER) {
+		std::u16string anim = GeneralUtils::ASCIIToUTF16(args[0], args[0].size());
+		GameMessages::SendPlayAnimation(entity, anim);
 	}
 
 	if (chatCommand == "list-spawns" && entity->GetGMLevel() >= GAME_MASTER_LEVEL_DEVELOPER) {


Jettford (changes requested):
My personal view on this is that the command is something fun for the game, it has brought enjoyment out of random thing such as waiting for people to arrive. I would like this to stay in the code base as is due to the fun it brings at moments you don't expect it to.

Xiphoseer:
Compromise: Make tools to add more animations as standard emotes while we can't make new ones anyway right now?

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -458,6 +453,11 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 
 		GameMessages::SendToggleGMInvis(entity->GetObjectID(), false, UNASSIGNED_SYSTEM_ADDRESS); // need to retoggle because it gets reenabled on creation of new character
 	}
+	
+	if ((chatCommand == "playanimation" || chatCommand == "playanim") && args.size() == 1 && entity->GetGMLevel() >= GAME_MASTER_LEVEL_DEVELOPER) {
+		std::u16string anim = GeneralUtils::ASCIIToUTF16(args[0], args[0].size());
```
Xiphoseer, line 458: Ugh, another case where we shouldn't use this function on user input.

averysumner: > Another case where we shouldn't use this function on user input.

Yeah... there's a ton of these.

averysumner:
> Compromise: Make tools to add more animations as standard emotes while we can't make new ones anyway right now?

I like this idea, I see no reason to include a debug command in the standard set for civilian accounts. You can do stuff like the rocket sit animation (locks you into a sitting position until you relog lol), teleport animations, animations that shrink you, etc.

MickVermeulen:
I think it's a really fun command, maybe it's indeed a good idea to move /playanimation to developers and introduce a new /emote command with a set of whitelisted animations that are useful / fun during gamplay.

Jettford:
or alternatively, blacklist ones we don't want?

averysumner:
> I think it's a really fun command, maybe it's indeed a good idea to move /playanimation to developers and introduce a new /emote command with a set of whitelisted animations that are useful / fun during gamplay.

This serves no additional purpose though, it’s no different from adding an emote command like “/salute” for example.

Jettford:
Adding emotes isn't an easy task for the regular DLU user, it definitely does serve a purpose.

averysumner:
> Adding emotes isn't an easy task for the regular DLU user, it definitely does serve a purpose. Creating tooling to edit the cdclient for clients isn't a clean solution

Letting normal users use a debug command isn’t a “clean” solution either.

Jettford:
Then we are at a stand still, tooling isn't an option and removing it is not wanted by at least a majority of the community (based upon a poll taken place in a LU discord)

averysumner:
> Then we are at a stand still, tooling isn't an option and removing it is not wanted by at least a majority of the community (based upon a poll taken place in a LU discord)

I’m okay with an addition to vanity code to add certain animations as “emotes” under a possible “/emote” command. I don’t think access to the full set of animations with a debug command should be possible by normal users though.

Jettford:
Ok I will add that as I suggested if no one else has any better ideas?

enteryournamehere:
> Ok I will add that as I suggested if no one else has any better ideas?

I think a separate command with a configurable subset of all animations available would be good, I'd just like to request there be an option to allow all animations minus a blacklist, as almost all animations are just fun and aren't negatively affecting gameplay or otherwise inappropriate.

Jettford:
I still don't want to see this just merged, I think some configuration or white/blacklist is needed.

averysumner:
DLU’s stated goal is an accurate recreation of the live game server (although we do have some extra features that were added before the FOSS release). Seeing as this was not intended behavior, and was not a feature included in live, I have made the decision to merge this. Feel free to make pull requests to add a toggle for this, or add an extra emote command and they can be considered separately.

Jettford:
This has received one change request and no review from 4 of the requested people. Please do not.

PR #392, opened 2022-01-18
Repair some encoding issues with commands and property menus.

Jettford:


dCommon/GeneralUtils.cpp
```diff
@@ -57,38 +57,21 @@ std::u16string GeneralUtils::ASCIIToUTF16(const std::string& string, size_t size
     return ret;
 }
 
-//! Converts a (potentially-ill-formed) UTF-16 string to UTF-8
-//! See: <http://simonsapin.github.io/wtf-8/#decoding-ill-formed-utf-16>
+//! Converts an UCS-2 / UTF-16 to std::string (ASCII)
 std::string GeneralUtils::UTF16ToWTF8(const std::u16string& string, size_t size) {
```
Xiphoseer, line 61:
```suggestion
std::string GeneralUtils::UTF16ToUTF8(const std::u16string& string, size_t size) {
```

Again, if you change the implementation, change the name 😉

dCommon/GeneralUtils.cpp
```diff
@@ -57,38 +57,21 @@ std::u16string GeneralUtils::ASCIIToUTF16(const std::string& string, size_t size
...
-            } else {
-                PushUTF8CodePoint(ret, u);
-            }
-        } else {
-            PushUTF8CodePoint(ret, u);
-        }
-    }
-
-    return ret;
+    std::wstring_convert<std::codecvt_utf8_utf16<char16_t>, char16_t> convert;
```
Xiphoseer, line 62: Oh, sorry that one is actually deprecated in favor of `std::codecvt<char16_t, char>` in c++17, which is what we're using afaik

Xiphoseer: cf. https://en.cppreference.com/w/cpp/locale/codecvt

Xiphoseer:
```
Error: /home/runner/work/DarkflameServer/DarkflameServer/dCommon/GeneralUtils.cpp:71:10: error: ‘wstring_convert’ is not a member of ‘std’
```

dCommon/GeneralUtils.cpp
Jettford: Ah yeah, that is a simple change

Jettford:
After more looking into this by me and Xipho, GCC has missing support for codecvt and it is deprecated in the standard with seemingly no replacement, this appears to be a massive mess. Something needs to be done because at the moment user input can be used to crash servers.

Xiphoseer:
As said in Discord, I recommend adding https://thephd.dev/any-encoding-ever-ztd-text-unicode-cpp

Jettford:
Ah yeah, that was stated after, I will add that later

tests/TestLDFFormat.cpp
```diff
@@ -22,7 +22,11 @@ int TestLDFFormat(int argc, char* *const argv) {
 	ASSERT_EQ(((LDFData<std::u16string>* )data)->GetValue(), u"VALUE");
 
 	// Check that the serialization is correct
-	ASSERT_EQ(data->GetString(), "KEY=0:VALUE");
+	if (data->GetString() != "KEY=0:VALUE"){
```
Xiphoseer, line 25: why did you change this?

dZoneManager/Level.cpp
```diff
@@ -193,13 +193,14 @@ void Level::ReadSceneObjectDataChunk(std::ifstream & file, Header & header) {
...
         
-        std::string sData = GeneralUtils::UTF16ToWTF8(ldfString);
+        std::string sData = GeneralUtils::UTF16ToUTF8(ldfString);
         std::stringstream ssData(sData);
         std::string token;
         char deliminator = '\n';
         
         while (std::getline(ssData, token, deliminator)) {
             LDFBaseData * ldfData = LDFBaseData::DataFromString(token);
+			if (!ldfData) continue;
```
Xiphoseer, line 203: Why this change?

thirdparty/text/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+add_library(ztd.text ${PROJECT_SOURCE_DIR}/thirdparty/text/source/ztd/idk/bit.intrinsic.c
```
Xiphoseer, line 1: Why not use [`CMAKE_CURRENT_SOURCR_DIR`](https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_SOURCE_DIR.html)?

thirdparty/text/README.md
```diff
@@ -0,0 +1,6 @@
+This is an unmodified version of 2 libraries 
+`https://github.com/soasis/idk`
+and
+`https://github.com/soasis/text`
+
+but was removed from their original CMake solution due to issues involving FetchContent
```
Xiphoseer, line 6: Can you describe ehat issues those were?

Xiphoseer (commented):
Not a fan of vendoring the whole library like that if we can avoid it. What did you try and how did it fail?

thirdparty/text/README.md
Jettford: No I can't, it wasn't erroring out but just not getting them. It was extremely weird

dCommon/GeneralUtils.cpp
```diff
@@ -57,38 +60,32 @@ std::u16string GeneralUtils::ASCIIToUTF16(const std::string& string, size_t size
...
-            } else {
-                PushUTF8CodePoint(ret, u);
-            }
-        } else {
-            PushUTF8CodePoint(ret, u);
-        }
-    }
+//! Converts an UCS-2 / UTF-16 to std::string (ASCII)
+std::string GeneralUtils::UTF16ToUTF8(const std::u16string& string, size_t size) {
+    std::u16string_view utf16_input(string.substr(0, MinSize(size, string)));
```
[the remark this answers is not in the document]
Jettford: Surely limiting what I am transcoding is the same as limiting the output of the transcode, unless size is in bytes?

tests/TestLDFFormat.cpp
Jettford: I was just temporarily trying to work out why it wasn't working on Linux, the test was working on Windows but wasn't on Linux

dZoneManager/Level.cpp
Jettford: Oh, that was a temp change I thought I fixed. I still need to work that out. My bad.

Jettford:
This can be fixed by someone else at this point, I don't really want to touch that CMake.

PR #393, opened 2022-01-19
Implemented a script that spawns a Golem Quick Build on Crux Prime dragons when they reach zero armor

EmosewaMC:
For the most part this script is copy and pasted from the Fv Maelstrom Dragon with added comments and removing the chest that would normally spawn when you kill the dragon.  
I have tested this on my local Ubuntu instance and had zero issues.  Golem will disappear properly if the player does not kill in the time frame and dragon will reset back to normal.  Dragon will resume normal behavior if quick build is not built.


Wincent01:
Did they spawn golems in live? Can't find any footage of it.

EmosewaMC:
They did - I recall part 94 or 95, Jamie fights a dragon and the golem spawns at zero armor.  Can get footage tomorrow with the proof

EmosewaMC:
@Wincent01 found some footage of maelstrom dragons spawning the qb
https://youtu.be/_R0-uQ_P_GI @ around 11 minutes

EmosewaMC:
@Wincent01 the above video shows the dragon QB spawning back in live at the final patch.  Just bumping this since its been a while since the last message

PR #396, opened 2022-01-20
Added check to prevent accidental sentinel passive proc

### LINKED ISSUE #394: BUG: Sentinel Panic Shields are not Live Accurate

### Issue Description

The series of passive skills sometimes referred to as "panic shields" which belong to the 4-kit bonuses for Sentinel Knight, Samurai and Space Ranger do not function in the same way they did during live. As of present, they deviate in two main aspects:

1 - these skills trigger when the player takes damage resulting in them having an armor value of 0, regardless of initial armor value.
2 - these skills will trigger again immediately as soon as condition 1 is met, without an apparent cooldown.

As an additional addendum, the Samurai's panic shield also still has the known issue reported here: https://github.com/DarkflameUniverse/DarkflameServer/issues/211

### Reproduction steps

Wear a sufficient number of the appropriate pieces of gear to obtain the passive skill.

Take damage down to 0 armor.

Wait for the passive effect to timeout.

Take damage again while at 0 armor.

### Expected Behavior

These skills should:

1) only trigger when the player takes damage from a nonzero armor value down to zero armor and should not trigger if the player already has zero armor when they take damage.

2) have an internal cooldown that prevents the ability from triggering again immediately after their effect expires, either from depleting the shield or running out the duration of the effect. The exact duration of this cooldown is impossible to say, but there's some evidence that it should be about 11 seconds after the ability's 30 second lifespan expires or about 40+ seconds after the ability triggers.

### Environment

The planet Earth.

EmosewaMC:
Added a boolean to check if the player is at zero armor already and if so, do not trigger the passive ability.  Tested on Ubuntu and had zero issues triggering the passive from full armor and the passive did not trigger when at less than zero armor.
Fixes #394

### DIFF
--- a/dGame/dComponents/DestroyableComponent.cpp
+++ b/dGame/dComponents/DestroyableComponent.cpp
@@ -246,8 +246,7 @@ void DestroyableComponent::SetArmor(int32_t value) {
     m_DirtyHealth = true;
 
 	// If Destroyable Component already has zero armor do not trigger the passive ability again.
-	bool hasArmor = true;
-	if (m_iArmor == 0) hasArmor = false;
+	bool hadArmor = m_iArmor > 0;
 	
     auto* characterComponent = m_Parent->GetComponent<CharacterComponent>();
     if (characterComponent != nullptr) {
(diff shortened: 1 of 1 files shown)


DarwinAnim8or (commented):
I believe it may be better to name this variable "hadArmor", since it refers to the state the armor had before the new value was set.

Other than that, it looks good-- the only (non-blocking) comment I have is that I would have simplified the boolean to just be a single line, IE:

bool hadArmor = m_iArmor > 0;

EmosewaMC:
> I believe it may be better to name this variable "hadArmor", since it refers to the state the armor had before the new value was set.
> 
> Other than that, it looks good-- the only (non-blocking) comment I have is that I would have simplified the boolean to just be a single line, IE:
> 
> bool hadArmor = m_iArmor > 0;

Ah good call.  Sometimes these simplified checks slip past me.  Will make this change.

EmosewaMC:
@DarwinAnim8or Suggestion added.

PR #400, opened 2022-01-21
Remove <filesystem> dependency

TheMatt2:
Unfortunately, Google Cloud Compute Engine does not support the `<filesystem>` library.

Even if the special flag `-lstdc++fs` is passed, the build still fails to link the appropriate functions.

```
cloudshell:~$ g++ -g -lstdc++fs -std=gnu++17 test.cpp
/usr/bin/ld: /tmp/ccYF9KdM.o: in function `std::filesystem::is_regular_file(std::filesystem::__cxx11::path const&)':/usr/include/c++/8/bits/fs_ops.h:213: undefined reference to `std::filesystem::status(std::filesystem::__cxx11::path const&)'/usr/bin/ld: /tmp/ccYF9KdM.o: in function `std::filesystem::__cxx11::path::path<char [9], std::filesystem::__cxx11::path>(char const (&) [9], std::filesystem::__cxx11::path::format)':/usr/include/c++/8/bits/fs_path.h:184: undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'collect2: error: ld returned 1 exit status
```

Google Cloud uses GCC 8 (`gcc (Debian 8.3.0-6) 8.3.0`), but proper filesystem support was not added until GCC 9.

As a workaround, this PR removes the use of the filesystem library from DLU.

### DIFF
--- a/dMasterServer/MasterServer.cpp
+++ b/dMasterServer/MasterServer.cpp
@@ -5,7 +5,7 @@
 #include <map>
 #include <string>
 #include <thread>
-#include <filesystem>
+#include <fstream>
 
 #ifdef _WIN32
 #include <bcrypt/BCrypt.hpp>
@@ -80,10 +80,12 @@ int main(int argc, char** argv) {
 
 	//Check CDClient exists
 	const std::string cdclient_path = "./res/CDServer.sqlite";
-	if (!std::filesystem::is_regular_file(cdclient_path)) {
-		Game::logger->Log("WorldServer", "%s does not exist\n", cdclient_path.c_str());
+	std::ifstream cdclient_fd(cdclient_path);
+	if (!cdclient_fd.good()) {
+		Game::logger->Log("WorldServer", "%s could not be opened\n", cdclient_path.c_str());
 		return -1;
 	}
+	cdclient_fd.close();
 
 	//Connect to CDClient
 	try {


averysumner:
DarkflameServer is made to compile under C++17, versions before this are unsupported. I'd also like to note that Debain 8 lost support over a year ago (https://wiki.debian.org/LTS) and should no longer be used. You can also just use update your compiler.

Xiphoseer:
@averysumner They're talking about GCC 8 which is used in Debian 10, which is the default image for google cloud.

This is a de-facto change in the minimal because we don't actually have any checks in CMake for minimal compiler. support afaik.

One other component: Some poeple might be confusing "cloudshell" with "compute" because the former seems to be a locked down debian 10 that can run the `gcloud` CLI and could previously compile DLU, while the latter can be configured to use sth like Ubuntu 20.04 which would be what we use in CI

TheMatt2:
The GCC 8 compiler experimentally supports C++17, but needs the flag `-lstdc++fs` to work with the filesystem library.

See https://gcc.gnu.org/gcc-9/changes.html

> Improved support for C++17, including:
> The C++17 implementation is no longer experimental.
> Parallel algorithms and <execution> (requires Thread Building Blocks 2018 or newer).
> <memory_resource>.
> Using the types and functions in <filesystem> does not require linking with -lstdc++fs now.

Despite this flag is supposed to allow C++17 with filesystem, it doesn't seem to work on Google Cloud's Debian 10 image. (It may be possible to add a workaround that detects a GCC 8 compiler and adds special flags to make it work, https://gist.github.com/berkus/aafb15b06f753f025e009d5dd2ae7e5b)

Since the only use of the filesystem library in DLU seems to be to check the existence of one file on startup, it would seem to be trivial to support this C++17 minus filesystem, at least while Google Cloud is using a default image with GCC 8.


So the code is C++17 regardless. The question is do we want to support the GCC 8 compiler.

averysumner:
Why are we supporting a compiler with experimental C++17 support?

PR #402, opened 2022-01-21
Added Comprehensive Slash Command Documentation

DrOverbuild:
This PR takes the original command documentation from the README and adds comprehensive details for each command available in game. 

This creates a new file `/docs/Commands.md`. All available commands are split up by their use — general commands, moderation, server operation, and development. They are then sorted by their required permission level and then by alphabetical order.

This detailed list of commands should help both server operators run their servers and contributors get started with development.

This PR also pioneers a common practice among many open source projects, where markdown documentation is placed in a top level `docs` folder. In this folder we can continue to contribute markdown documentation in order to help with server setup and operation, as well as development and contributions. Documentation here would be more specific to Darkflame Universe and its development than documentation found at https://lu-dev.net.


Jettford:
Would it be possible, to move this to a vanity xml file or something along those lines so you could include this with a bit of server code. e.g.
```
/help <command>
/help <type (moderation, general, developer etc)>

**Comment by DrOverbuild** [comment]
I certainly think a `/help` command would be helpful but I think a markdown page is also just as helpful. It just really depends on the context. For example, if I can’t remember the commands I’m looking for it would be much easier to browse through a markdown page than scroll through a bunch of messages in chat. But if I’m in game and need to remember the usage of a command, `/help` would be far easier.

If we held this data in xml, json, or even in the database, we would probably want to come up with a way to automatically build a markdown file based on the data to keep the `/help` command synchronized with the markdown page.

**Comment by cooltrain7** [comment]
This looks great, imo both are a good idea. I personally would like commands with no requirement to show as 0 and not just blank. As for gmlevel, you might want to note (somehow) that it depends on having a max gmlevel higher than 0.

**Comment by DrOverbuild** [comment]
> I think setting up commands as an class might be a good idea, then during startup you load your xml/json for the commands with the name, argument, and help paramaters. Though thats another PR.

Yes, while reading through the code trying to understand what each command does, I badly wanted to refactor the entire command handler. I’ve built command handlers that make use of polymorphism: each command has a class that overrides a method. It makes handling commands much easier, but yeah, that’s another PR.

**Comment by averysumner** [comment]
> > I think setting up commands as an class might be a good idea, then during startup you load your xml/json for the commands with the name, argument, and help paramaters. Though thats another PR.
> 
> Yes, while reading through the code trying to understand what each command does, I badly wanted to refactor the entire command handler. I’ve built command handlers that make use of polymorphism: each command has a class that overrides a method. It makes handling commands much easier, but yeah, that’s another PR.

Class per command seems overkill. You could clean up our current system a lot if you used a registration function that took the command name, args, and a lambda for the code.

**Comment by sersorrel** [comment]
feedback based on my own notes, which I never got around to making a pr for:

- `mute` requires GAME_MASTER_LEVEL_JUNIOR_DEVELOPER, which is 6, not 2
- `mailitem` requires GAME_MASTER_LEVEL_MODERATOR, which is 3, not 8
- `shutdownuniverse` requires GAME_MASTER_LEVEL_OPERATOR, which is 9, not 8
- `teleport` requires GAME_MASTER_LEVEL_JUNIOR_MODERATOR, which is 2, not 8
- `inspect`'s first argument is required, not optional, and isn't an entity ID
- speaking of `inspect`: I guess it doesn't really need to be documented in detail right here? but it would be nice to have documentation for it _somewhere_... <details><summary>my notes on <code>inspect</code> (mostly untested, probably not terribly accurate)</summary>
`inspect <component> (-m <waypoint> | -a <animation> | -s | -p | -f (faction) | -t)`
    - finds the closest entity with the given component or LDF variable (ignoring players and racing cars)
    - prints its ID(?), distance from the player, and whether it is sleeping, as well as the the IDs of all components the entity has
    - with `-m`, if the entity has a moving platform component, sends it to the given waypoint, or stops the platform if `waypoint` is -1
    - with `-a`, plays the given animation on the entity(?)
    - with `-s`, prints the entity's settings and spawner ID
    - with `-p`, prints the entity's position
    - with `-f`, if the entity has a destroyable component, prints whether the entity is smashable and its friendly and enemy faction IDs; if `faction` is specified, adds that faction to the entity
    - with `-t`, if the entity has a phantom physics component, prints the effect type, direction, directional multiplier, and whether the effect is active; in any case, if the entity has a trigger, prints the trigger ID
</details>

currently undocumented:

- `/gminvis`: toggles invisibility for the character, though it's currently a bit buggy; requires nonzero gmlevel
- `/gmimmume <value>`: sets the character's GMImmune state (pass `1` to make yourself immune to damage, `0` to undo); requires gmlevel 8
- alternate names for commands (`gmlevel` is also `setgmlevel` and `makegm`, `playanimation` is also `playanim`, `setinventorysize` is also `setinvsize`, `teleport` is also `tele`)
- `testmap` accepts an optional "force" argument between the zone and clone ID, to forcibly transfer to a zone that isn't marked as accessible (this can softlock your character, though, if you e.g. try to teleport to Frostburgh)

miscellaneous other suggestions:

- for `skip-ags` and `skip-sg`, it might be useful to include the names of the missions, which (I _think_) are "Impress the Sentinel Faction" and "Monarch of the Sea" respectively
- under `lookup`, "discription" should be "description"
- under `leave-zone`, "Nimbus station" should be "Nimbus Station"

**Comment by averysumner** [comment]
> feedback based on my own notes, which I never got around to making a pr for:
> 
>     * `mute` requires GAME_MASTER_LEVEL_JUNIOR_DEVELOPER, which is 6, not 2
> 
>     * `mailitem` requires GAME_MASTER_LEVEL_MODERATOR, which is 3, not 8
> 
>     * `shutdownuniverse` requires GAME_MASTER_LEVEL_OPERATOR, which is 9, not 8
> 
>     * `teleport` requires GAME_MASTER_LEVEL_JUNIOR_MODERATOR, which is 2, not 8
> 
>     * `inspect`'s first argument is required, not optional, and isn't an entity ID
> 
>     * speaking of `inspect`: I guess it doesn't really need to be documented in detail right here? but it would be nice to have documentation for it _somewhere_... my notes on `inspect` (mostly untested, probably not terribly accurate)
>       `inspect <component> (-m <waypoint> | -a <animation> | -s | -p | -f (faction) | -t)`
>       
>       * finds the closest entity with the given component or LDF variable (ignoring players and racing cars)
>       * prints its ID(?), distance from the player, and whether it is sleeping, as well as the the IDs of all components the entity has
>       * with `-m`, if the entity has a moving platform component, sends it to the given waypoint, or stops the platform if `waypoint` is -1
>       * with `-a`, plays the given animation on the entity(?)
>       * with `-s`, prints the entity's settings and spawner ID
>       * with `-p`, prints the entity's position
>       * with `-f`, if the entity has a destroyable component, prints whether the entity is smashable and its friendly and enemy faction IDs; if `faction` is specified, adds that faction to the entity
>       * with `-t`, if the entity has a phantom physics component, prints the effect type, direction, directional multiplier, and whether the effect is active; in any case, if the entity has a trigger, prints the trigger ID
> 
> 
> currently undocumented:
> 
>     * `/gminvis`: toggles invisibility for the character, though it's currently a bit buggy; requires nonzero gmlevel
> 
>     * `/gmimmume <value>`: sets the character's GMImmune state (pass `1` to make yourself immune to damage, `0` to undo); requires gmlevel 8
> 
>     * alternate names for commands (`gmlevel` is also `setgmlevel` and `makegm`, `playanimation` is also `playanim`, `setinventorysize` is also `setinvsize`, `teleport` is also `tele`)
> 
>     * `testmap` accepts an optional "force" argument between the zone and clone ID, to forcibly transfer to a zone that isn't marked as accessible (this can softlock your character, though, if you e.g. try to teleport to Frostburgh)
> 
> 
> miscellaneous other suggestions:
> 
>     * for `skip-ags` and `skip-sg`, it might be useful to include the names of the missions, which (I _think_) are "Impress the Sentinel Faction" and "Monarch of the Sea" respectively
> 
>     * under `lookup`, "discription" should be "description"
> 
>     * under `leave-zone`, "Nimbus station" should be "Nimbus Station"

All of this is just more reason to entirely refactor the slash command system. We could refactor it to have a dynamic help command based on register commands, args, etc.

**Comment by Xiphoseer** [comment]
This PR essentially adds a HTML file, not a markdown file. Mardown does have a way to generate tables, it might be nice to use that.

My 2ct suggestion:

- Add one file per command. This makes ot much easier for git to track changes
- Use https://jekyllrb.com/docs/front-matter/ for metadata like required GM level, summary, tags, related commands
- Add a nice html layout page that uses the markdown and metadata
- Publish that as a DLU github pages

IMO having the docs in the implementation might be nice, but I'm not sure it makes sense to hide good docs in there, especially as fields in classes. That works for summaries but nor so much for full explainers that don't fit in chat.

A good middleground may also be scraping them from comments in the source, but that's more involved.

**Comment by MichaelFedora** [comment]
Github flavored markdown does have table formatting, re:

```markdown
| Command | Usage    | Description                                                 | Admin Level Requirement |
| ------- | -------- | ----------------------------------------------------------- | ----------------------- |
| credits | /credits | Displays the names of the people behind Darkflame Universe. | 0                       |
```

as well as

```markdown
Command|Usage|Description|Admin Level Requirement
-|-|-|-
credits|/credits|Displays the names of the people behind Darkflame Universe.|0
```

produce

Command | Usage | Description | Admin Level Requirement
-|-|-|-
credits|/credits|Displays the names of the people behind Darkflame Universe.|0

whether in comments or in `.md` files. I think using that is better than doing `<table>` whatnot. (depending on the editor, you can make it auto-format so you don't have to re-do spacing every time as well, btw).

--

re: xipho:

- one file per command would be hell for upkeep / maintenance IMO
- nice dedicate docs page might be good but since we're on GitHub, might as well stick a flavored markdown file in the repo, no?

alternatively, re-doing the command handler and just doing in-game docs ~~might~~ would be more useful (but more time consuming)

Xiphoseer:
> one file per command would be hell for upkeep / maintenance IMO

Can you elaborate on that? I would think that it reduces maintenance burden because you cannot ever have "reorder to make alphabetic" PRs and you need to keep it up to date whether it's a single file or multiple. 

> nice dedicate docs page might be good but since we're on GitHub, might as well stick a flavored markdown file in the repo, no?

Yeah, every one of those steps I added was meant to be an optional addition to the previous step. I.e. adding the docs to the repo would be nice and then we could turn them into pages if they're sufficiently helpful.

Jekyll supports GitHub flavored Markdown by default [^1], so it's really just added flexibility in styling.

[^1]: https://jekyllrb.com/docs/configuration/markdown/

MichaelFedora:
> Can you elaborate on that? I would think that it reduces maintenance burden because you cannot ever have "reorder to make alphabetic" PRs and you need to keep it up to date whether it's a single file or multiple.

You don't have to re order commands unless you are re-naming them :thinking: You just add entries in between others. If people are adding commands at the bottom/top/in the wrong spot, those can be called out in the PR.

There is a balance between "having to modify X files" and "having to modify a file X times"; I would call the former hell over the latter, but I will agree that it's a preference.

I will further say that as an end-user I'd prefer all the commands to be in one place for a reference rather then sprawled between 50 different files. So I still think there is reason to support the mono-doc.

codeshaunted:
> > Can you elaborate on that? I would think that it reduces maintenance burden because you cannot ever have "reorder to make alphabetic" PRs and you need to keep it up to date whether it's a single file or multiple.
> 
> You don't have to re order commands unless you are re-naming them 🤔 You just add entries in between others. If people are adding commands at the bottom/top/in the wrong spot, those can be called out in the PR.
> 
> There is a balance between "having to modify X files" and "having to modify a file X times"; I would call the former hell over the latter, but I will agree that it's a preference.
> 
> I will further say that as an end-user I'd prefer all the commands to be in one place for a reference rather then sprawled between 50 different files. So I still think there is reason to support the mono-doc.
> 
> If the command list gets very large, it can be segmented into category based docs (a good compromise) but because there are so little right now that still seems like to be more tedious then beneficial.

All of these problems are solved by a commands system rework with a dynamic help command. You could even make it generate a documentation file if you feel so inclined.

Jettford:
While yes it would that is a lot of effort and something that this person may not be able/willing to do so in the meantime before someone else gets around to it we should just merge this.

EmosewaMC:
I think this is fine for now.  Some semantic differences and this should def be rebased on main but unless someone else is going to refractor the slash command system, this gives users the ability to at least know what the commands are without having to combine through a cpp file.

EmosewaMC (approved):

codeshaunted:
> While yes it would that is a lot of effort and something that this person may not be able/willing to do so in the meantime before someone else gets around to it we should just merge this.

I disagree, I think this will just get left to rot as new commands get added or old ones change. Better to wait for a full-fledged solution than ship a half baked one.

Jettford:
> I would suggest requiring any contributors to update this document before approving their PR if it includes an update to a command or a new command. Although, I suppose that would be too much overhead.

I think that commands are a rare addition and would be something manageable.

Jettford (approved):

Jettford:
and we could even go as far as using something along the lines of [this](https://github.com/marketplace/actions/verify-changed-files) to verify if the command handler file is changed both are.

aronwk-aaron (changes requested):
I'm fine with this doc, as long as it is actually markdown.
please use something like this to just convert the table from html to md. https://jmalarcon.github.io/markdowntables/
Tested it and this works pretty well

Jettford:
@DrOverbuild Here is a [markdown version of your documentation](https://www.toptal.com/developers/hastebin/vidicenise.rb) you can use to fulfil the above feedback. (Thanks for the tool Aaron)

DrOverbuild:
Thank you for the feedback, I will update with markdown tables, along with addressing sersorrel's feedback when I can get a chance to look at this.

aronwk-aaron:
> DrOverbuild Here is a [markdown version of your documentation](https://www.toptal.com/developers/hastebin/vidicenise.rb) (no idea why it thinks it is Ruby) you can use to fulfil the above feedback. (Thanks for the tool Aaron)

Laziness is one of [the three virtues](https://thethreevirtues.com/)

codeshaunted:
> > I disagree, I think this will just get left to rot as new commands get added or old ones change. Better to wait for a full-fledged solution than ship a half baked one.
> 
> I would suggest requiring any contributors to update this document before approving their PR if it includes an update to a command or a new command. Although, I suppose that would be too much overhead.

> and we could even go as far as using something along the lines of [this](https://github.com/marketplace/actions/verify-changed-files) to verify if the command handler file is changed both are.

I don’t see why this energy can’t just be put into making a real commands system. The current one is a holdover from the early days and has a lot of issues. In my mind it’s a much better use of our time to just update the system rather than fiddle with markdown files forevermore.

Jettford:
That is outside of the scope of this pull request, this is a good in-between while we await someone to refactor that system.

DrOverbuild:
Alright I have rebased from the latest main branch, converted all of the tables to markdown, and updated with @sersorrel's notes. (Thank you sersorrel very much for your feedback!)

aronwk-aaron (approved):
LGTM

EmosewaMC (approved):

aronwk-aaron:
Merging, maintainers are aware if something is added or changed with slash commands, this will need to be updated.
I will also make an issue about refactoring the slash command system so that that can be tracked separately. Docs are needed and waiting months for someone to refactor the slash command system is not a good reason to not get this in here now.

codeshaunted:
Was this really merged with no reference to it in the README?

Jettford:
The documentation for the slash commands is for server maintainers not the average user, I don't see to much of a reason that it needs to be but if you think it should be there, then just add it.

PR #409, opened 2022-01-24
Frakjaw Battle instance fixes

### LINKED ISSUE #362: BUG: Frakjaw Battle Instance does not properly initialize/scale for Multiplayer

### Issue Description

The Battle Against Frakjaw instance has issues when used by a team of players.

Once the team has loaded in, the camera is locked and the game takes an extra minute or so to unlock and spawn the first wave.

On occasion, the instance will fail to initialize entirely and the team will be left outside the statue and will not load into the instance at all within a reasonable amount of time.

Beyond this point, the instance runs the single-player version of the instance and does not properly scale with team size.

Loot does, however, scale properly.

As an aside, the instance still consumes 2 Green Imaginite per use.

### Reproduction steps

Queue for the instance with 2 or more players.

### Expected Behavior

The instance should load consistently, should consume only 1 Green Imaginite and, once loaded, should spawn the first wave almost immediately. Additionally, the contents of each wave as well as Frakjaw's HP should scale based on playercount.

### Environment

Undefenestrated Windows 10.

EmosewaMC:
Fixes #362 
- Address the issue where the minigame would consume 2 Green Imaginate per play.  Now only 1 is correctly consumed.
- Addresses an issue where the minigame was not initializing properly for instances of more than 1 player.  What was happening was the variable `totalPlayersLoaded` was not adding players correctly to its list and for some reason either overwriting the previous list of players each time or was giving each player their own `totalPlayersLoaded` list.  This caused all players lists to have to wait for the ActivityTimerStart and also defaulted the instance to always being single player, even if there were 2 or more players.

Now the instance variable `totalPlayersLoaded` correctly has the number of players that would be loaded in the instance and starts the activity when the size of `totalPlayersLoaded` matches `team->members.size()` exactly. 
Tested on local Ubuntu instance with 1 player, 2 players and 3 players.  Instance correctly loaded and initialized for all counts of players.

### DIFF
--- a/dScripts/NjMonastryBossInstance.cpp
+++ b/dScripts/NjMonastryBossInstance.cpp
@@ -47,9 +47,8 @@ void NjMonastryBossInstance::OnStartup(Entity *self) {
 void NjMonastryBossInstance::OnPlayerLoaded(Entity *self, Entity *player) {
     ActivityTimerStop(self, WaitingForPlayersTimer);
 
-    // Join the player in the activity and charge for joining
+    // Join the player in the activity
     UpdatePlayer(self, player->GetObjectID());
-    TakeActivityCost(self, player->GetObjectID());
 
     // Buff the player
     auto* destroyableComponent = player->GetComponent<DestroyableComponent>();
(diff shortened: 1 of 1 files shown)


Marcono1234:
Would it be necessary to have `OnPlayerExit` remove the player from `totalPlayersLoaded` (and update `TotalPlayersLoadedVariable`)?

EmosewaMC:
> Would it be necessary to have `OnPlayerExit` remove the player from `totalPlayersLoaded` (and update `TotalPlayersLoadedVariable`)?

Can probably do that.

PR #415, opened 2022-01-25
Fix UserManager not logging chosen name properly

Marcono1234:
Previously when the user chose a predefined name, the logging call would simply log an empty string because `name` is empty and their name was only stored in `predefinedName`.

### DIFF
--- a/dGame/UserManager.cpp
+++ b/dGame/UserManager.cpp
@@ -270,7 +270,7 @@ void UserManager::CreateCharacter(const SystemAddress& sysAddr, Packet* packet) 
         return;
     }
     
-	Game::logger->Log("UserManager", "AccountID: %i is creating a character with name: %s\n", u->GetAccountID(), name.c_str());
+	Game::logger->Log("UserManager", "AccountID: %i is creating a character with name: %s\n", u->GetAccountID(), name != "" ? name.c_str() : predefinedName.c_str());
     
     //Now that the name is ok, we can get an objectID from Master:
     ObjectIDManager::Instance()->RequestPersistentID([=](uint32_t objectID) {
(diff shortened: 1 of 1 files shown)


DarwinAnim8or (changes requested):
I think this would be a good oppertunity to change that log message to just log both (if a to be moderated name is entered)

Marcono1234:
A few lines above it is already logging "Got predefined name: %s\n". Should I remove that and combine it in the modified logging call here? E.g.:
```cpp
if (name == "") {
  Game::logger->Log("UserManager", "AccountID: %i is creating a character with predefined name: %s\n", u->GetAccountID(), predefinedName.c_str());
} else {
  Game::logger->Log("UserManager", "AccountID: %i is creating a character with name: %s (temporary: %s)\n", u->GetAccountID(), name.c_str(), predefinedName.c_str());
}
```

averysumner:
> A few lines above it is already logging "Got predefined name: %s\n". Should I remove that and combine it in the modified logging call here? E.g.:
> 
> ```c++
> if (name == "") {
>   Game::logger->Log("UserManager", "AccountID: %i is creating a character with predefined name: %s\n", u->GetAccountID(), predefinedName.c_str());
> } else {
>   Game::logger->Log("UserManager", "AccountID: %i is creating a character with name: %s (temporary: %s)\n", u->GetAccountID(), name.c_str(), predefinedName.c_str());
> }
> ```

Go for it.

EmosewaMC (approved):
This looks good to me.  @averysumner @DarwinAnim8or If this looks good to you two let's get this merged

DarwinAnim8or (approved):
Ah, thanks for bumping this, had forgotten about it. Looks good

PR #419, opened 2022-01-25
Improve GitHub issue templates

Marcono1234:
Tries to improve the GitHub issue templates by:
- Requesting the DarkflameServer version
This makes it later easier to find out which version was affected, and whether the bug might have already been fixed.
- Requesting environment information for performance issues reports
Information about the environment might be quite useful when analyzing performance issues.

Feel free to close this pull request in case you do not think that these changes are useful.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -18,6 +18,14 @@ body:
         - label: >
             I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.
           required: true
+  - type: input
+    id: server-version
+    attributes:
+      label: DarkflameServer Version
+      description: >
+        DarkflameServer version or commit SHA
+    validations:
+      required: true
   - type: textarea
     id: problem
     attributes:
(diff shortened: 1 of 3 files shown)


Xiphoseer:
Commit Hash is nice, but IMO it needs to give a bit more guidance to tell people how to find out which commit they're using. Maybe sth like

Output of `git log --oneline -1`

Marcono1234:
> Output of `git log --oneline -1`

That also seems to include branch names and the commit message. Would `git rev-parse --short HEAD` (as described [here](https://stackoverflow.com/a/5694416)) be more suitable?

Xiphoseer:
> > Output of `git log --oneline -1`
> 
> That also seems to include branch names and the commit message. Would `git rev-parse --short HEAD` (as described [here](https://stackoverflow.com/a/5694416)) be more suitable?

Both work really, it doesn't matter much as long as the output is unambiguous.

Marcono1234:
Have included it in the description. It looks now like this:
Version field screenshot

Is that what you had in mind?

EmosewaMC:
@Xiphoseer or @MickVermeulen could one of you two give this a look?  I believe it's ready to merge but 1 of you two should give it a once over before then

MickVermeulen (approved):
LGTM!

PR #421, opened 2022-01-27
fixed symlinks for client files

TheRealHypR:
Hey,
it looks like the symlink for the client files created inside the docker container point to the wong folder.

The path **/client/client** is not created inside the container. 
The path is only **/client**

After changing, I didn't get the _file not found_ error anymore.

### DIFF
--- a/docker/start_server.sh
+++ b/docker/start_server.sh
@@ -2,18 +2,18 @@
 
 function symlink_client_files() {
     echo "Creating symlinks for client files"
-    ln -s /client/client/res/macros/ /app/res/macros
-    ln -s /client/client/res/BrickModels/ /app/res/BrickModels
-    ln -s /client/client/res/chatplus_en_us.txt /app/res/chatplus_en_us.txt
-    ln -s /client/client/res/names/ /app/res/names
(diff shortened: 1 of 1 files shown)


Jettford:
I am pretty sure that the `/client/client` directory is correct as the container unpacks the clients

Jettford:
Don't quote me on that though @TheNoim will know more

TheNoim:
> 

> I am pretty sure that the `/client/client` directory is correct as the container unpacks the clients

Yes, it is. @TheRealHypR probably has a different client layout. I still had no time to correct the documentation and/or to detect the client layout automatically. Currently, the docker setup only supports packed clients with the full layout including the launcher.

TheNoim:
I started to work on a PR which fixes docker for different client layouts #432

Xiphoseer:
Ok, closing this as "working as intended" because merging it would break the packed client case.

The naming isn't ideal but in terms of live, a LU installation has three folders, `client`, `versions` and `patcher` and the root of that is what the script expects to be pointed at right now, and @TheNoim seems to be interested in making it work for both cases.

PR #431, opened 2022-02-03
Fixed an issue where leave-zone was working outside of instanced zones

EmosewaMC:
Addressed an issue where the leave-zone command was working outside of instanced zones and was also putting players in the incorrect zone when it was used.  
Tested on personal server with no issues.

### DIFF
--- a/dGame/dUtilities/SlashCommandHandler.cpp
+++ b/dGame/dUtilities/SlashCommandHandler.cpp
@@ -305,48 +305,15 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 		const auto currentZone = dZoneManager::Instance()->GetZone()->GetZoneID().GetMapID();
 
 		auto newZone = 0;
-
-		switch (currentZone)
-		{
-		// Avant Gardens instanced zones
-		case 1001:
-		case 1101:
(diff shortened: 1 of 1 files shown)


Wincent01:
I'd prefer if it was rather a check for if the player is in a normal world, i.e zoneID%100==0. This excludes normal zones but leaves the command open to getting folks out of limbo in a non-functional secondary or non-existent world.

EmosewaMC:
Good idea Wincent.  I'll add that change and test it.

averysumner:
> I'd prefer if it was rather a check for if the player is in a normal world, i.e zoneID%100==0. This excludes normal zones but leaves the command open to getting folks out of limbo in a non-functional secondary or non-existent world.

I mean not that Portabello is a "normal world"... but just %100 won't catch everything. It's probably best to disable this unless the player is in a private instance. I believe *most* of the issues surrounding players getting stuck in normal gameplay zones have been resolved (minus that property one).

EmosewaMC:
If the above commit is not an ideal way of solving this problem ill reconsider other solutions.  The above checks should suffice for dlu since its based on the last live patch but if there are too many potential cases of instanced zones that this doesnt cover then I can always refactor it to use a case switch or something

Issue #417, opened 2022-01-25
BUG: Docker Compose server fails to connect to database in some cases


Marcono1234:
### Make sure you've done the following:

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### Issue Description

When using the Docker Compose setup, it looks like in some cases the initial connection from `MasterServer` to the database can fail, and the following is logged:
> [MasterServer] Got an error while connecting to the database: Unable to connect to localhost

It appears this happens when `docker/start_server.sh` starts the server before the database has finished starting up. 
This issue **does not** occur when either:
- `start_server.sh` takes a long time, e.g. because client resources have not been extracted yet, and in the meantime the database has enough time to start
- Database initialization has already been performed previously, then startup is comparatively fast and finishes before the server is started

### Reproduction steps

Might depend on your environment, which affects how fast the initial database setup is.

1. Fully build and start the project with `docker compose up --build`
2. Once everything is started and running, stop it again by pressing <kbd>Ctrl</kbd> + <kbd>C</kbd> in the same console window
3. Run `docker compose down`
4. In Docker Desktop (or from the command line) delete the `darkflameserver_database` volume
5. Start everything again with `docker compose up`
:x: An error is logged for the 'DarkflameServer' container:
    > [MasterServer] Got an error while connecting to the database: Unable to connect to localhost

### Expected Behavior

Server start should wait until the database is running

### Environment

WSL 2, Docker

Issue #399, opened 2022-01-20
BUG: ChatServer doesnt work in Docker | Teams / adding Friends not working(No Notification/Way to accept)


DasRaschloch:
### Make sure you've done the following:

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [x] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### Issue Description

Trying to invite someone to a party or sending a friend request doesn't work at all. There is no Popup or any other way to accept any invitation as there is no notification. As far as this got mentioned before in other Threads this has something to do with the ChatServer not running/crashing. But when i access the Logs of the DarkflameDocker or with docker-compose logs -f --tail 100
there isn't even an entry about the ChatServer getting initialized in the first place, or an entry of a Crash/restart.

### Reproduction steps

Installing Darkflame with Docker as described in Docker.md and Docker_Windows.md

### Expected Behavior

Being able to join Groups/add Friends

### Environment

Linux based System. Darkflame is installed with Docker.

Jettford:
@TheNoim Was this something that you encountered?

TheNoim:
Hm, interesting

TheNoim:
Hey, can you try running `docker compose exec darkflame ps -ax` and send me the output? Want to know whether the ChatServer really crashed or if this is some kind of connection issue.

DasRaschloch:
docker-compose exec darkflame ps -ax:
 ```
 PID TTY      STAT   TIME COMMAND
    1 ?        Ss     0:00 /bin/bash /start_server.sh
   33 ?        Sl    30:23 ./MasterServer
   36 ?        Sl    29:00 ./ChatServer
   38 ?        Sl    32:30 ./WorldServer -zone 0 -port 3000 -instance 0 -maxclients 999 -clone 0
   42 ?        S      0:00 sudo ./AuthServer
   45 ?        Sl    28:51 ./AuthServer
  106 pts/0    Ss+    0:00 sh
 1194 pts/1    Rs+    0:00 ps -ax
```

TheNoim:
> There you go @TheNoim
> 
> docker-compose exec darkflame ps -ax:
> 
>  ```
> 
>  PID TTY      STAT   TIME COMMAND
> 
>     1 ?        Ss     0:00 /bin/bash /start_server.sh
> 
[...]

Issue #145, opened 2021-12-09
World Server will not reconnect to Chat Server


jelimoore:
**Expected Behavior**
When connection to Chat server is lost, the World server attempts to reconnect periodically

**Current Behavior**
When Chat Server crashes or otherwise stops running, all current World Server instances will lose Chat Server connectivity. Ironically, Chat itself still works but Friends and Teams quit working. However, any new World Server instances started after the Chat Server have been restarted have functional Friends and Teams features.

**Possible Solution**
When connection to the Chat Server is lost, periodically poll for Chat Server to reconnect.

**Steps to Reproduce**
1. Start MasterServer
2. Join world
3. Kill ChatServer
4. Restart ChatServer
5. Attempt to add a friend

**Your Enviorment**
Ubuntu 20.04 LTS, x86 CPU VM, VMWare ESXi 7.0 U2

DarwinAnim8or:
This is a good point, I'll look into it

udop1:
Just to add to this, I have a screenshot of what happens when someone tries to add friends. (This is using WSL, Ubuntu 20.04 LTS). Hopefully, it's helpful somewhat.

![image](https://user-images.githubusercontent.com/52384714/145637067-968c0598-ebe9-45bd-b5ff-9f27a6062fa7.png)

EmosewaMC:
This should be resolved on the latest commit.  If the issue persists please re-open

PR #432, opened 2022-02-04
refactor(docker): improve docker experience

TheNoim:
This PR should add some promised improvements for users who are using docker. 

My current goal is it to improve the following things:

- [X] Auto detect client type and root folder
- [X] Allow to manual set client type
- [X] Notify darkflame server if setup fails
- [X] Database hardening 
- [x] Wait for the database to be ready #417
- [x] Add a wrapper for `MasterServer -a` to make sure the server is ready
- [x] Create docker images via GitHub Actions and push it to github registry
- [ ] Update documentation


docker/setup.sh
```diff
@@ -62,33 +83,80 @@ update_ini_values
...
 fi
 
-if [[ ! -f "/client/extracted" ]]; then
+# Try to auto detect client type
+if [[ "${CLIENT_TYPE@L}" == "auto" ]]; then
+    if [[ -f "/client/legouniverse.exe" ]]; then
+        # Look for a unpacked file. If it doesn't exist, then we can assume the user provided a packed client
+        # with a client structure like a unpacked client. At this point we can't continue the process. 
+        # Without the versions directory lunpack can't extract the client files
+        if [[ ! -f "/client/res/maps/nexustower/nd_nexus_tower_7x1_audio.lvl" ]]; then
```
Jettford, line 95: I would use cdclient.fdb not a map file that might be changed by future content.

docker/setup.sh
```diff
@@ -62,33 +83,80 @@ update_ini_values
...
+            echo "[ERROR] Without the versions directory you need to unpack the client by yourself"
+            echo "[ERROR] The server can't boot without an unpacked client"
+            exit 13
+        fi
+        CLIENT_TYPE=unpacked
+        CLIENT_ROOT_DIR=/client
+        touch /docker/extracted
+    elif [[ -f "/client/client/legouniverse.exe" ]]; then
+        # If this file exist, we can assume the user provided an unpacked client but with a different client root structure
+        if [[ -f "/client/client/res/maps/nexustower/nd_nexus_tower_7x1_audio.lvl" ]]; then
```
Jettford, line 106: Same here

docker/setup.sh
TheNoim: I picked this particular file because I know that there will be issues when this file is missing. But, yeah, maybe cdclient.fdb is a better choice.

migrations/dlu/1_security.sql
```diff
@@ -0,0 +1,18 @@
+-- root password is set by the docker image / env variables, so no need to change it here
```
Jettford, line 1: Should this exist as a standard migration to run no matter the configuration?

TheNoim: Yes

Jettford: Surely this is just rerunning what people configure in mysql_secure_installation?

docker/setup.sh
Jettford: > I picked this particular file because I know that there will be issues when this file is missing
If cdclient.fdb is missing and the client is running unpacked then it will crash on start, it has to have this file where as it doesn't need some audio scene from NT

migrations/dlu/1_security.sql
TheNoim: Yeah, it is just a small change and not really 100% required. But it is recommend.

docker/setup.sh
TheNoim: > > I picked this particular file because I know that there will be issues when this file is missing
> 
> If cdclient.fdb is missing and the client is running unpacked then it will crash on start, it has to have this file where as it doesn't need some audio scene from NT

The client won't join the server if this file is missing (client missmatch, if I remember correctly). I will add `cdclient.fdb` as check.

docker/Dockerfile
```diff
@@ -42,11 +42,14 @@ RUN unzip /build/resources/navmeshes.zip -d /build/cmake_build/res/maps
 
 FROM gcc:11 as runtime
 
+ADD https://github.com/ufoscout/docker-compose-wait/releases/download/2.9.0/wait /wait
```
Xiphoseer, line 45: Please add a comment about what that does and why it's needed

migrations/dlu/1_security.sql
Xiphoseer: AFAICT, the global_priv table is an addition by MariaDB and may break in the face of an (Oracle) MySQL database.

In general, I'm not a fan of adding more vendor specific code, that moves the codebase further away from the ability to support other DBs like Postgres.

Maybe put setup-specific migrations in a subfolder named `mariadb`?

.env.example
```diff
@@ -1,5 +1,14 @@
 # Full path to the LEGO Universe client
+# You should provide the directory which contains the versions and client folder. 
+# If you don't a versions and client directory, provide the directory which contains the legouniverse.exe
```
Marcono1234, line 3:
```suggestion
# If you don't have a versions and client directory, provide the directory which contains the legouniverse.exe
```

Marcono1234:
Could you please also fix that the current Docker setup seems to be unable to properly connect the world server with the chat server? I think this has been mentioned somewhere else already here at this GitHub repository but I cannot find it.

The underlying issue might be that here the variable `$CHAT_SERVER_PORT` is used, but no value for it is set:
https://github.com/DarkflameUniverse/DarkflameServer/blob/b7d7146821d8e3f0dd711b7e7f91e39745993a9f/docker/setup.sh#L34

Additionally the default ports in the `.ini` files do not match the default ports in code, e.g. the port here in the [`WorldServer` code](https://github.com/DarkflameUniverse/DarkflameServer/blob/b8c1f7629e4dc351ed24a2a8618a9c398a22d475/dWorldServer/WorldServer.cpp#L195) is different than the default one in the [`worldconfig.ini`](https://github.com/DarkflameUniverse/DarkflameServer/blob/b7d7146821d8e3f0dd711b7e7f91e39745993a9f/resources/worldconfig.ini#L12). So when in one case you provide a port through the `.ini` file, and in the other case (as seen here) leave it blank, then the servers cannot connect. I assume that might be the intended behavior by the maintainers though.

(Also unless I am overlooking something in the log files, `WorldServer` does not seem to notice the connection failure to the chat server; but that is probably unrelated to the Docker setup.)

TheNoim:
@Marcono1234 can you try whether the latest commit resolves the issue or not?

docker/start_server.sh
```diff
@@ -43,7 +55,12 @@ else
     echo "Server already initialized"
 fi
 
+echo "Wait for database to be ready"
+
+/wait
```
Marcono1234, line 60: Not completely sure, but it looks like this might be causing the following now:
> [Warning] Aborted connection 3 to db: 'unconnected' user: 'unauthenticated' host: '172.18.0.5' (This connection closed normally without authentication)

I assume that can be ignored, but in case this really comes from `wait`, maybe there is a way to have it properly close the connection?

Marcono1234 (commented):
Thanks! I think it resolves the issue with the chat server. I was able to send a friend request once. However, ChatServer is crashing for me, but with the same stack trace as shown in https://github.com/DarkflameUniverse/DarkflameServer/issues/145#issuecomment-991275990, so I assume this is unrelated to the Docker setup.

Would be good if someone else experiencing the chat server issues or one of the maintainers could have a look whether it works as it should.

docker/start_server.sh
TheNoim: This doesn't matter at all. This is just how it works.

docker/setup.sh
```diff
@@ -1,5 +1,26 @@
 #!/bin/bash
 
+if [[ $SCRIPT_DEBUG == "1" || ${SCRIPT_DEBUG@L} =~ "true" ]]; then
```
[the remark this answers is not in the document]
TheNoim: > Sorry for my ignorance, but what is @L used for, or what does it mean?

It converts a string to lowercase

> Additionally, why do you use =~ "true"? If I understand it correctly =~ is used for regex pattern matching, but "true" is not a regex pattern (or rather a literal string).

Probably I wanted to check for a substring at some point and just forgot to change this. But hey, now even something like true with a space in front should work as value.

docker-compose.yml
```diff
@@ -1,103 +1,110 @@
...
+            context: .
+            dockerfile: ./docker/setup.Dockerfile
+        environment:
+            - DATABASE=${MARIADB_DATABASE:-darkflame}
+            - DATABASE_HOST=database
+            - DATABASE_USER=${MARIADB_USER:-darkflame}
+            - DATABASE_PASSWORD=${MARIADB_PASSWORD:-darkflame}
+            - EXTERNAL_IP=${EXTERNAL_IP:-darkflame}
+            - CLIENT_TYPE=${CLIENT_TYPE:-auto}
+            - CLIENT_ROOT_DIR=${CLIENT_ROOT_DIR}
```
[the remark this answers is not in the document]
TheNoim: Now it only gets overwritten if the variable is empty. This was intended so that you can set the path by yourself if you have an unusual setup.

docker/setup.sh
```diff
@@ -89,11 +89,16 @@ function fdb_to_sqlite() {
...
-    echo "Client not found."
-    echo "Did you forget to mount the client into the \"/client\" directory?"
+    echo "[ERROR] Client not found."
+    echo "[ERROR] Did you forget to mount the client into the \"/client\" directory?"
     exit 12
 fi
 
+if [[ "${CLIENT_TYPE@L}" != "packed" && "${CLIENT_TYPE@L}" != "unpacked" && "${CLIENT_TYPE@L}" != "auto" ]]; then
+    echo "[ERROR] Unknown CLIENT_TYPE"
+    exit 15
```
Marcono1234, line 99: Should this be unique? Because you are using the same exit code when `CLIENT_ROOT_DIR` was not detected.

TheNoim: Thanks, just quickly skimmed through the code to find out if this code is already used. Will correct it.

docker-compose.yml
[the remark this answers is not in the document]
Marcono1234: Ah ok. Do you want to set a default value of an empty string then here in the `docker-compose.yml` (for example `${CLIENT_ROOT_DIR-}`) to avoid the warning from Docker, and to avoid confusing users?

The current handling for `CLIENT_TYPE == auto` and a non-empty `CLIENT_ROOT_DIR` might be a bit problematic. The checks (such as `if [[ -f "/client/legouniverse.exe" ]]`) seem to ignore `CLIENT_ROOT_DIR`. So this could fail even though `CLIENT_ROOT_DIR` points to a valid directory. Would it make sense to consider the `CLIENT_ROOT_DIR` in the checks, or alternatively to simply fail if the user sets  `CLIENT_TYPE == auto` and a non-empty `CLIENT_ROOT_DIR`?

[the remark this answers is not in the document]
TheNoim: Hm, I expect `CLIENT_TYPE == auto` only if someone don't specifies `CLIENT_ROOT_DIR`.

> Would it make sense to consider the CLIENT_ROOT_DIR in the checks, or alternatively to simply fail if the user sets CLIENT_TYPE == auto and a non-empty CLIENT_ROOT_DIR

I think, this is the way to go :)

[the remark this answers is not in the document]
TheNoim: Ok, I added a warning. I don't think it should fail hard at this point, because it gets overwritten anyway.

docker/setup.sh
```diff
@@ -75,20 +163,20 @@ if [[ ! -f "/client/extracted" ]]; then
...
     echo "*.fdb" >> globs.txt
 
-    lunpack -g ./globs.txt /client/
+    lunpack -g ./globs.txt $CLIENT_ROOT_DIR/..
 
-    touch /client/extracted
+    touch /docker/extracted
 else
     echo "Client already extracted. Skip this step..."
     echo "If you want to force a re-extract, just delete the file called \"extracted\" in the client directory"
```
Marcono1234, line 171: Is this still correct? If I understand it correctly, with the changes of this PR the `extracted` and `migrated` file (see message below which needs to updated as well) are now created in the `shared_configs` volume. Therefore they won't exist in the `client` directory; unless someone used the previous Docker setup, but then the files simply have no effect.

docker/start_server.sh
```diff
@@ -43,7 +55,12 @@ else
...
+
 # start the server
 echo "Starting MasterServer"
 ./MasterServer
+echo "MasterServer exited with $?"
 tail -f /dev/null
```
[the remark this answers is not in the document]
TheNoim: I am still not sure if I want to keep `tail -f /dev/null`. I see pros and cons. Will think about it.

docker/setup.sh
TheNoim: Oh, now I understand what you mean. Yes, then you are completely right. The message isn't up to date .

TheNoim: I changed the message which should be more helpful.

Iterator256:
Is this blocked on something that one could help with? If I understand correctly a fix for #399 is included in this. It would be great if this could be merged eventually because as it stands right now the LU experience in docker isn't great because there is no way to add friends.

TheNoim:
> 

I just didn't have time for this. But if this is desperate needed, I could make the missing changes in another PR when I have time again.

brick-brick-brick:
@TheNoim Well I wouldn't call it "desperate", this is just a game after all. Thank you for your work on dockerizing all of this btw!

.github/workflows/docker.yml
```diff
@@ -0,0 +1,282 @@
...
+                  context: .
+                  push: true
+                  tags: ${{ steps.meta-darkflame-171023.outputs.tags }}
+                  labels: ${{ steps.meta-darkflame-171023.outputs.labels }}
+                  file: ./docker/Dockerfile
+                  build-args: |
+                      BUILD_VERSION=171023
+                      BUILD_THREADS=2
+
+    build-darkflame-171022:
```
Marcono1234, line 149: If the only difference here is 171022 vs 171023, would it then maybe be possible to use a [build matrix](https://docs.github.com/en/actions/using-jobs/using-a-build-matrix-for-your-jobs) to only have to define this job once?

dandeto:
Quick fix you may want to implement in this PR as well:
Step 1 in Docker.md says "Copy .env.example."
This is impossible since .dockerignore ignores .env.example in this repo :)

jgkawell:
> This is impossible since .dockerignore ignores .env.example in this repo :)

Shouldn't be an issue since `.dockerignore` only ignores files for the Docker build context. I bet you're thinking of `.gitignore` which doesn't track changes for ignored files and so they won't be included in the repo.

aronwk-aaron:
closing due to being stale

PR #437, opened 2022-02-06
Addressed specific race smash missions not progressing

EmosewaMC:
Missions like [1407](https://explorer.lu-dev.net/missions/1407) and [1406](https://explorer.lu-dev.net/missions/1406) use taskparam1 set to 17 to denote the mission as being smashing a specific smashable in a race.  Also addressed an issue where dragon egg smashables did not count towards generic smashing missions.  This is because the dragon eggs use a different smash server than generic race smashables.

Tested changes in Forbidden Valley and Nimbus Station race and the above missions progress without error.


Jettford (approved):

EmosewaMC:
@Wincent01 Believe you are the one to go to for racing and script stuff.  This PR addresses the racing smash missions not progressing and works without issue.  When you got a moment take a look if you can

PR #442, opened 2022-02-07
Fully implemented Avant Gardens Survival Buff Station

### LINKED ISSUE #422: BUG: AGS Buff Stations still don't spawn powerups

### Issue Description

The powerups make the sound effect and show the particle effect but nothing actually drops.

### Reproduction steps

build a buff station and wait

### Expected Behavior

the stations should spawn powerups.

### Environment

ubuntu server

EmosewaMC:
Fixes #422 
Fully implements the buff station and all functions.  Now correctly spawns powerups.

Tested on local instance and the spawner correctly spawns powerups and smashes itself without error.

### DIFF
--- a/dScripts/AgSurvivalBuffStation.cpp
+++ b/dScripts/AgSurvivalBuffStation.cpp
@@ -4,6 +4,7 @@
 #include "GameMessages.h"
 #include "SkillComponent.h"
 #include "dLogger.h"
+#include "TeamManager.h"
 
 void AgSurvivalBuffStation::OnRebuildComplete(Entity* self, Entity* target) {
     auto destroyableComponent = self->GetComponent<DestroyableComponent>();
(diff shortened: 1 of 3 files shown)


dScripts/AgSurvivalBuffStation.cpp
```diff
@@ -1,15 +1,43 @@
...
+    }
+    if (timerName == "DropLife") {
+        powerupToDrop = lifePowerup;
+        self->AddTimer("DropLife", dropLifeTimer);
+    }
+    if (timerName == "Dropimagination") {
+        powerupToDrop = imaginationPowerup;
+        self->AddTimer("Dropimagination", dropImaginationTimer);
+    }
+    if (target != nullptr) GameMessages::SendDropClientLoot(target, self->GetObjectID(), powerupToDrop, 0, self->GetPosition());
```
Marcono1234, line 42: Does this only create powerups for the player who built the Survival Buff Station? If so, then it might not be accurate to live, see https://youtu.be/iWe3KZv2TJE?t=92. The Survival Buff Station was built by a team member ([here](https://youtu.be/iWe3KZv2TJE?t=66)) but the powerups are available for all team members. Unfortunately I don't know what the exact correct behavior is; whether powerups appeared for every player, or whether they were "shared" somehow, so one player could steal them from another player.

I think the same also applied to the powerup spawners; but that is probably off-topic here and rather related to #344.

EmosewaMC: Second part is a separate issue.  First part is not apart of the original script.  Original script only sends the buffs to the player that built it.

EmosewaMC: Updated script to drop loot buffs for all members in a team.

EmosewaMC:
@MickVermeulen @averysumner If one of you two could give this a once over and review it that'd be great!  This has been tested on my local server and does the following:
spawns powerups (heal imagine armor) once every 2.5 seconds, destroys itself after 25 seconds and heals all players every 2.5 seconds within radius as well.  If theres anything else I missed please let me know!

Issue #408, opened 2022-01-23
BUG: Stuck at "Escape the Venture Explorer"


Nevyn19:
### Make sure you've done the following:

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### Issue Description

Until the conclusion of the mission everything seems OK.
Built the rocket, selected the rocket launch pad, rocket animation OK, and then nothing... the game hangs at this point:

![image](https://user-images.githubusercontent.com/65671010/150680587-b9c2cf6a-0b36-4dd4-ad6c-4553f8cff992.png)

If I force client quit and re-enter, it restarts from the moment I have to activate the escape pod, rocket animation and hang again.

### Reproduction steps

* Complete all missions on Venture Explorer
* Build the rocket
* Go to launch ramp and use the rocket

### Expected Behavior

Leave the base and going to Avant garderns

### Environment

* I have installed the server on my private local server (Ubuntu LTS 20.04).
* Client is: 0d862f71eedcadc4494c4358261669721b40b2131101cbd6ef476c5a6ec6775b (unpacked client, includes extra locales, rar compressed)
* DarkflameUniverse with fresh pull

This is the output of the server:

`[23-01-22 14:18:05] [AuthPackets] Received client version: 171022

[23-01-22 14:18:05] [WorldServer] User User authenticated with correct key.

[23-01-22 14:18:05] [Character] Loaded xmlData for character User (2)!

[23-01-22 14:18:10] [MasterServer] Received zone transfer req

[23-01-22 14:18:10] [InstanceManager] Searching for an instance for mapID 1000/0

[23-01-22 14:18:10] [MasterServer] Instance: 0/0/0 -> 0

[23-01-22 14:18:10] [MasterServer] Instance: 1100/0/2 -> 0

[23-01-22 14:18:10] [MasterServer] Instance: 1000/0/3 -> 1

[23-01-22 14:18:10] [MasterServer] Responding to transfer request 4 for zone 1000 0

[23-01-22 14:18:10] [MasterServer] Sent affirmation request 4 to 1000/0

[23-01-22 14:18:10] [MasterServer] Got affirmation request of transfer 4

[23-01-22 14:18:10] [MasterServer] Got affirmation of transfer 4

[23-01-22 14:18:10] [MasterServer] Affirmation complete 4

[...]

PR #443, opened 2022-02-07
Log WorldServer shutdown when MasterServer connection is missing

Xiphoseer:
Log when a WorldServer initiates shutdown because the master server connection went missing or wasn't established in the first place.


Xiphoseer:
cc #408

PR #447, opened 2022-02-08
Fix spam-log on master disconnect

Xiphoseer:
Only update and log shutdown for missing master connection once.

The log I added in #443 causes spams the log until the main loop terminates. This PR adds a second conditional that prevents this log if `worldShutdownSequenceStarted` is already set.

### DIFF
--- a/dWorldServer/WorldServer.cpp
+++ b/dWorldServer/WorldServer.cpp
@@ -313,7 +313,8 @@ int main(int argc, char** argv) {
 		if (!Game::server->GetIsConnectedToMaster()) {
 			framesSinceMasterDisconnect++;
 
-			if (framesSinceMasterDisconnect >= 30) {
+			int framesToWaitForMaster = ready ? 10 : 200;
+			if (framesSinceMasterDisconnect >= framesToWaitForMaster && !worldShutdownSequenceStarted) {
 				Game::logger->Log("WorldServer", "Game loop running but no connection to master for 30 frames, shutting down");
 				worldShutdownSequenceStarted = true;
 			}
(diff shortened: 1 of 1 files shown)


dWorldServer/WorldServer.cpp
```diff
@@ -313,8 +313,9 @@ int main(int argc, char** argv) {
 		if (!Game::server->GetIsConnectedToMaster()) {
 			framesSinceMasterDisconnect++;
 
-			if (framesSinceMasterDisconnect >= 30) {
-				Game::logger->Log("WorldServer", "Game loop running but no connection to master for 30 frames, shutting down");
+			int framesToWaitForMaster = ready ? 10 : 200;
+			if (framesSinceMasterDisconnect >= framesToWaitForMaster && !worldShutdownSequenceStarted) {
+				Game::logger->Log("WorldServer", "Game loop running but no connection to master for %d frames, shutting down", framesToWaitForMaster);
```
EmosewaMC, line 318: Should probably be a new line here

Xiphoseer: In the log message or after the call to `Log`? I'm not sure why you would want a newline? Width of terminals? C++ formatter rules?

EmosewaMC: I mean in the log - I noticed in a screenshot that the log just filled up the whole terminal window
image

Wincent01: There should indeed be a "\n" at the end of the log format

Xiphoseer: Ah, now I get it, thanks!

PR #449, opened 2022-02-09
Rehook dPhysics to main framerate

DarwinAnim8or:
Should fix bug causing double enter and exit events

### DIFF
--- a/dWorldServer/PerformanceManager.cpp
+++ b/dWorldServer/PerformanceManager.cpp
@@ -2,123 +2,99 @@
 
 #include "UserManager.h"
 
-#define HIGH 16
-#define MEDIUM 33
-#define LOW 66
+//Times are 1 / fps, in ms
+#define HIGH 16     //60 fps
+#define MEDIUM 33   //30 fps
+#define LOW 66      //15 fps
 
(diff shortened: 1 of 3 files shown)


DarwinAnim8or:
@Jettford if you could test this out that'd be great

EmosewaMC:
changes work fine on my local ubuntu instance and missions that would trigger 3 times before now only trigger once.  Death planes that would trigger based on phantom collisions now also only remove 1% of your coins instead of 3%

DarwinAnim8or:
Sounds good then, thank you @EmosewaMC !

Wincent01:
How much does this affect the performance? I implemented this to cut down on the physics steps needed for bigger worlds like Crux Prime.

EmosewaMC:
I didn't notice any change in performance (can test more worlds but I tested every type of world tick rate).  I can do some more aggressive testing if needed.

Wincent01:
It's probably not going to be noticeable, but the physics will take double the performance in some worlds. Would like to see a solution in the future to properly solve this issue. Physics doesn't need the same step rate, but I'm fine which this as a temporary fix as it seemed to cause double triggers.

PR #451, opened 2022-02-10
Shooting Gallery daily mission fix

### LINKED ISSUE #448: BUG:

### Issue Description

Daily Mission for Cannon Cove Shoting Galery not work.

### Reproduction steps

I Accept this daily Mission from Venture League Vendor on Nimbus Plaza. 

### Expected Behavior

After accepting the task and starting the shooting gallery mini-game, points in the task are not counted.

### Environment

Server runing on Windowsvia WSL 1

EmosewaMC:
Fixes #448 
Not the best change however any shooting gallery missions that require the player to earn a specific score in one game for some reason fail to keep their state between world changes unlike other missions.  This has been fixed in this PR alongside a couple other bugs I noticed while working on this:
1) MakeReadyToComplete being called instead of a hard flag set to MISSION_STATE_READY_TO_COMPLETE since this method is already implemented and properly sets repeatable missions to the correct complete state.
2) These lines `return GetType() == MissionTaskType::MISSION_TASK_TYPE_MINIGAME ? progress == info->target : progress >= info->targetValue;` in MissionTask made absolutely no sense because info->target is a task target not a target _value_.  Both of these ternary operators have been removed to properly progress minigame tasks.
3) When progressing MissionTasks, again, for some reason the values were being set to info->target and not info->targetValue.  This has been corrected.
4) Bodge mission progression in SGCannon has been removed now that the task can properly complete with the main change of the PR (echoing SetProgress for minigame missions)
5) Comments that were deprecated have been removed.
6) For some reason SGCannon was resetting its score in GameOver and also in ResetVars.  This has been changed so it only happens once.

### DIFF
--- a/dGame/dMission/MissionTask.cpp
+++ b/dGame/dMission/MissionTask.cpp
@@ -78,14 +78,7 @@ void MissionTask::SetProgress(const uint32_t value, const bool echo)
 
 	std::vector<float> updates;
 	updates.push_back(static_cast<float>(progress));
-
-	GameMessages::SendNotifyMissionTask(
-		entity,
-		entity->GetSystemAddress(),
-		static_cast<int>(info->id),
-		static_cast<int>(1 << (mask + 1)),
-		updates
-	);
+	GameMessages::SendNotifyMissionTask(entity,	entity->GetSystemAddress(), static_cast<int>(info->id), static_cast<int>(1 << (mask + 1)), updates);
 }
 
 
@@ -190,15 +183,13 @@ bool MissionTask::InParameters(const uint32_t value) const
 
 bool MissionTask::IsComplete() const
 {
-    // Minigames are the only ones where the target value is a score they need to get but the actual target is the act ID
-	return GetType() == MissionTaskType::MISSION_TASK_TYPE_MINIGAME ? progress == info->target : progress >= info->targetValue;
+	return progress >= info->targetValue;
 }
 
 
 void MissionTask::Complete()
 {
-    // Minigames are the only ones where the target value is a score they need to get but the actual target is the act ID
-	SetProgress(GetType() == MissionTaskType::MISSION_TASK_TYPE_MINIGAME ? info->target : info->targetValue);
+	SetProgress(info->targetValue);
 }
 
 
@@ -354,7 +345,7 @@ void MissionTask::Progress(int32_t value, LWOOBJID associate, const std::string&
 		}
 
 		if(info->targetGroup == targets && value >= info->targetValue) {
-			SetProgress(info->target);
+			SetProgress(info->targetValue);
 			break;
 		}
 		break;
--- a/dGame/dMission/Mission.cpp
+++ b/dGame/dMission/Mission.cpp
@@ -109,7 +109,7 @@ void Mission::LoadFromXml(tinyxml2::XMLElement* element) {
         } else {
             const auto value = std::stoul(task->Attribute("v"));
 
-            m_Tasks[index]->SetProgress(value, false);
+            m_Tasks[index]->SetProgress(value, m_Tasks[index]->GetType() == MissionTaskType::MISSION_TASK_TYPE_MINIGAME);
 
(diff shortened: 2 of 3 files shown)


dGame/dMission/MissionTask.cpp
```diff
@@ -352,9 +343,16 @@ void MissionTask::Progress(int32_t value, LWOOBJID associate, const std::string&
 		if (info->target != gameID) {
 			break;
 		}
-
+		// Lego Universe is a great game with very easy to understand mission progression.
+		// If we are progressing a mission for shooting gallery we need to set its
+		// completion to 1 and not the targetValue beacuse that makes sense
+		// compared to all other missions and tasks :)
+		if(info->targetGroup == targets && value >= info->targetValue && GetMission()->IsMission() && info->target == 1864 && info->targetGroup == "performact_score") {
```
EmosewaMC, line 350: Im not sure if this whole if statement is needed, however i am being more safe than sorry here.  Something needs to be done to the charxml to fix players who are in a state where this value is larger than it should be.

MickVermeulen:
Can you confirm you've tested this on _all_ minigames? It's been a while but I recall having those ternaries in there as for some minigames exact scores were required where for other minigames minimum scores were required

EmosewaMC:
If that was true this was setting it to the wrong value.  Info->target is the activity ID target of the mission and not the task target value.  So for shooting gallery's case this always set progress to 1864 and would not set it to the value for the mission.  Ill test other minigames later.

EmosewaMC:
Tested all changes on some achievements from ag footraces and ns time smashes as well as ag survival.  All these missions completed as expected.

dGame/dMission/MissionTask.cpp
```diff
@@ -78,14 +78,7 @@ void MissionTask::SetProgress(const uint32_t value, const bool echo)
...
 	updates.push_back(static_cast<float>(progress));
-
-	GameMessages::SendNotifyMissionTask(
-		entity,
-		entity->GetSystemAddress(),
-		static_cast<int>(info->id),
-		static_cast<int>(1 << (mask + 1)),
-		updates
-	);
+	GameMessages::SendNotifyMissionTask(entity,	entity->GetSystemAddress(), static_cast<int>(info->id), static_cast<int>(1 << (mask + 1)), updates);
```
MickVermeulen, line 81: What's up with the odd spacing in the function call here?

dGame/dMission/MissionTask.cpp
```diff
@@ -352,9 +343,16 @@ void MissionTask::Progress(int32_t value, LWOOBJID associate, const std::string&
 		if (info->target != gameID) {
 			break;
 		}
-
+		// Lego Universe is a great game with very easy to understand mission progression.
+		// If we are progressing a mission for shooting gallery we need to set its
+		// completion to 1 and not the targetValue beacuse that makes sense
+		// compared to all other missions and tasks :)
```
MickVermeulen, line 349: I think the comment is funny but going forward I'd prefer a more pragmatic and serious explanation here

MickVermeulen (commented):
Have some minor comments, all the progression stuff doesn't look off to me and I trust that it works based on your testing

dGame/dMission/MissionTask.cpp
EmosewaMC: Likely just me accidentally tabbing while editing the file.  Can fix later

dGame/dMission/MissionTask.cpp
EmosewaMC: I'll clean this up.  Now that I've settled from the discovery I'll remove the snarky comment.

MickVermeulen: It's still good to leave _a_ comment in there about why the decision was made, just less subjective is already good :)

EmosewaMC:
@MickVermeulen Updated comments and whitespacing.

MickVermeulen (approved):
LGTM

PR #452, opened 2022-02-10
Addressed GM Operators not being able to spawn in cars in non-racing worlds

### LINKED ISSUE #351: BUG: gmlevel 9 Does Not Spawn Cars Properly

### Issue Description

When a player drags a built race car from their backpack to the game, the server will spawn the player in the car and lock them in place.  This would require intervention to free the player as they are now stuck in a car that cant do anything.  A server restart would fix the player being stuck.

A larger issue caused by this is there is no limit to the number of cars a player can spawn with this method and this will lead to degraded server performance until the server is restarted which would clear all of the cars.

image

### Reproduction steps

1) Build a car using modular building
2) Drag the car onto the screen repeatedly

### Expected Behavior

A pop saying you must be near a modular car build site should pop up

### Environment

Server: Ubuntu
Client: Windows 10

EmosewaMC:
Fixes #351 
Addresses an issue where GMs were unable to spawn in cars in non-racing worlds.   Tested all changes in Nimbus Station without any issues.  The way this works is almost identical to how players are loaded into cars for races and loaded out of cars in races.  Dying currently does not work and I dont know how to fix it however unequipping and re-equipping the car fixes the issue and respawns the player.

### DIFF
--- a/dGame/dComponents/InventoryComponent.cpp
+++ b/dGame/dComponents/InventoryComponent.cpp
@@ -916,7 +916,7 @@ void InventoryComponent::EquipItem(Item* item, const bool skipChecks)
 
 			auto startRotation = NiQuaternion::LookAt(startPosition, startPosition + NiPoint3::UNIT_X);
 			auto angles = startRotation.GetEulerAngles();
-			angles.y -= M_PI;
+			angles.y -= PI;
 			startRotation = NiQuaternion::FromEulerAngles(angles);
 
 			GameMessages::SendTeleport(m_Parent->GetObjectID(), startPosition, startRotation, m_Parent->GetSystemAddress(), true, true);
--- a/dGame/dComponents/InventoryComponent.cpp
+++ b/dGame/dComponents/InventoryComponent.cpp
@@ -23,6 +23,7 @@
 #include "CharacterComponent.h"
 #include "dZoneManager.h"
 #include "PropertyManagementComponent.h"
+#include "DestroyableComponent.h"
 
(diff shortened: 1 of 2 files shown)


dGame/dComponents/InventoryComponent.cpp
```diff
@@ -1019,6 +1052,14 @@ void InventoryComponent::EquipItem(Item* item, const bool skipChecks)
 
 void InventoryComponent::UnEquipItem(Item* item)
 {
+	// if (item->GetLot() == 8092 && m_Parent->GetGMLevel() >= GAME_MASTER_LEVEL_OPERATOR && hasCarEquipped == true)
```
Jettford, line 1055: I would just delete this code, will likely never be used.

EmosewaMC: Ah yeah, was going to try to use this function at some point but forgot to deprecate it

averysumner:
Doesn’t build on windows, please address.

Jettford:
(The issue is the use of M_PI in the InventoryComponent, replace that with the constant present in [dCommonVars](https://github.com/DarkflameUniverse/DarkflameServer/blob/main/dCommon/dCommonVars.h#L51))

EmosewaMC:
> (The issue is the use of M_PI in the InventoryComponent, replace that with the constant present in [dCommonVars](https://github.com/DarkflameUniverse/DarkflameServer/blob/main/dCommon/dCommonVars.h#L51))

Thanks for the reminder @Jettford .  Just been putting this off for awhile due to other more pressing matters

Jettford:
Issues found by previous reviewers solved.

Merging.

PR #456, opened 2022-02-12
Fetch zlib library over HTTPS

Marcono1234:


averysumner:
Have you tested to ensure it still works when changed to HTTPS?

Marcono1234:
To be honest, I had not tried it yet because GitHub supports HTTPS and I had checked in the browser that this URL can be reached over HTTPS as well. Additionally the workflow run would have caught it, in case this does not work.

Since this only affects the Windows build and I don't have a setup for building on Windows locally at the moment, I have let GitHub build it on a separate branch, [and the build succeeded](https://github.com/Marcono1234/DarkflameServer/actions/runs/1855607223).

(I have now also rebased this PR onto the latest `main` changes because by accident I had based it on an older version; though this does not have any effect on the changes)

Jettford:
I am confused on why this only affects Windows? also zlib should probably be removed from the Windows building as it is broken and isn't used

averysumner:
> I am confused on why this only affects Windows? also zlib should probably be removed from the Windows building as it is broken and isn't used

No, zlib should stay and be fixed. Windows is useful for testing.

averysumner:
> To be honest, I had not tried it yet because GitHub supports HTTPS and I had checked in the browser that this URL can be reached over HTTPS as well.
> 
> Since this only affects the Windows build and I don't have a setup for building on Windows locally at the moment, I have let GitHub build it on a separate branch, [and the build succeeded](https://github.com/Marcono1234/DarkflameServer/actions/runs/1855607223).
> 
> (I have now also rebased this PR onto the latest `main` changes because by accident I had based it on an older version; though this does not have any effect on the changes this PR makes)

If you're not willing to test this on Windows I'm just going to close this PR. It's a fine change as long as someone can test it; i don't have a Windows server setup right now either.

Marcono1234:
Feel free to close this PR; though I don't see how downloading a ZIP file over HTTPS instead of HTTP from GitHub could in any way negatively affect the build.

averysumner:
> Feel free to close this PR; though I don't see how downloading a ZIP file over HTTPS instead of HTTP from GitHub could in any way negatively affect the build.

Adding encryption complicates downloads, and also means that they could break if for some reason a TLS certificate expires upstream and isn't replaced. I am not too familiar with CMake's FetchContent and I don't know how it works with HTTPS. I'll keep this open and try to test it at some point soon.

Jettford:
You know that the above FetchContent call does use HTTPS right?

EmosewaMC:
Testing this on Windows today.  If it builds there then I will merge this.

EmosewaMC:
This built and compiled the binary on windows and was able to start the server on windows.  Merging

Issue #350, opened 2022-01-04
BUG: Server Crashes with Assertion Error `GeneralUtils::ASCIIToUTF16` Assertion `c > 0 && c <= 127`


TheMatt2:
### Make sure you've done the following:

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### Issue Description

The server will crash if `GeneralUtils::ASCIIToUTF16`  is passed a string that falls outside of this range.
The server does not actually exit, but hangs. Any player trying to login will be stuck on 0% in world load.

Specifically the server seems to be crashing on processing the text `8066 - Model Reward Café Table`.
The character "é" seems to evaluate to the ascii value 233, which falls outside of the assertion made by `GeneralUtils::ASCIIToUTF16` that all values are less than 127.

### Reproduction steps

A user with appropriate permission can type into the char window:

    /lookup table

The server will crash a few seconds later.

### Expected Behavior

Accented characters should be loaded correctly from ASCII to UTF16.
On assertion failure, the server should cleanly exit.

### Environment

Ubuntu under Docker

Traceback
```
darkflameserverdev     | WorldServer: /DarkflameServer/dCommon/GeneralUtils.cpp:53: std::u16string GeneralUtils::ASCIIToUTF16(const string&, size_t): Assertion `c > 0 && c <= 127' failed.
darkflameserverdev     | Fatal error 6
darkflameserverdev     | Stacktrace:
darkflameserverdev     | ./WorldServer(+0xe739f)[0x560b8ae5939f]
darkflameserverdev     | /lib/x86_64-linux-gnu/libpthread.so.0(+0x153c0)[0x7f8f35a1b3c0]
darkflameserverdev     | /lib/x86_64-linux-gnu/libc.so.6(gsignal+0xcb)[0x7f8f3550518b]
darkflameserverdev     | /lib/x86_64-linux-gnu/libc.so.6(abort+0x12b)[0x7f8f354e4859]
darkflameserverdev     | /lib/x86_64-linux-gnu/libc.so.6(+0x25729)[0x7f8f354e4729]
darkflameserverdev     | /lib/x86_64-linux-gnu/libc.so.6(+0x36f36)[0x7f8f354f5f36]
[...]

PR #457, opened 2022-02-13
Add server version command

Marcono1234:
Adds a `/server-version` command which prints the server version as well as the Git commit SHA in chat:
Chat screenshot

The version and the Git commit SHA are stored in the `worldconfig.ini` file and are set during the CMake build.
This feature is not super useful because apparently you cannot copy text from chat (unless I am overlooking something), but maybe it is a bit helpful nonetheless. If you know a way to allow the user to copy the text (possibly with something other than the chat), please let me know; that might be helpful in combination with #419 to make it easier for the user to obtain the version information.

This pull request also changes the Docker setup a bit to use the `.ini` files prepared during build instead of having the DarkflameSetup container copy the originals and then share them with the server. I have marked this as draft to wait until #432 is merged.

This is mainly a proof of concept; no worries if you are not interested in this. Any feedback is appreciated.

Note: These changes might cause issues with existing local `CMakeCache.txt` files.

### DIFF
--- a/dGame/dUtilities/SlashCommandHandler.cpp
+++ b/dGame/dUtilities/SlashCommandHandler.cpp
@@ -26,6 +26,7 @@
 #include "UserManager.h"
 #include "BitStream.h"
 #include "dCommonVars.h"
+#include "dConfig.h"
 #include "GeneralUtils.h"
 #include "Entity.h"
 #include "EntityManager.h"
@@ -260,6 +261,15 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 
 		// Fix the destroyable component
 		destroyableComponent->FixStats();
+	}
+
+	if (chatCommand == "server-version")
+	{
+		auto& serverName = Game::config->GetValue("server_name");
+		auto& serverVersion = Game::config->GetValue("server_version");
+		auto& gitCommitSha = Game::config->GetValue("git_commit_sha");
+		ChatPackets::SendSystemMessage(sysAddr, GeneralUtils::ASCIIToUTF16(serverName) + u" " + GeneralUtils::ASCIIToUTF16(serverVersion) + u" (" + GeneralUtils::ASCIIToUTF16(gitCommitSha) + u")");
+		return;
 	}
 
 	if (chatCommand == "credits" || chatCommand == "info")
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,15 +6,8 @@ services:
     build:
       context: .
       dockerfile: ./docker/setup.Dockerfile
-    environment:
-      - DATABASE=${MARIADB_DATABASE:-darkflame}
-      - DATABASE_HOST=database
-      - DATABASE_USER=${MARIADB_USER:-darkflame}
-      - DATABASE_PASSWORD=${MARIADB_PASSWORD:-darkflame}
-      - EXTERNAL_IP=${EXTERNAL_IP:-darkflame}
     volumes:
       - ${CLIENT_PATH:?missing_client_path}:/client
-      - shared_configs:/docker/
 
   database:
     container_name: DarkflameDatabase
@@ -44,9 +37,14 @@ services:
       args:
         - BUILD_THREADS=${BUILD_THREADS:-1}
         - BUILD_VERSION=${BUILD_VERSION:-171022}
+    environment:
+      - DATABASE=${MARIADB_DATABASE:-darkflame}
+      - DATABASE_HOST=database
+      - DATABASE_USER=${MARIADB_USER:-darkflame}
+      - DATABASE_PASSWORD=${MARIADB_PASSWORD:-darkflame}
+      - EXTERNAL_IP=${EXTERNAL_IP:-darkflame}
     volumes:
       - ${CLIENT_PATH:?missing_client_path}:/client:ro
-      - shared_configs:/shared_configs:ro
     depends_on:
       - database
     ports:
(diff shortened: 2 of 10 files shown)


averysumner:
A better solution would be to configure a header file with a server version and server commit macro.

DarwinAnim8or:
I agree with avery, the version is already defined as a preprocessor define, see:

- [CMakeVariables.txt](https://github.com/DarkflameUniverse/DarkflameServer/blob/main/CMakeVariables.txt)

You could add the git commit hash there, and also include the NET_VERSION in your logger. Maybe even include the compile time by simply using the C++ macro __TIMESTAMP__ (as used in dWorldServer.cpp)

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -262,6 +263,15 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 		destroyableComponent->FixStats();
 	}
 
+	if (chatCommand == "server-version")
+	{
+		auto& serverName = Game::config->GetValue("server_name");
+		auto& serverVersion = Game::config->GetValue("server_version");
+		auto& gitCommitSha = Game::config->GetValue("git_commit_sha");
+		ChatPackets::SendSystemMessage(sysAddr, GeneralUtils::ASCIIToUTF16(serverName) + u" " + GeneralUtils::ASCIIToUTF16(serverVersion) + u" (" + GeneralUtils::ASCIIToUTF16(gitCommitSha) + u")");
```
Xiphoseer, line 271: Don't use this function for text from a config file. It crashes when eg. `serverName` contains non-ascii characters.

docker-compose.yml
```diff
@@ -6,15 +6,8 @@ services:
     build:
       context: .
       dockerfile: ./docker/setup.Dockerfile
-    environment:
-      - DATABASE=${MARIADB_DATABASE:-darkflame}
-      - DATABASE_HOST=database
-      - DATABASE_USER=${MARIADB_USER:-darkflame}
-      - DATABASE_PASSWORD=${MARIADB_PASSWORD:-darkflame}
-      - EXTERNAL_IP=${EXTERNAL_IP:-darkflame}
```
Xiphoseer, line 14: These changes seem unrelated?

docker/setup.Dockerfile
```diff
@@ -13,11 +13,10 @@ RUN apk add sqlite bash --no-cache
 WORKDIR /setup
 
 # copy needed files from repo
-COPY resources/ resources/
```
Xiphoseer, line 16: Is this PR based on some other branch?

dGame/dUtilities/SlashCommandHandler.cpp
Marcono1234: Yes you are right; I guess this is blocked then by #350 since there exists no conversion function for UTF-8 to UTF-16 yet?

docker-compose.yml
Marcono1234: These changes were required to perform variable substitution in the `.ini` files. Previously the `.ini` files were copied from the main repository to the _setup_ service which then used these variables to update the `.ini` files which were then shared through a volume with the _darkflame_ (server) service. The `.ini` files prepared during CMake build were simply discarded.

Therefore I have moved all these environment variables to the _darkflame_ service. Otherwise the original `.ini` files without substituted variables would be used.

docker/setup.Dockerfile
Marcono1234: See https://github.com/DarkflameUniverse/DarkflameServer/pull/457#discussion_r808258121 above.

dGame/dUtilities/SlashCommandHandler.cpp
Xiphoseer: Technically it exists without that as std::codecvt_utf8_utf16.

averysumner:
> Should the source value [defined in worldconfig.ini](https://github.com/DarkflameUniverse/DarkflameServer/blob/42f6f2f10b5971dd13faa18e2018892ce21ce3c3/resources/worldconfig.ini#L9) then for consistency also be moved to CMakeVariables.txt?

I guess `source` could technically be gleaned from git origin, but that may be messy. If you want to add the git commit hash, use a configured header file, there's no reason for it to be in a configuration file.

> Is __TIMESTAMP__ really the correct choice here? It seems to describe the time when the file was last modified (see [documentation](https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html)). Maybe a combination of __DATE__ and __TIME__ might be more appropriate.

As far as I'm aware these macros aren't compiler agnostic and shouldn't be used. A better solution would be to use CMake to get the time and put that in a configured header.

Jettford:
I am extremely confused on why this PR includes significant amount of Docker changes, that seem unrelated. Is there a reason for this?

Marcono1234:
Sorry, maybe I should have explained the Docker changes a bit more in detail in the description. In https://github.com/DarkflameUniverse/DarkflameServer/pull/457#discussion_r808282710 I gave more information on this.

codeshaunted:
@Marcono1234 will you be improving this with the requested changes or can we close this?

Marcono1234:
> @Marcono1234 will you be improving this with the requested changes or can we close this?

Sorry, I think I won't continue with this pull request. If you want you can of course use the changes of these pull request as starting point for this feature, implemented in the way you prefer. But for me with my somewhat limited C++ knowledge and without knowing exactly what you have in mind, I don't think it would be very useful to continue here.

Thanks nonetheless for your extensive feedback!

PR #474, opened 2022-03-24
Addressed hard coded character create values for shirts and pants

EmosewaMC:
Previously, character create used hard coded vectors and ids to get the LOTs for items to be used for character creation.  With this PR the CDServer is queried and the resulting LOT from the query returns the LOT of the shirt that is requested.  This is done by merging the ItemComponent and ComponentsRegistry tables together and getting all LOTs with an ItemComponent of 11.  Then this table is merged with the Objects table where we get the LOT which has an internal note of, for shirts, character create shirt, or for pants, cc pants, and the requested decal and/or color1 the client had chosen.  Should this query return nothing or fail, a default shirt and pants are chosen.  This allows users to add new shirts and pants to the CDServer/cdclient without needing to hard code them.

If this query can be cached for efficiency I'll happily do so.  

Tested creating characters with various shirts, pants and colors.  Server created the character with the correct shirt and pants and did not fall into the fail case.

### DIFF
--- a/dGame/UserManager.cpp
+++ b/dGame/UserManager.cpp
@@ -6,6 +6,7 @@
 
 #include "Database.h"
 #include "Game.h"
+#include "dLogger.h"
 #include "User.h"
 #include <WorldPackets.h>
 #include "Character.h"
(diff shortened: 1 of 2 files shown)


DarwinAnim8or:
This isn't "hot code" so to speak, so it's fine to leave it as is so long as it works properly. 
Thank you for finally fixing this though!

PR #485, opened 2022-04-01
Added support to soft delete mail

EmosewaMC:
Added support through a column in the MySQL database to soft delete mail from players so it is kept for moderation purposes.  System mail is still hard deleted however so as to not clog up the database.

Tested sending mail to an offline player, deleting mail from a player, deleting system mail and recieving mail from system and players.  All worked as intended.  

Another migration has been added as well for the DLU database to make the column available.

### DIFF
--- a/dGame/dUtilities/Mail.cpp
+++ b/dGame/dUtilities/Mail.cpp
@@ -282,8 +282,9 @@ void Mail::HandleSendMail(RakNet::BitStream* packet, const SystemAddress& sysAdd
 }
 
 void Mail::HandleDataRequest(RakNet::BitStream* packet, const SystemAddress& sysAddr, Entity* player) {
-	sql::PreparedStatement* stmt = Database::CreatePreppedStmt("SELECT * FROM mail WHERE receiver_id=? limit 20;");
+	sql::PreparedStatement* stmt = Database::CreatePreppedStmt("SELECT * FROM mail WHERE receiver_id= ? AND is_deleted = ? LIMIT 20;");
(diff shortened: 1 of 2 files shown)


DarwinAnim8or:
I'd like to add to this that due to privacy laws, you cannot keep user data for longer than you're supposed to. 
All of these mails should be set up to be deleted automatically after say 90 days?

EmosewaMC:
> I'd like to add to this that due to privacy laws, you cannot keep user data for longer than you're supposed to. All of these mails should be set up to be deleted automatically after say 90 days?

Understandable.  This would need to be detected then by a SQL job afaict in order to do this.  Since this would require the user to know to do this and set it up should this PR be closed then?  Not sure how to detect and delete rows like this without a job at the moment.

DarwinAnim8or:
Yeah I think the only compromise that could be made here is that MasterServer could scan for old messages and delete them upon startup, but that's not an ideal solution either.

EmosewaMC:
> Yeah I think the only compromise that could be made here is that MasterServer could scan for old messages and delete them upon startup, but that's not an ideal solution either.

Agreed.  Im sure there is an ideal solution for this that doesnt involve stealth editing a table on startup of master server.  Will mark as a draft until then.

averysumner:
> I'd like to add to this that due to privacy laws, you cannot keep user data for longer than you're supposed to. All of these mails should be set up to be deleted automatically after say 90 days?

I mean it's questionably legal/a good idea to be hosting a public server in any official capacity. Do we really care about this?

PR #487, opened 2022-04-01
Added a GM send after QB complete

EmosewaMC:
This PR makes it so when a player completes a quick build, they play a little celebration animation.  This has been verified against the packet caps from live and is sent when the player completes a quick build as you see here.  I also spotted an error with the SendRebuildNotifyState GM that would incorrectly send that a quickbuild that is beginning to be built to a state of "COMPLETED" when the state should be set to "BUILDING".

Tested a couple of quick builds in nimbus station and Roo Morgg Anchor and the player celebrated without issue.

### DIFF
--- a/dGame/dComponents/RebuildComponent.cpp
+++ b/dGame/dComponents/RebuildComponent.cpp
@@ -380,7 +380,7 @@ void RebuildComponent::StartRebuild(Entity* user) {
 
 		EntityManager::Instance()->SerializeEntity(user);
 
-		GameMessages::SendRebuildNotifyState(m_Parent, m_State, eRebuildState::REBUILD_COMPLETED, user->GetObjectID());
+		GameMessages::SendRebuildNotifyState(m_Parent, m_State, eRebuildState::REBUILD_BUILDING, user->GetObjectID());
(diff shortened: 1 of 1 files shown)


DarwinAnim8or (approved):
Could these state changes also be causing other bugs with QBs?
Either way, this is correct in that there was supposed to be a celebration anim

EmosewaMC:
> Could these state changes also be causing other bugs with QBs?

As in my changes adding bugs or the change fixing bugs?

DarwinAnim8or:
The latter, fixing possible bugs currently present because the state isn’t correct.
I guess we’ll see if other bugs pop up because of the state change 😝

Van: David ***@***.***>
Verzonden: zondag 3 april 2022 11:29
Aan: ***@***.***>
CC: ***@***.***>; ***@***.***>
Onderwerp: Re: [DarkflameUniverse/DarkflameServer] Added a GM send after QB complete (PR #487)


Could these state changes also be causing other bugs with QBs?

As in my changes adding bugs or the change fixing bugs?

EmosewaMC:
Ill do some more regression testing before merging this then.  I seem to semi-consistently get this bug (being the QBs unable to be built) with the anchors that apes spawn so I will look into this a bit.

PR #490, opened 2022-04-06
resolved MacOS compilation warnings

EmosewaMC:
Resolved numerous compilation warnings on MacOS and also fixed a couple of errors:
Numerous missing overrides for overridden functions
Trading manager was doing an assignment instead of a comparison for some reason so this has been fixed.
BaseConsoleTeleportServer was not returning should the RocketLaunchpadComponents be null.  This has been resolved.

### DIFF
--- a/dGame/TradingManager.cpp
+++ b/dGame/TradingManager.cpp
@@ -60,7 +60,7 @@ void Trade::SetCoins(LWOOBJID participant, uint64_t coins)
     {
         m_CoinsA = coins;
     }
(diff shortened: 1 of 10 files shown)


DarwinAnim8or:
Is this for MacOS in general, or also for ARM?

Jettford (approved):

EmosewaMC:
> Is this for MacOS in general, or also for ARM?

This resolved warnings on my 2012 mac, my newer mac had the same warnings as well so I assume so?

PR #493, opened 2022-04-07
Pets no longer cause client hitches in Nimbus Station and Forbidden Valley

EmosewaMC:
Previously due to an error, pets were given item components even when pets need to be excluded from having item components.  This would cause serialization issues on the client and the client would lag when loading pets in Nimbus Station and Forbidden valley.  This also fixes the issue with the "You got X number of bricks out of Y!" error.  This was happening because the client does all the work for checking the bricks and if the bricks are all correct, only then does the server send a TrybuildResult.  Otherwise nothing should be sent.  

All work is verified against packet captures from live, lcdr's toolset, and testing.
testing involved 2 clients and various combinations of taming pets and loading pets from bags.  No behavior has changed for pets besides no more lag in the critical areas.

### DIFF
--- a/dGame/Entity.cpp
+++ b/dGame/Entity.cpp
@@ -231,7 +231,8 @@ void Entity::Initialize()
 		m_Components.insert(std::make_pair(COMPONENT_TYPE_RACING_STATS, nullptr));
 	}
 
-	if (compRegistryTable->GetByIDAndType(m_TemplateID, COMPONENT_TYPE_ITEM) > 0) {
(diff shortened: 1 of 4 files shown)


EmosewaMC:
Requested a few reviewers just to make sure nothing adverse was affected since this does touch Entity.cpp construction.

EmosewaMC:
Merging because this has been running just fine on Brick Mesa for about 5 days now and I believe the change is correct according to the info provided by lcdr.

PR #503, opened 2022-04-12
Dynamically check if zone is viable from existence of file

aronwk-aaron:
Removes hard-coded valid zones and checks against the map file exists.

(ignore all the white-space, my linter was overly aggressive, can remove all that if wanted)

Tested with valid maps, invalid maps with no file path in cdserver, and invalid maps with a file path in cdserver but no map on disk


aronwk-aaron:
@EmosewaMC did some more testing on this, as requested. and looks good

Some other ways of checking this would be [faster](https://stackoverflow.com/questions/12774207/fastest-way-to-check-if-a-file-exists-using-standard-c-c11-14-17-c), but fstream is already used in this location.

Additionally, a close is not needed, since [fstream closes itself when it leaves a scope](https://stackoverflow.com/questions/748014/do-i-need-to-manually-close-an-ifstream)

EmosewaMC (approved):
LGTM.  For future reference lets keep whitespace changes and feature changes in separate PRs.

PR #504, opened 2022-04-13
Prevent integer underflow in item removal

NinjaOfLU:
Previously, the only check that the user wasn't trashing more items than they had was clientsided, and this could be bypassed by contacting the server to remove items via a console or the like, and then trashing them before the server could respond, resulting in the count for the items being less than iStackCount. This check prevents that underflow.


NinjaOfLU:
Noted some other issues with the Maelstrom brick console not updating mission items, whilst testing this works, but this PR hasn't touched the code for that, so that would appear to be an existing issue (I assume it'd just need to be added into the script). In short, this PR has now been built, tested, and confirmed to work (turns out that I _can_ do basic maths - I think).

PR #508, opened 2022-04-17
Frostburgh Server fixes

aronwk-aaron:
Implement some missing scripts that Frostburgh uses


aronwk-aaron:
Current Status:

WhFans is base on AgFans
Since we have both of these lua scripts in the client, comparing them adjusting according should "work", since WhFans is just a trimmed down AgFans
However, the world crashes on startup when calling ToggleFX in the OnStartup

RockHydrantBroken is based on HydrantBroken
We don't have the lua script for HydrantBroken, but given that we have the cpp script we can use that as a starting point.
Current state is that the animation plays the animation, but the bouncers that correlate to this aren't found

dScripts/WhFans.cpp
```diff
@@ -0,0 +1,66 @@
...
+	ToggleFX(self, false);
+}
+
+void WhFans::ToggleFX(Entity* self, bool hit) {
+	std::string fanGroup = "";
+	try {
+		fanGroup = self->GetGroups()[0];
+	} catch(...) {
+		fanGroup = "";
+	}
```
Wincent01, line 18: Do not use catch (...) unless absolutely necessary. Do something like
```c++
std::string fanGroup;
const auto& groups = self->GetGroups();
if (!groups.empty()) {
    fanGroup = groups[0];
} else {
    fanGroup = "";
}
```

dScripts/RockHydrantBroken.cpp
```diff
@@ -0,0 +1,47 @@
...
+		auto* bouncer = EntityManager::Instance()->GetEntity(self->GetVar<LWOOBJID>(u"bouncer"));
+
+		if (bouncer != nullptr)
+		{
+			GameMessages::SendBouncerActiveStatus(bouncer->GetObjectID(), false, UNASSIGNED_SYSTEM_ADDRESS);
+
+			GameMessages::SendNotifyObject(bouncer->GetObjectID(), self->GetObjectID(), u"disableCollision", UNASSIGNED_SYSTEM_ADDRESS);
+		}
+
+		self->Kill();
```
Wincent01, line 41: `self->Kill()` will not show the smashing animation, did you intend to use `self->Smash()`? The script this is based on is old, so it might be worth investigating if it was correct.

dScripts/WhFans.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+	std::vector<Entity*> fanVolumes = EntityManager::Instance()->GetEntitiesInGroup(fanGroup);
+
+	auto* renderComponent = static_cast<RenderComponent*>(self->GetComponent(COMPONENT_TYPE_RENDER));
```
Wincent01, line 22: Use modern `self->GetComponent<RenderComponent*>()` instead, and check for nullptr.

Wincent01 (changes requested):
Some minor feedback about the patterns used. I did the scripts these are based on, they are old, not sure if they are correct in every way.

dScripts/WhFans.cpp
aronwk-aaron: Noted, still experimenting with the fans. Currently in a state where it causes the the world to crash on load and was trying to get it to not

dScripts/RockHydrantBroken.cpp
aronwk-aaron: I think Kill() is correct, since there is no model to smash at that point, just the entity that is making the bouncer, which is invisible

dScripts/WhFans.cpp
aronwk-aaron: Oh man that nullptr is what was crashing the world server, Fans now work :D

aronwk-aaron:
Fans and hydrant now work as intended
Need to check if the falling ice needs to be implemented serverside or not, as well as get a cdclient migration for the missing FB missions.

dScripts/RockHydrantBroken.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+void RockHydrantBroken::OnStartup(Entity* self)
+{
+	self->AddTimer("playEffect", 1);
+
+	const auto hydrant = "hydrant" + self->GetVar<std::string>(u"hydrant");
+
+	const auto bouncers = EntityManager::Instance()->GetEntitiesInGroup(hydrant);
+
+	Game::logger->Log("RockHydrantBroken", "Broken Rock Hydrant spawned (%s)\n", hydrant.c_str());
```
EmosewaMC, line 13: When finished testing please remove this log

dScripts/RockHydrantSmashable.cpp
```diff
@@ -1,24 +1,23 @@
...
 	info.spawnerID = self->GetSpawnerID();
-	info.spawnerNodeID = 0;
 
-	Entity* newEntity = EntityManager::Instance()->CreateEntity(info, nullptr);
-	if (newEntity) {
-		EntityManager::Instance()->ConstructEntity(newEntity);
-	}
-}
\ No newline at end of file
+	Game::logger->Log("RockHydrantBroken", "Rock Hydrant spawned (%s)\n", data->GetString().c_str());
```
EmosewaMC, line 18: Same here - extra logging not needed when done testing

dScripts/RockHydrantSmashable.cpp
```diff
@@ -1,24 +1,23 @@
...
-	NiPoint3 pos = physics->GetPosition();
+void RockHydrantSmashable::OnDie(Entity* self, Entity* killer)
+{
+	const auto hydrantName = self->GetVar<std::u16string>(u"hydrant");
 
-	EntityInfo info;
+	LDFBaseData* data = new LDFData<std::string>(u"hydrant", GeneralUtils::UTF16ToWTF8(hydrantName));
+
+	EntityInfo info {};
 	info.lot = 12293;
```
EmosewaMC, line 12: Would be a bit better to give this number a name so it's not just random.  Can define it in the header for this file as a private

dScripts/WhFans.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+#include "CppScripts.h"
+#include "GameMessages.h"
+#include "EntityManager.h"
+#include "PhantomPhysicsComponent.h"
+
+class WhFans : public CppScripts::Script
+{
+public:
+	void OnStartup(Entity* self);
```
EmosewaMC, line 10: All of these methods here need overrides after their declaration.  `OnStartup(Entity* self) override;`

EmosewaMC: Togglefx does not need an override

EmosewaMC (changes requested):
Some more small changes I notice while reviewing. Overall well done, just minor things

aronwk-aaron:
@EmosewaMC resolved all your comments and re-tested, also adjusted the scripts that these scripts are based on.

aronwk-aaron:
Going to leave this as just the scripts and will discuss the migration for cdclient and missing locale entries with others

EmosewaMC (approved):
lgtm

PR #516, opened 2022-04-23
Live accurate item loot sources

EmosewaMC:
This PR finishes the implementation of the SendAddItemToInventoryClientSync GameMessage and now sends the correct loot source when an item is earned.  This allows achievements to give their items as they did in live, 1 at a time when the achievement flips over.  Basic testing of moving items to the vault and picking up items showed no issues.  Have not seen any duplicate items anywhere but will be testing more in the coming days.
Some side issues fixed while working on this:
- All instances of a loot source are given the enum prefix eLootSourceType:: and some instances of inventory types were given the same treatment.
- Vendor sell inventory now shows the scroll bar when more items that would fit in the inventory are sold.  For some reason this still requires the player to refresh the vendor be exiting and re-opening the interface.


dGame/dComponents/InventoryComponent.cpp
```diff
@@ -282,8 +284,8 @@ void InventoryComponent::AddItem(
 
 			continue;
 		}
-
-		auto* item = new Item(lot, inventory, slot, size, {}, parent, showFlyingLoot, isModMoveAndEquip, subKey);
+		Game::logger->Log("InventoryComponent", "new item with source %i\n", lootSourceType);
```
EmosewaMC, line 287: remove log when done

EmosewaMC: removed

PR #522, opened 2022-04-26
Added vendor refreshing 

EmosewaMC:
Fixed a few issues in VendorComponent.
- Corrected serialization to only happen on construction.
- Added functionality to refresh the vendor based on info from the vendor component table
- some whitespaceing inconsistencies.
- Sorted includes.
Tested the vendor in Nimbus Station and when the time is up to refresh the inventory, the vendor inventory refreshes, as opposed to previously where the world would need to reset in order to refresh the inventory.


dGame/dComponents/VendorComponent.cpp
```diff
@@ -1,116 +1,135 @@
...
+	RefreshInventory(true);
 }
 
 VendorComponent::~VendorComponent() = default;
 
 void VendorComponent::Serialize(RakNet::BitStream* outBitStream, bool bIsInitialUpdate, unsigned int& flags) {
-    outBitStream->Write1();
-    outBitStream->Write1(); // this bit is REQUIRED for vendor + mission multiinteract
-    outBitStream->Write(HasCraftingStation());
+	// Only serialize this entity on construction.
```
EmosewaMC, line 16: This needs to be done every time we serialize, not just on construction.

Issue #526, opened 2022-04-29
BUG: "Something in the maelstrom" mission from Master Fong Shader does not play cinematic


MustacheMan2:
### Make sure you've done the following:

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### DarkflameServer Version

16b108f

### Issue Description

When turning in the "Something in the Maelstrom" mission for Master Fong Shader in the great tree, a cinematic doesn't play.

### Reproduction Steps

turn in the mission to Master Fong Shader

### Expected Behavior

https://youtu.be/mZk8w1Ru6FM?t=872

a cinematic identical to the one played when interacting with the maelstrom orb in NT should play

### Environment

ubuntu server

PR #531, opened 2022-04-30
Fong darkitect

aronwk-aaron:
Add the cutscene to play when interacting with Fong

As seen here:
https://www.youtube.com/watch?v=mZk8w1Ru6FM&t=872s

Resolves: #526

### DIFF
--- a/dScripts/Darkitect.cpp
+++ b/dScripts/Darkitect.cpp
@@ -1,0 +1,35 @@
+#include "Darkitect.h"
+#include "MissionComponent.h"
+#include "DestroyableComponent.h"
+#include "EntityManager.h"
+#include "GameMessages.h"
+#include "Character.h"
+
+void Darkitect::Reveal(Entity* self, Entity* player)
+{
+	const auto playerID = player->GetObjectID();
+
+	GameMessages::SendNotifyClientObject(self->GetObjectID(), u"reveal", 0, 0, playerID, "", player->GetSystemAddress());
+
+	self->AddCallbackTimer(20, [this, self, playerID]() {
+		auto* player = EntityManager::Instance()->GetEntity(playerID);
+
+		if (!player) return;
+
+		auto* destroyableComponent = player->GetComponent<DestroyableComponent>();
+		auto* missionComponent = player->GetComponent<MissionComponent>();
+		auto* character = player->GetCharacter();
+
+		if (destroyableComponent != nullptr && missionComponent != nullptr && character != nullptr) {
+			destroyableComponent->SetArmor(0);
+			destroyableComponent->SetHealth(1);
+			destroyableComponent->SetImagination(0);
+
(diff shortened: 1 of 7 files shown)


dScripts/Darkitect.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+#include "CppScripts.h"
+
+class Darkitect : public CppScripts::Script
```
EmosewaMC, line 4: This doesnt need to inherit the CppScripts::Script interface since it is a helper class.  This can safely be removed along with the include

dScripts/FvFong.cpp
```diff
@@ -0,0 +1,12 @@
+#include "FvFong.h"
+#include "Darkitect.h"
+#include "MissionComponent.h"
+
+void FvFong::OnMissionDialogueOK(Entity* self, Entity* target, int missionID, MissionState missionState)
+{
+    if (missionID == 734 && missionState == MissionState::MISSION_STATE_READY_TO_COMPLETE)
+    {
```
EmosewaMC, line 8: This curly brace should be trailing the if statement not on a new line

dScripts/Darkitect.cpp
```diff
@@ -0,0 +1,40 @@
...
+void Darkitect::Reveal(Entity* self, Entity* player)
+{
+    const auto playerID = player->GetObjectID();
+
+    GameMessages::SendNotifyClientObject(self->GetObjectID(), u"reveal", 0, 0, playerID, "", player->GetSystemAddress());
+
+    self->AddCallbackTimer(20, [this, self, playerID]() {
+        auto* player = EntityManager::Instance()->GetEntity(playerID);
+
+        if (player == nullptr)
```
EmosewaMC, line 17: This can be simplified a bit.  `if (!player) return;`

dScripts/Darkitect.cpp
```diff
@@ -0,0 +1,40 @@
...
+        {
+            return;
+        }
+
+        auto* destroyableComponent = player->GetComponent<DestroyableComponent>();
+        auto* missionComponent = player->GetComponent<MissionComponent>();
+        auto* character = player->GetCharacter();
+
+        if (destroyableComponent != nullptr && missionComponent != nullptr && character != nullptr)
```
EmosewaMC, line 26: Same trailing braces change in this file too

EmosewaMC (commented):
Well done overall.  Liked the splitting of the main work to a new file instead of copy and pasting code.  Few style comments but nothing blocking.

dScripts/Darkitect.h
aronwk-aaron: ah, yes! I made note of that and lost access to my server yesterday before I could make that change

EmosewaMC (approved):
lgtm gonna let this build on windows before merging

PR #552, opened 2022-05-09
replace the survey link with the repo link

aronwk-aaron:
(not that anything can load in the in-game browser anyway)


Wincent01:
There is a repo link specified in the config file, maybe link to that instead of a const string?

aronwk-aaron:
> There is a repo link specified in the config file, maybe link to that instead of a const string?

Can do, was considering that

aronwk-aaron:
Testing with the default and with an empty string and everything worked as expected

PR #554, opened 2022-05-10
WIP: Fully implement Property Behaviors

EmosewaMC:
This is an in progress PR and you may _lose your save data_ or _be unable to run your server_ should you use this branch on a production environment.  Use at your own risk.

Done:
The UI is currently fully functional and operates as closely as it does to its live counter part.  
Figure out how to allow behaviors to be dragged onto models without needing to be in the behaviors GUI
Implement server side physics simulation for rotation behaviors
Store the information in the database somehow
Load the behaviors from the database
Allow custom behaviors to be stored in the players inventory

To Implement:
Implement state changing on impact, on proximity leave and on proximity enter, follow player, flee from player and do damage nodes

Bugs:
- Strips should be usable again at the end of the strip and not only when a reset node is applied.  
- Picking up a model causes it to lose all behaviors.  This is because we destroy the entity of the model on pickup.  Maybe destroy it upon being put away instead?
- Move to Home needs to move towards home at a speed, not a teleport

There are of course bugs that will appear if you choose to test this branch out.  

This branch is getting split into smaller, more incremental PRs to make merging less of a headache.  

- [x] Deserializer merged
- [x] Model Component serialization merged
- [x] Fix Entity Timers
- [ ] PR for tests - in review
- [ ] PR for required migrations - in review
- [ ] TBD PR for implementing the saving, loading
- [ ] TBD PR for implementing the behavior system structure
- [ ] TBD PR for implementing each individual behavior 

node


EmosewaMC:
Enough of this has been implemented into the main branch that this WIP PR is no longer needed.  Any further updates will happen in smaller PRs.

PR #557, opened 2022-05-17
Lup launcher setting from lvl

aronwk-aaron:
Get setting from the lvl file instead of cdclient

### DIFF
--- a/dGame/dComponents/RocketLaunchLupComponent.cpp
+++ b/dGame/dComponents/RocketLaunchLupComponent.cpp
@@ -6,16 +6,13 @@
 
 RocketLaunchLupComponent::RocketLaunchLupComponent(Entity* parent) : Component(parent) {
 	m_Parent = parent;
-
-	// get the lup worlds from the cdclient
-	std::string query = "SELECT * FROM LUPZoneIDs;";
-	auto results = CDClientDatabase::ExecuteQuery(query);
-	while (!results.eof()) {
-		// fallback to 1600 incase there is an issue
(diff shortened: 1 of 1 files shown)


dGame/dComponents/RocketLaunchLupComponent.cpp
```diff
@@ -1,21 +1,17 @@
...
-
-	// get the lup worlds from the cdclient
-	std::string query = "SELECT * FROM LUPZoneIDs;";
-	auto results = CDClientDatabase::ExecuteQuery(query);
-	while (!results.eof()) {
-		// fallback to 1600 incase there is an issue
-		m_LUPWorlds.push_back(results.getIntField(0, 1600));
-		results.nextRow();
+	std::string zoneString = GeneralUtils::UTF16ToWTF8(m_Parent->GetVar<std::u16string>(u"MultiZoneIDs"));
+	std::stringstream ss(zoneString);
```
EmosewaMC, line 9: Believe this is unnecessary.  There should be a tool in general utils to split a string by a delimiter ans return a vector of the values.

EmosewaMC (changes requested):

dGame/dComponents/RocketLaunchLupComponent.cpp
aronwk-aaron: There are but none return the right type

EmosewaMC:
Resolved requested change in discord.  What tests have been run on this?

aronwk-aaron:
> Resolved requested change in discord. What tests have been run on this?

Tested going to all LUP worlds and they worked as expected

PR #559, opened 2022-05-18
climbable server side serializarion implementation

aronwk-aaron:
Implementation of how to check if something is climbable from the server side and set it as such
Feedback very welcome.

Tested with the different types of climbable objects and made sure it matched with client side side scripted objects
Tested non-climbable objects to make sure they aren't climbable


dGame/dComponents/SimplePhysicsComponent.h
```diff
@@ -14,16 +14,24 @@
 
 class Entity;
 
+enum eClimbableType : int32_t {
+	CLIMBABLE_TYPE_NOT,
```
EmosewaMC, line 18: This should specify zero as the starting number
`climbable_type_not = 0,
other values`

EmosewaMC (commented):
only have 1 non-blocking comment.  other than that this looks fine

dGame/dComponents/SimplePhysicsComponent.cpp
```diff
@@ -17,17 +17,28 @@ SimplePhysicsComponent::SimplePhysicsComponent(uint32_t componentID, Entity* par
     m_Position = m_Parent->GetDefaultPosition();
     m_Rotation = m_Parent->GetDefaultRotation();
     m_IsDirty = true;
+
+    std::u16string climbable_type = m_Parent->GetVar<std::u16string>(u"climbable");
```
Wincent01, line 21: const auto& claimableType = ...
Is more in line with our conversations.

dGame/dComponents/SimplePhysicsComponent.h
```diff
@@ -14,16 +14,24 @@
 
 class Entity;
 
+enum eClimbableType : int32_t {
```
Wincent01, line 17: Would prefer an enum class here. @EmosewaMC opinion? What has been the conversion recently?

EmosewaMC: I'd agree with @Wincent01 that this should be an enum class to protect against any possible bad comparisons.  
See [this](https://stackoverflow.com/questions/18335861/why-is-enum-class-preferred-over-plain-enum) stack overflow thread which explains why to prefer classes over plain enums.

aronwk-aaron: Neat, did not know that! Will change accordingly

EmosewaMC:
Changes have been addressed - merging

Issue #119, opened 2021-12-08
BUG: Slow Kraken in Shooting Gallery


MustacheMan2:
The Kraken is very slow and loses all of its difficulty in the process.

averysumner:
Can you show evidence of its speed in live? Jamesster play through link?

MustacheMan2:
[https://m.youtube.com/watch?v=xilmSHk8oq8&list=PL9EFB8500E625ADC7&index=39](url) 7:50

The Kraken is significantly slower than it should be in DLU

averysumner:
Interesting… possibly a hardcoded value somewhere. Any idea @Wincent01?

aronwk-aaron:
Hard coded in the script:
https://github.com/DarkflameUniverse/DarkflameServer/blob/58d2cee984cccdb268b0718ebe5407ec535510fa/dScripts/SGCannon.cpp#L988-L994

Last two entries on Line 992 are the min and max speed, will tinker with these till it looks live accurate

PR #562, opened 2022-05-24
Make Kraken speed live accurate

aronwk-aaron:
Resolves issue #119 
Made the speed match the simple physics speed, since the speed in the script was not correct.
Tested with all three krakens, and they move as fast as the did in live.

This may be indicative of a larger issue with SG movement speeds, but other things looked normal in comparison to live

### DIFF
--- a/dScripts/SGCannon.cpp
+++ b/dScripts/SGCannon.cpp
@@ -989,8 +989,8 @@ std::vector<std::vector<SGEnemy>> SGCannon::GetWaves() {
             {
                 std::vector<std::string> { "Wave_1_Ness_1", "Wave_1_Ness_2", "Wave_2_Ness_1" },
                 2565, 10.0, 15.0, true, 10.0, 15.0,
-                2.0, 10000, false, 0.0, 1.0,
-                1.0, true, true
+                7.0, 10000, false, 0.0, 7.0,
+                7.0, true, true
             },
 
             // Friendly 1


EmosewaMC:
This title does not provide any context as to what this fixes.  Please update the title.

EmosewaMC:
Where did you confirm these speeds as well?  Is there a packet capture you looked at or is this a guess?

aronwk-aaron:
> Where did you confirm these speeds as well? Is there a packet capture you looked at or is this a guess?

The new speed is from the [simple physics component ](https://explorer.lu-dev.net/objects/2565/3)
while the speed is defined in the SG_Cannon.lua script, [this speed does not match live recording](https://youtu.be/xilmSHk8oq8?t=569) where as the speed from the cdclient (7 instead of 2/1) does match live.

I looked at packet captures, but couldn't find anything that had it's speed.

EmosewaMC:
>  [this speed does not match live recording](https://youtu.be/xilmSHk8oq8?t=569) where as the speed from the cdclient (7 instead of 2/1) does match live.

This footage doesn't actually show the krakens moving for enough time to verify its speed.  According to packet captures the kraken is supposed to follow a moving platform waypoint and not have its speed set by the simple physics.  Is the speed being set via the simple physics in dlu at the moment?  
image
See the following image for the correct serialization of the kraken.  Is DLU currently following this method of serialization by doing it via the moving platform component or are we doing this via another method?

PR #567, opened 2022-05-27
Only make teams for minigames if we have more than one participant

### LINKED ISSUE #467: BUG: Playing singleplayer minigames puts you in a team

### Issue Description

When joining the shooting gallery, or a survival/battle instance solo, you will be put into a "team" which means the messages will pop up in chat and the team chat will open, remaining open after you leave the minigame.

### Reproduction steps

join any minigame alone

### Expected Behavior

you should not be put in a team if there are no teammates, such as shown here https://youtu.be/xilmSHk8oq8?t=360

### Environment

ubuntu server

aronwk-aaron:
Resolves #467 

Tested with mini-games where
 * you can enter as one player (no team made)
 * Enter with multiple players (with 2 or more a team was made)
 * multiplayer racing (with 2 or more a team was made)
 * single player racing (no team made)

Things behaved as expected and the minigames played as expected

### DIFF
--- a/dGame/dComponents/ScriptedActivityComponent.cpp
+++ b/dGame/dComponents/ScriptedActivityComponent.cpp
@@ -487,22 +487,24 @@ void ActivityInstance::StartZone() {
 	    return;
 
 	auto* leader = participants[0];
-
-	CBITSTREAM;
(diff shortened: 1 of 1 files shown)


EmosewaMC (approved):
Looks reasonable to me.

PR #574, opened 2022-06-06
Fix unhealthy docker containers

ff-sclark:
Added cURL to brickbuildfix and account-manager docker containers. Now docker should report these as healthy.


aronwk-aaron (commented):
So I understand why you added the curls, but there are no defined health checks?

ff-sclark:
The health checks are defined in the docker-compose.yml file. Currently only brickbuildfix and account-manager have health checks.

aronwk-aaron:
ah okay,somehow completely overlooked the dockercompose

Issue #123, opened 2021-12-08
Unable to use shift to build or talk to NPCs


mwolverine63:
When I first log into the game, I can use shift to talk to NPCs.

After a few minutes, shift will stop working. I have to log out and log back in to get shift to start working again, so I can interact with NPCs or build.

I am running an Ubuntu server on a macOS client

IAmVolvic:
This is the same as https://github.com/DarkflameUniverse/DarkflameServer/issues/122

averysumner:
> This is the same as #122

This sounds like a client issue unrelated to #122.

Pr3roxDLC:
Im experiencing the same problem with players "timing out" randomly, no disconnects, just the client getting no more packets from the server, checked it with WireShark on Windows for the client and tshark on my vps and both showed the server just stopping sending packets to the client while the client still tries to talk back to the client

For me the best way to fix this is by logging back out, and either just logging back in, or logging out, logging back in and when it gets stuck while loading closing the game, as this causes the server to recieve a "disconnect packet" resulting in the server actually removing the player and allowing them to log back it

Server Running on a Ubuntu 18.04 VPS using MariaDB, the source code being built for the Normal Client and the clients using lcrd's unpacked client

Pr3roxDLC:
@mwolverine63 does it also cause your game to stop dropping stuff when killing enemies/smashing stuff?

mwolverine63:
Yes, that happens too.

IAmJacket:
This is currently happening to me. Stuck trying to talk to Sky Lane. 

Pressing shift to interact with an NPC or to build will not do anything, and immediately afterwards broken objects will not drop coins or loot. Broken items will then reappear a second after breaking as if nothing happens. All UI components seem to stop working as well, attempting to change your Minifigure will not result in exiting the server, you'll still be playing.

Idkahn:
I am running into this issue too. From what I can tell through Wireshark, there are some packets sent from the server while it's frozen but only when I try to interact with or attack something, and they seemingly do nothing in game. Other than that, nothing comes in from the server.

There are also messages like `[StunBehavior]: Failed to find target (##################)!` that are logged to the server console when I attack something, where the number is what I assume is the id of the object or enemy.

averysumner:
We've been unable to reproduce any of this behavior. @Idkahn the
`StunBehavior` messages are unrelated.

averysumner:
Has this issue persisted?

[...]

PR #584, opened 2022-06-16
Fix an issue where quickbuilds were sometimes unbuildable

### LINKED ISSUE #122: Quickbuild sometimes just stops working. Even for other players
image

EmosewaMC:
Address an issue where quickbuilds would become unbuildable.  The main issue lied within serializing parent/child info too often for some reason / serializing it when the info wasnt dirty.  Only serializing this info when it is actually dirty and has changed has addressed the issue and allows quickbuilds to never break.

Since this PR changes how Parents and Children work the following was tested.

First I tested the 100% consistent reproduction in Avant Gardens (finish a build, spawn a build, finish the spawned build, original build respawns, serialize spawned build) and the build never became unbuildable.  The engineers turret that is spawned with the helmet with the 50% consistent reproduction had never encountered the issue where it was unbuildable.  Racing still works as well as spawning a car from your backpack.  Pets still spawn and set their owner correctly and taming them still builds the item as intended.

Fixes #122

### DIFF
--- a/dGame/Entity.cpp
+++ b/dGame/Entity.cpp
@@ -562,19 +562,6 @@ void Entity::Initialize()
 			comp->SetPostImaginationCost(rebCompData[0].post_imagination_cost);
 			comp->SetTimeBeforeSmash(rebCompData[0].time_before_smash);
 
-			const auto rebuildActivatorValue = GetVarAsString(u"rebuild_activators");
-
-			if (!rebuildActivatorValue.empty()) {
-				std::vector<std::string> split = GeneralUtils::SplitString(rebuildActivatorValue, 0x1f);
-				NiPoint3 pos;
-
-				pos.x = std::stof(split[0]);
-				pos.y = std::stof(split[1]);
-				pos.z = std::stof(split[2]);
-
-				comp->SetActivatorPosition(pos);
-			}
-
 			const auto rebuildResetTime = GetVar<float>(u"rebuild_reset_time");
 
 			if (rebuildResetTime != 0.0f) {
@@ -977,8 +964,8 @@ void Entity::WriteBaseReplicaData(RakNet::BitStream* outBitStream, eReplicaPacke
 	}
 	
 	// Only serialize parent / child info should the info be dirty (changed) or if this is the construction of the entity.
-	outBitStream->Write((m_ParentEntity != nullptr || m_ChildEntities.size() > 0) && (m_IsParentChildDirty || packetType == PACKET_TYPE_CONSTRUCTION));
-	if ((m_ParentEntity != nullptr || m_ChildEntities.size() > 0) && (m_IsParentChildDirty || packetType == PACKET_TYPE_CONSTRUCTION)) {
+	outBitStream->Write(m_IsParentChildDirty || packetType == PACKET_TYPE_CONSTRUCTION);
+	if (m_IsParentChildDirty || packetType == PACKET_TYPE_CONSTRUCTION) {
 		m_IsParentChildDirty = false;
 		outBitStream->Write(m_ParentEntity != nullptr);
 		if (m_ParentEntity) {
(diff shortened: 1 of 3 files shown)


EmosewaMC:
I am seeking a few reviewers to verify these changes and ensure that I have not done something critically wrong or that this breaks something crucial.

dGame/Entity.cpp
```diff
@@ -974,8 +975,11 @@ void Entity::WriteBaseReplicaData(RakNet::BitStream* outBitStream, eReplicaPacke
 		}
 		else outBitStream->Write0(); //No GM Level
 	}
-	outBitStream->Write((m_ParentEntity != nullptr || m_ChildEntities.size() > 0));
-	if (m_ParentEntity || m_ChildEntities.size() > 0) {
+	
+	// Only serialize parent / child info should the info be dirty (changed) or if this is the construction of the entity.
+	outBitStream->Write((m_ParentEntity != nullptr || m_ChildEntities.size() > 0) && (m_IsParentChildDirty || packetType == PACKET_TYPE_CONSTRUCTION));
```
Wincent01, line 980: Maybe eval this long Boolean expression to a variable and use that, for readability.

Wincent01 (approved):
Great find!

dGame/dComponents/RebuildComponent.cpp
```diff
@@ -22,6 +22,18 @@ RebuildComponent::RebuildComponent(Entity* entity) : Component(entity) {
...
 		m_Precondition = new PreconditionExpression(GeneralUtils::UTF16ToWTF8(checkPreconditions));
 	}
+
+	// Should a setting that has the build activator position exist, fetch that setting here and parse it for position.
+	// It is assumed that the user who sets this setting uses the correct character delimiter (character 31)
+	auto positionAsVector = GeneralUtils::SplitString(m_Parent->GetVarAsString(u"rebuild_activators"), 31);
+	if (positionAsVector.size() == 3) {
+		m_ActivatorPosition.x = std::stof(positionAsVector[0]);
+		m_ActivatorPosition.y = std::stof(positionAsVector[1]);
+		m_ActivatorPosition.z = std::stof(positionAsVector[2]);
```
Xiphoseer, line 32: Can you add a try/catch around these? bad config values shouldn't cause exceptions

dGame/Entity.cpp
EmosewaMC: It is the exact same boolean no?  Simplifying will be applied in the morning though to make this easier to read :)

dGame/dComponents/RebuildComponent.cpp
EmosewaMC: Yeah sure.  Will toss those in so they default to using the quickbuilds spawn point should the position be bad

dGame/dComponents/RebuildComponent.cpp
```diff
@@ -22,6 +22,18 @@ RebuildComponent::RebuildComponent(Entity* entity) : Component(entity) {
 	{
 		m_Precondition = new PreconditionExpression(GeneralUtils::UTF16ToWTF8(checkPreconditions));
 	}
+
+	// Should a setting that has the build activator position exist, fetch that setting here and parse it for position.
+	// It is assumed that the user who sets this setting uses the correct character delimiter (character 31)
+	auto positionAsVector = GeneralUtils::SplitString(m_Parent->GetVarAsString(u"rebuild_activators"), 31);
```
Xiphoseer, line 28: not that it matters, i think 0x1F would feel less magic here than 31

dGame/dComponents/RebuildComponent.cpp
Wincent01: Use GeneralUtils::TryParse

Jettford:
I am good to approve this once the already mentioned config parsing is resolved but other then that LGTM.

aronwk-aaron (commented):
Agree with the changes noted by others. Will approve once those have been addressed

DarwinAnim8or (approved):

aronwk-aaron:
Wont this fix cover #123 as well, since that's the result of this serialization being incorrect?

EmosewaMC:
The addressed feedback has been implemented and all tests have passed once again.  I am glad we are finally able to have solved this issue

PR #587, opened 2022-06-18
Address pets not consuming imagination on summon or while being active

### LINKED ISSUE #538: BUG: Pets do not cost Imagination to summon

### DarkflameServer Version

427c4a8c33fdb8eca06abd49477e70816f00530b

### Issue Description

In live, summoning pets cost a fixed 1 imagination. This behavior is not reflected in DLU.

### Reproduction Steps

* Double click on a pet in your model bag to summon.
* Note that it does not subtract 1 imagination when summoning.

### Expected Behavior

LEGO Universe (Part 40) @ 7:16
https://youtu.be/H2IYH2y-NXs?t=436

1 imagination should be subtracted. Presumably there was also a denying message if the player has 0 imagination?

### Environment

Windows client, Linux server

EmosewaMC:
No one wanted this but it is a bug that needed to be fixed unfortuantely.  By default this value is disabled on DLU but can be enabled with a config setting.

Tested equipping a pet, taming a pet, running out of imagination, reaching zero while having a pet out, spawning a pet with zero imagination, despawning a pet with zero imagination, letting a pet despawn from zero imagination and that despawning the pet doesn't consume imagination over time after it dies.

Fixes #538


aronwk-aaron (approved):
LGTM

PyEddy:
If Darkflame's goal is to be as close to live accurate as possible then I would suggest making this setting on by default and those who want to change that can go into worldconfig.ini and change that. I think people who are setting up DLU servers (while people eventually move onto modding anyway) for the first time would like to have an experience as close to live as possible.

That's just my thoughts on this tho

EmosewaMC:
> If Darkflame's goal is to be as close to live accurate as possible then I would suggest making this setting on by default and those who want to change that can go into worldconfig.ini and change that. I think people who are setting up DLU servers (while people eventually move onto modding anyway) for the first time would like to have an experience as close to live as possible.
> 
> That's just my thoughts on this tho

We have DLU vanity NPCs on by default which arent in live.  This will be fine.

PR #588, opened 2022-06-18
Game Message (GM) system refactor

Jettford:
The goals of this pull request are to improve the poor job at implementing an originally object based message system used to communicate most game data.

The current goals are:
- Improve game message events/handlers
- Change game messages into structs that magically (de-)serialize
- Reduce bloat in bloated files such as GameMessages.cpp (and accompanying header)


dGame/dComponents/Component.h
```diff
@@ -43,10 +44,17 @@ class Component
      */
     virtual void LoadFromXml(tinyxml2::XMLDocument* doc);
 
+    using Handler = void(Component::*)(GameMessage*);
```
codeshaunted, line 47: We use std::function in other places for callbacks, it'd probably be wise to use it here too.

codeshaunted:
This is definitely a good idea and I've been wanting to do it for ages before DLU even went FOSS. A few things though:

1. Header file per game message will likely get messy very quickly... in theory it would compile faster though because of less include bloat. We should have a discussion about whether or not to do this.
2. It would be much cleaner to have game message classes auto-generated at compile time from a definitions file. This can somewhat easily be done via an extra C++ target that is built and called from CMake (could even do this with a Python script, but Python + CMake can get messy).
3. The way you've designed the game message classes isn't very elegant and in my opinion needs some improvement. The following is some code demonstrating how to do this in a much cleaner way with delegating constructors:
```cpp
class GameMessage {
public:
	GAME_MSG GetID() { return m_ID; }
	Entity* GetEntity() { return m_Associate; }
protected:
	GameMessage(GAME_MSG ID, Entity* associate) : m_ID(ID), m_Associate(associate) {}
	GAME_MSG m_ID;
	Entity* m_Associate;
};

class EquipItem : public GameMessage {
public:
	EquipItem(Entity* associate, bool immediate = true, LWOOBJID objectID = LWOOBJID_EMPTY) : GameMessage(GAME_MSG_EQUIP_ITEM, associate), m_Immediate(immediate), m_ObjectID(objectID) {}
	// ideally these wouldn't be public
	bool m_Immediate;
	LWOOBJID m_ObjectID;
};
```
4. Maybe we should think about using kaitai? I know @lcdr had been working on a kaitai version of lu_packets.

dGame/dComponents/Component.h
Jettford: `std::function` seems just like a less memory efficient way of doing the same thing.

codeshaunted: > `std::function` seems just like a less memory efficient way of doing the same thing.

The same could be said about game message objects.

EmosewaMC:
Would it be possible to do this in smaller increments with one GM at a time to keep reviews and the pr as a whole reasonable?  Doing all of this in one PR will be incredibly cumbersome and could introduce issues with how many changes there would be at once

Jettford:
Umm no? There are hundreds of game messages.

codeshaunted:
> Umm no? There are hundreds of game messages.

It would be ideal to introduce the system slowly, with a hybrid approach at first to avoid a HUGE merge conflict at the end of this.

Jettford:
Yeah it would be that is why we had a whole conversation about that and came to that conclusion.

codeshaunted:
> Yeah it would be that is why we had a whole conversation about that and came to that conclusion.

Just making sure we’re all on the same page.

PR #591, opened 2022-06-19
Implement Precompiled Headers to speed up compile time

EmosewaMC:
This commit adds 5 Precompiled Headers to dramatically speed up compile time  This PR sped up compile time from 20 minutes system time to 5 minutes system time.

### DIFF
--- a/dGame/dBehaviors/BehaviorSlot.h
+++ b/dGame/dBehaviors/BehaviorSlot.h
@@ -1,5 +1,8 @@
 ﻿#pragma once
 
+#ifndef BEHAVIORSLOT_H
+#define BEHAVIORSLOT_H
+
 enum class BehaviorSlot
 {
 	Invalid = -1,
@@ -9,3 +12,5 @@ enum class BehaviorSlot
 	Head,
 	Consumable
 };
+
+#endif
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,7 +96,7 @@ if (__include_backtrace__ AND __compile_backtrace__)
 
     link_directories(${backtrace_SOURCE_DIR}/.libs/)
     include_directories(${backtrace_SOURCE_DIR})
-endif(__include_backtrace__)
+endif()
 
 endif()
 endif(UNIX)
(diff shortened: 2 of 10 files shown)


dGame/dBehaviors/BehaviorSlot.h
```diff
@@ -1,5 +1,8 @@
 ﻿#pragma once
 
+#ifndef BEHAVIORSLOT_H
```
codeshaunted, line 3: What's with the traditional header guards in addition to `#pragma once`?

EmosewaMC: For some reason the pragma once is not preventing the file from being included twice in a compile, but using traditional header guards prevented the issue
image

Do you know of a way I can use the PCH while still using pragma once?

codeshaunted: I mean `#pragma once` is non-standard (but it's implemented by *most* compilers). I've advocated in the past for traditional guards to be used for this very reason. I don't know of any way to fix this, what compiler are you using?

EmosewaMC: For now ill be only using the Entity.h PCH since thats the only one that doesnt require us to add the traditional include guards.  Will look into a way to make sure this compiles on gcc with other PCH in the future.  Even for now, having that one file as a PCH has reduced compile times drastically

EmosewaMC: There may be other third party headers I can pre compile I that ill look into though

EmosewaMC:
This is ready for review as is.  Please review commit comments for details on the cmake configuration I went with to setup the precompiled headers.  It is necessary as far as I can tell to list out the files individually due to some issues I was having with compilation.

aronwk-aaron:
tested on my dev laptop and here are the results: (with make -j8)

Old time:
```
real    5m29.272s
user    27m24.778s
sys     1m38.665s
```

New time:
```
real    2m29.638s
user    13m25.271s
sys     0m54.045s

```

Great improvements!

PR #623, opened 2022-07-09
CDClient rework

Wincent01:
Utilizes `boost::interprocess` to share maps between worlds.
The master server loads the tables into memory, and the worlds hook into it when they want to get an entry.

This solves the issue of skill delay and saves memory, but introduces the third-party library Boost.
There should be a conversation about the introduction of Boost — it is a large library and adds extra steps to the installation. Other options should be considered.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -331,6 +331,15 @@ target_link_libraries(dNet dDatabase)
 target_link_libraries(dGame dDatabase)
 target_link_libraries(dChatFilter dDatabase)
 
+# Include Boost
+find_package(Boost COMPONENTS interprocess)
+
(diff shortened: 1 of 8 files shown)


Jettford:
I don't personally believe that adding a library as large as boost is worth it for saving 18mb per world server. Other options should be investigated first in my opinion.

codeshaunted:
> I don't personally believe that adding a library as large as boost is worth it for saving 18mb per world server. Other options should be investigated first in my opinion.

I don't want to add boost either but this is a pretty significant memory save... Are there any other similar cross platform IPC libraries? You could also just use pipes but that could get a bit sketchy.

PR #642, opened 2022-07-13
Update CMake configuration for easier maintenance

Jettford:
This pull request improves our CMake build system to avoid issues such as everything being inside of a single CMakeLists file and use of things such as glob to gather file names.


dCommon/ZCompression.h
```diff
@@ -2,6 +2,8 @@
 
 #include <cstdint>
 
+#ifndef _WIN32
```
codeshaunted, line 5: These platform macros are sketchy at best and I think we should move away from using them. Maybe alongside these CMake changes we could introduce platform detection within CMake that defines our own macros (e.g. `DARKFLAME_PLATFORM_WINDOWS`, `DARKFLAME_PLATFORM_LINUX`, etc). CMake can much more reliably detect the platform where combinations of these macros will often fail due to compiler parity issues.

dChatFilter/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+set(DCHATFILTER_SOURCES "dChatFilter.cpp")
+
+add_library(dChatFilter ${DCHATFILTER_SOURCES})
```
codeshaunted, line 3: We should probably explicitly specify `STATIC` or `SHARED` here if we don't want this to be influenced by the value `BUILD_SHARED_LIBS`. Traditionally DLU has been built with almost entirely static linking, but we could change this as there are several shared dependencies between the server executables.

dCommon/ZCompression.h
aronwk-aaron: If these weren't reliable, then I think every piece of c++ software would be having issues, no reason to fix what ain't broken

codeshaunted: > If these weren't reliable, then I think every piece of c++ software would be having issues, no reason to fix what ain't broken

I recall having issues with clang not working on the initial DLU release due to these macros being used. We also tend to use these platform macros as an analog for compiler (Windows == MSVC, Linux == GCC, etc). It really isn't that hard to implement detection in CMake and it looks nicer in code as well. Here's an example: https://github.com/codeshaunted/soul/blob/0286b126ff89d2e6ba5b285b631e6b1ece28d40f/CMakeLists.txt#L27

dChatFilter/CMakeLists.txt
EmosewaMC: Agreed.  Let's not leave anything up to randomness and this is a good chance to reduce the binary size a touch for the libs used across all the binaries like raknet, dCommon, dNet etc.

dCommon/ZCompression.h
Jettford: Someone else can do this later if wanted, but I won't be.

dChatFilter/CMakeLists.txt
Jettford: Yeah that is a good idea. Will do.

Jettford:
Yeah I will revert this to a draft
Too broken at the moment

Jettford:
I can't... seem to revert it to a draft?

Jettford:
Thank you, could not find that.

EmosewaMC:
> WorldServer and ChatServer do not start and seg fault when started. I don't know whats wrong at the moment but this cannot be merged as is.

The issue looks to be something with regex assignment of a regex string.  Investigating...

luxaritas:
Looks like the segfault issue was that the compiler flags were being set after the add_subdirectory calls rather than before, so the compilation was being done with the wrong flags. Moving that seems to have fixed it

CMakeLists.txt
```diff
@@ -177,15 +177,38 @@ file(
...
+# Compiler flags:
+# Disabled deprecated warnings as the MySQL includes have deprecated code in them. 
+# Disabled misleading indentation as DL_LinkedList from RakNet has a weird indent.
+# Disabled no-register
+# Disabled unknown pragmas because Linux doesn't understand Windows pragmas.
+if(UNIX)
+	if(APPLE)
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++17 -O2 -Wuninitialized -D_GLIBCXX_USE_CXX11_ABI=0 -D_GLIBCXX_USE_CXX17_ABI=0 -fPIC")
+	else()
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++17 -O2 -Wuninitialized -D_GLIBCXX_USE_CXX11_ABI=0 -D_GLIBCXX_USE_CXX17_ABI=0 -static-libgcc -fPIC")
```
codeshaunted, line 189: `-std=gnu++17` is redundant as we already set the standard to C++17 at the beginning of this file 
https://github.com/DarkflameUniverse/DarkflameServer/blob/17b8e7045f9a0f297e53a721ac3953e71143d9f7/CMakeLists.txt#L5
We should also do 
```cmake
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
in order to enforce the C++17 requirement.

Jettford: There are differences between `gnu++17` and `std++17`, there is a requirement for it

codeshaunted: > There are differences between `gnu++17` and `std++17`, there is a requirement for it

The [CMake docs](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html) seem to imply that it should be doing `gnu++`, but alright.

Jettford: The standard is not required however.

Jettford:
@EmosewaMC @aronwk-aaron Could you please test and review, as far as I have tested. It works on Windows and Ubuntu 20

CMakeLists.txt
codeshaunted: > The standard is not required however.

I recall C++17 being required for something related to Mersenne Twister... If not, I know it at least C++11 is required, so maybe we should change the general standard requirement to reflect that.

Jettford: I don't think it really matters.

EmosewaMC:
Will test on aarch and mac today

aronwk-aaron (approved):
LGTM, tested on on ubuntu 18.04

EmosewaMC (approved):

PR #645, opened 2022-07-16
Added running mysql migrations to build script

HailStorm32:
Added the awesome QoL option Jett added to MasterServer to the build script.

The intention here is that it will reduce the amount of users who report crashes after pulling an update because they forgot to run the migrations. With the migration command added to the build script, users will now only need to pull and build (what they always did) and not need to worry about running migrations.


EmosewaMC:
This cant be used as is.  There is a reason this has to be done later in the setup.  You will need to:
A) Change MasterServer to not check for dependencies before running the command
B) Find some way to check that a database actually exists first before blindly running commands.
If you read the README you will observe that you run the ./MasterServer -m command after setting up all the client files.  Please make these changes.

HailStorm32:
Some good points. I'll look into it

HailStorm32:
@EmosewaMC I have now added checks before running the migration command. It will only run if the sqlite database exists (aka the client files have been moved over) and if it can make a connection to the mysql database

EmosewaMC:
If we are going to be having this build script do this work MasterServer needs to be changed to allow the migration runner to run without checking whether a cdclient exists or not.  We should also be printing out an error should there be no connection or database and not just exiting the program without printing the error.

EmosewaMC:
On second thought we shouldnt be doing this in a shell script.  A database may not be on the localhost and so the command will not do anything locally but will need to be run remotely.  All of this should be handled in MasterServer only and not in a shell script

HailStorm32:
My thought process for it being in the build script was so that it would only run after a build, and not every time the server is run. That being said, looking at the migration runner code, it looks like that code keeps track of the migrations its run and wont rerun them. However, it still seems wasteful to run the command every time.

I can add a check to the script to see if `mysql_host` is anything but `localhost` and run the same mysql command, but connect it to the remote host. 

As for checking if mysql is installed, that isn't necessary. The mysql command I run returns no errors, and so any errors that do occur will trigger the migration command not to run. I read from stderr, so any error (even command not found errors) trips the system so to speak. This has been tested this on a system without mysql installed, and it runs as expected.

EmosewaMC:
I wouldn't add a check for localhost - there shouldn't be any special cases here since we know the address and can connect to it via the MariaDbConnector so checking for localhost is wasteful.  Running the server with the -m command is not wasteful because if the server is built again, you cannot guarantee there are no new migrations under a build.  Doing these checks with a shell script skirts around these checks and makes it so that those who run the command manually still get the same error.  It would be more efficient to just let Master do all this scripting work than to have a shell script that does effectively what Master does but on a dirtier level.

HailStorm32:
will do, I'll look into adding this to the MasterServer

build.sh
```diff
@@ -6,4 +6,28 @@ cd build
...
 
 # Run make to build the project. To build utilizing multiple cores, append `-j` and the amount of cores to utilize, for example `make -j8`
-make
\ No newline at end of file
+make
+
+# Run mysql migrations
+
+#Check we have sqlite db
+if [ -f res/CDServer.sqlite ]; then
```
EmosewaMC, line 14: Is this shell script in the build folder?  I don't see how this would be checking the right directory if you are in the root of the DarkflameServer codebase.

HailStorm32: This tripped me up as well. The script is in the root dir of the repo, but the script changes its dir to build, so all paths reflect that. 

Had the paths from the root dir, but after some errors, realized the script was written to change its directory

HailStorm32:
As requested I moved the migration running to MasterServer. I didn't include any checks, as MasterServer already runs those checks.

Jettford:
These changes that you have made to MasterServer are known to cause issues. Do not submit pull requests that you have not tested.

HailStorm32:
> These changes that you have made to MasterServer are known to cause issues. Do not submit pull requests that you have not tested.


This compiles and runs fine. What issues are you referring to?

Jettford:
This does not run fine.

aronwk-aaron:
This does not work they way you are understanding it, this will cause crashes

PR #646, opened 2022-07-16
Address timers being iterated through poorly

EmosewaMC:
Currently, timers attached to the Entity are iterated through based on begin -> end, and if a timer is done, we delete the timer at that position from the vector.  The problem with this is if you are deleting a timer, you need to not step forward one space.  If you do so, you will miss updating some timers, and also update timers this frame that may not be intended to update this frame.  This PR intends to correct this error so timers fire at the correct times and at an expected, consistent rate.


EmosewaMC:
This needs to be iterated on more since m_Timers is edited outside the scope through other timers.

aronwk-aaron (approved):

Jettford (approved):

PR #649, opened 2022-07-17
Parent and Child Deletion Improvements

EmosewaMC:
Improve our child deletion process by deleting children if they happen to be null and also add support to delete parents from children.

Appears to fix semi random crashes on all platforms caused by the method.


aronwk-aaron (approved):
LGTM

PR #652, opened 2022-07-17
Blacklist and chat changes

Jettford:
- Implement best friend chat (Use of blacklist)
- Fix bad word highlighting


CMakeLists.txt
```diff
@@ -84,13 +84,14 @@ make_directory(${CMAKE_BINARY_DIR}/locale)
 make_directory(${CMAKE_BINARY_DIR}/logs)
 
 # Copy ini files on first build
```
EmosewaMC, line 86: Comment needs to be updated since it moves more than just ini file now

dChatFilter/dChatFilter.h
```diff
@@ -20,17 +20,17 @@ class dChatFilter
...
+	void ReadWordlistPlaintext(const std::string& filepath, bool whiteList);
+	bool ReadWordlistDCF(const std::string& filepath, bool whiteList);
+	void ExportWordlistToDCF(const std::string& filepath, bool whiteList);
+	std::vector<std::string> IsSentenceOkay(const std::string& message, int gmLevel, bool whiteList = true);
 
 private:
 	bool m_DontGenerateDCF;
-	std::vector<size_t> m_Words;
+	std::vector<size_t> m_NoNoWords;
+	std::vector<size_t> m_YesYesWords;
```
EmosewaMC, line 31: I get the funny nature of the lists but this will just cause confusion for people looking at these variables in the future.  I recommend changing them to be m_ApprovedWords and m_RejectedWords

dNet/ClientPackets.cpp
```diff
@@ -284,17 +287,77 @@ void ClientPackets::HandleChatModerationRequest(const SystemAddress& sysAddr, Pa
...
 	}
 
+	bool isBestFriend = false;
+
+	if (chatLevel == 1) {
+		// Private chat
+		LWOOBJID idOfReceiver = LWOOBJID_EMPTY;
+
+		{
+			sql::PreparedStatement* stmt = Database::CreatePreppedStmt("SELECT name FROM charinfo WHERE name = ?");
```
EmosewaMC, line 310: I recommend changing this to be a unique_ptr so when you leave scope this memory is automatically managed.  Right now its a memory leak since it is not deleted.

dNet/ClientPackets.cpp
```diff
@@ -284,17 +287,77 @@ void ClientPackets::HandleChatModerationRequest(const SystemAddress& sysAddr, Pa
...
+			stmt->setString(1, receiver);
+
+			sql::ResultSet* res = stmt->executeQuery();
```
EmosewaMC, line 313: Same thing here - unique_ptr would manage this memory for you.

dNet/ClientPackets.cpp
```diff
@@ -284,17 +287,77 @@ void ClientPackets::HandleChatModerationRequest(const SystemAddress& sysAddr, Pa
...
+
 	if (!bAllClean) {
-		unacceptedItems.insert(std::make_pair((char)0, (char)message.length()));
+		for (const auto& item : segments) {
+			if (item == "") {
+				unacceptedItems.insert({ (char)0, (char)message.length()});
+				break;
+			}
+
+			unacceptedItems.insert({ message.find(item), item.length() });
```
EmosewaMC, line 359: How does this work for multiple of the same word being rejected in the same sentence?  From my observations this would only ever highlight the first word and never highlight the later appearances of them.  

It also seems inefficient to use .find here when you parsed the string above already and know what word is rejected and where it is.  I would recommend changing how this gets the index of the first letter of the word such that it is done in isSentenceOkay above when we split the word up the first time.

In my mind something like the word number mapped to its length that way there isnt a need to .find() the word again and also supports rejecting multiple of the same word in a sentence.  Then instead of splitting the word up by space in isSentenceOkay, we split it up here and pass that to the method so we can use it as a reference here for what words to accept or reject.

dNet/WorldPackets.cpp
```diff
@@ -192,20 +192,23 @@ void WorldPackets::SendChatModerationResponse(const SystemAddress& sysAddr, bool
...
-		bitStream.Write(it->second);
+	for (auto it : unacceptedItems) {
+		bitStream.Write<uint8_t>(it.first); // start index
+		bitStream.Write<uint8_t>(it.second); // length
 	}
 
+    for (int i = unacceptedItems.size(); 64 > i; i++) {
+        bitStream.Write<uint16_t>(0);
+    }
+
```
EmosewaMC, line 211: Formatting here is all over the place.  Would be great to clean this up.

dNet/ClientPackets.cpp
```diff
@@ -284,17 +287,77 @@ void ClientPackets::HandleChatModerationRequest(const SystemAddress& sysAddr, Pa
...
+
+			sql::ResultSet* res = stmt->executeQuery();
+
+			if (res->next()) {
+				idOfReceiver = res->getInt("id");
+			}
+		}
+
+		if (user->GetIsBestFriendMap().find(receiver) == user->GetIsBestFriendMap().end() && idOfReceiver != LWOOBJID_EMPTY) {
+			sql::PreparedStatement* stmt = Database::CreatePreppedStmt("SELECT * FROM friends WHERE (player_id = ? AND friend_id = ?) OR (player_id = ? AND friend_id = ?) LIMIT 1;");
```
EmosewaMC, line 321: Same thing here - unique_ptr for the query and result set to manage the memory for you.

dNet/ClientPackets.cpp
```diff
@@ -284,17 +287,77 @@ void ClientPackets::HandleChatModerationRequest(const SystemAddress& sysAddr, Pa
 		receiver.push_back(static_cast<uint8_t>(character));
 	}
 
+	if (!receiver.empty()) {
+		if (std::string(receiver.c_str(), 4) == "[GM]") {
+			receiver = std::string(receiver.c_str() + 4, receiver.size() - 4);
+		}
+	}
+
```
EmosewaMC, line 295: This is a bit confusing to read at first glance - I would recommend using .substr here instead to clarify what specifically is going on here instead of this method

dChatFilter/dChatFilter.cpp
```diff
@@ -86,46 +95,56 @@ bool dChatFilter::ReadWordlistDCF(const std::string& filepath) {
...
 	}
 }
 
-bool dChatFilter::IsSentenceOkay(const std::string& message, int gmLevel) {
-	if (gmLevel > GAME_MASTER_LEVEL_FORUM_MODERATOR) return true; //If anything but a forum mod, return true.
-	if (message.empty()) return true;
+std::vector<std::string> dChatFilter::IsSentenceOkay(const std::string& message, int gmLevel, bool whiteList) {
+	if (gmLevel > GAME_MASTER_LEVEL_FORUM_MODERATOR) return { }; //If anything but a forum mod, return true.
+	if (message.empty()) return { };
+	if (!whiteList && m_NoNoWords.empty()) return { "" };
```
EmosewaMC, line 116: Why is this one an empty string vs the other two above?

dChatFilter/dChatFilter.cpp
```diff
@@ -134,8 +153,4 @@ size_t dChatFilter::CalculateHash(const std::string& word) {
 	size_t value = hash(word);
 
 	return value;
-}
-
-bool dChatFilter::IsInWordlist(size_t word) {
-	return std::find(m_Words.begin(), m_Words.end(), word) != m_Words.end();
-}
+}
```
EmosewaMC, line 156: Having new lines at end of files is ideal

dChatFilter/dChatFilter.h
```diff
@@ -20,17 +20,17 @@ class dChatFilter
...
 	std::vector<size_t> m_UserUnapprovedWordCache;
 
 	//Private functions:
 	size_t CalculateHash(const std::string& word);
-	bool IsInWordlist(size_t word);
-};
+};
```
EmosewaMC, line 36: Same here - new line at EOF

EmosewaMC (changes requested):

dChatFilter/dChatFilter.cpp
Jettford: Why?

dChatFilter/dChatFilter.cpp
Jettford: Because that is disallowing chat if someone removes the whitelist

dNet/ClientPackets.cpp
Jettford: I will add comments

dNet/ClientPackets.cpp
Jettford: > How does this work for multiple of the same word being rejected in the same sentence? From my observations this would only ever highlight the first word and never highlight the later appearances of them.

This runs for every every flagged segment by the functions in the filter?

Jettford: > It also seems inefficient to use .find here when you parsed the string above already and know what word is rejected and where it is. I would recommend changing how this gets the index of the first letter of the word such that it is done in isSentenceOkay above when we split the word up the first time.

We don't already know where it is as far as I know?

dChatFilter/dChatFilter.cpp
EmosewaMC: Because it is standard for how some files read to EOF
c.f. https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline#:~:text=If%20content%20is%20added%20to,you%20actually%20wanted%20to%20see.

dChatFilter/dChatFilter.h
Jettford: Done

dChatFilter/dChatFilter.h
Jettford: Done

dChatFilter/dChatFilter.cpp
Jettford: Done

dNet/ClientPackets.cpp
EmosewaMC: > > How does this work for multiple of the same word being rejected in the same sentence? From my observations this would only ever highlight the first word and never highlight the later appearances of them.
> 
> This runs for every every flagged segment by the functions in the filter?

.find returns the first occurrence of the word found and cares not about ones that are found already

dNet/ClientPackets.cpp
Jettford: I don't really like using unique pointers, I have added the deletes back.

dNet/ClientPackets.cpp
Jettford: Similar to previous comment

dNet/ClientPackets.cpp
Jettford: Similar to previous comment

dNet/ClientPackets.cpp
Jettford: > .find returns the first occurrence of the word found and cares not about ones that are found already

Ah I see what you mean, I thought about that but thought wasn't worth implementing properly. If people can't gather a word is not ok in one place but ok in another then that is there problem. Will look into fixing the logic.,

CMakeLists.txt
Jettford: Changed the comment

aronwk-aaron (approved):
LGTM

EmosewaMC (approved):

PR #655, opened 2022-07-19
Fix Model Component Serialization

EmosewaMC:
Address numerous issues with the Model Component to get it to a more stable state:
- Correct Model Component enum value to its intended value.
- Add the Destroyable Component to Entities that have the Model Component and don't already have a Destroyable Component
- Move Item Component initialization down further so we can setup the Model Component first and, if we have a Model Component, do not assign an Item Component since we handle that serialization in the Model Component
- Sync Entity settings for Entities that contain a Model Component.  Do not sync settings otherwise since this causes issues
- Serialize a Destroyable Component after the Render Component should the Entity have a Model Component and not have a Destroyable Component serialized yet.
- Allow SetFaction to set the faction to -1, should the ignoreChecks flag be set
- Comment and address Model Component serialization as well as make it correct with what live captures show
- Formatting (rip 4 spaces)
- Add the live accurate settings to models that are not Brick by Brick models.  This is crucial because of the Component Whitelist flag we send over the wire since this gives the client a specific Component overload to use as opposed to the classic one.
- Remove some commented out code around what I was working on

Comparing a diff of logs from the current DLU serialization vs this PRs, the client logs no longer show `Unable to de-serialize` errors for models and Brick by Brick models.  Models can also now have behaviors dragged onto them but no handling of this behavior is added.


dGame/dComponents/ModelComponent.h
```diff
@@ -12,51 +12,50 @@ class Entity;
...
 
-    /**
-     * The ID of the user that made the model
-     */
-    LWOOBJID m_userModelID;
+	/**
+	 * The ID of the user that made the model
+	 */
+	LWOOBJID m_userModelID;
 };
```
aronwk-aaron, line 61: No newline at EOF

aronwk-aaron (changes requested):

dGame/dComponents/ModelComponent.h
EmosewaMC: done

aronwk-aaron (approved):

PR #663, opened 2022-07-20
General AMF cleanup

EmosewaMC:
Proper memory management as well as style cleanup


Xiphoseer:
Any specific reason this deletes the endian-specific encoding of integers? RakNet/BitStream probably indeed takes care of this if configured correctly, but that's something that should be mentioned somewhere in the PR or in code IMHO.

EmosewaMC:
> Any specific reason this deletes the endian-specific encoding of integers? RakNet/BitStream probably indeed takes care of this if configured correctly, but that's something that should be mentioned somewhere in the PR or in code IMHO.

Yeah raknet or something else is already writing this in the correct endian for whatever reason.  In my testing we had the u64 backwards which caused issues. Hence my assumption that the others are backwards to.  Moving to writing the values as is resolves the issues

PR #664, opened 2022-07-21
Refactor UpdateEntities to not lose items if we add them while processing

aronwk-aaron:
if we are deleting entities, and we add an entity to delete, don't throw it out, instead delete it for real

### DIFF
--- a/dGame/EntityManager.cpp
+++ b/dGame/EntityManager.cpp
@@ -165,8 +165,7 @@ void EntityManager::UpdateEntities(const float deltaTime) {
 	    e.second->Update(deltaTime);
 	}
 
-	for (const auto entityId : m_EntitiesToSerialize)
-	{
+	for (const auto entityId : m_EntitiesToSerialize) {
 		auto* entity = GetEntity(entityId);
 
 		if (entity == nullptr) continue;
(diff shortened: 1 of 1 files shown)


dGame/EntityManager.cpp
```diff
@@ -161,111 +161,82 @@ void EntityManager::DestroyEntity(Entity* entity) {
 }
 
 void EntityManager::UpdateEntities(const float deltaTime) {
-	for (const auto& e : m_Entities) {
-	    e.second->Update(deltaTime);
+	for (auto entry = m_Entities.begin(); entry != m_Entities.end(); entry++) {
```
Wincent01, line 164: Not sure this form of iteration is any more efficient than `for (var i : list)`. Regardless: `const auto& entry` instead of just `auto entry` avoids a copy here.

dGame/EntityManager.cpp
```diff
@@ -161,111 +161,82 @@ void EntityManager::DestroyEntity(Entity* entity) {
...
+		entry->second->Update(deltaTime);
 	}
 
-	for (const auto entityId : m_EntitiesToSerialize)
-	{
-		auto* entity = GetEntity(entityId);
+	for (auto entry = m_EntitiesToSerialize.begin(); entry != m_EntitiesToSerialize.end(); entry++) {
```
Wincent01, line 168: `const auto& entry`. Given that you are de-referencing the entry below, I can't see this being more efficient than the short-hand form.

dGame/EntityManager.cpp
```diff
@@ -161,111 +161,82 @@ void EntityManager::DestroyEntity(Entity* entity) {
...
 			Game::server->Send(&stream, UNASSIGNED_SYSTEM_ADDRESS, true);
 		}
 	}
-
 	m_EntitiesToSerialize.clear();
 
-	for (const auto& entry : m_EntitiesToKill)
-	{
-		auto* entity = GetEntity(entry);
+	for (auto entry = m_EntitiesToKill.begin(); entry != m_EntitiesToKill.end(); entry++) {
```
Wincent01, line 194: `const auto& entry`, see above.

dGame/EntityManager.cpp
```diff
@@ -161,111 +161,82 @@ void EntityManager::DestroyEntity(Entity* entity) {
...
 		}
 	}
-
 	m_EntitiesToKill.clear();
 
-	for (const auto entry : m_EntitiesToDelete)
-	{
-		// Get all this info first before we delete the player.
-		auto entityToDelete = GetEntity(entry);
+	for (auto entry = m_EntitiesToDelete.begin(); entry != m_EntitiesToDelete.end(); entry++) {
```
Wincent01, line 207: `const auto& entry`, see above.

Wincent01 (changes requested):
I'm confused as to what this PR wants. Is it just a refactor? Please elaborate.

aronwk-aaron:
> I'm confused as to what this PR wants. Is it just a refactor? Please elaborate.

with `for (var i : list)` the end is defined at the start, and if things get added, it will not process them since they are beyond the pre-defined end.

This is a change for mounts, so that when logging out, the possessor destructor will call the possessable destructor, which is on another entity. Before, the possessable would not get deleted since it wold be added during this loop, after the pre-defined end of the list and then the list gets cleared, which means we have a stray entity that never gets cleaned up.

dGame/EntityManager.cpp
aronwk-aaron: can't be `const auto&` since we are modifying the value of the iterator to traverse the list

dGame/EntityManager.cpp
aronwk-aaron: having to de-reference it below due to entry being an iterator, which this is just how iterators work, as far as I know

aronwk-aaron:
https://github.com/DarkflameUniverse/DarkflameServer/pull/589/files#diff-0a0e96564b43df0789e17458e858f9377106251b3f75cb8f0da5f32eaaf61d2bR25-R28

this was my solution to working around the issue before, which does work; but I thought it would be better to fix the root issue

dGame/EntityManager.cpp
EmosewaMC: It can be passed by reference here though with &

dGame/EntityManager.cpp
EmosewaMC: https://m.cplusplus.com/reference/vector/vector/push_back/

It's not about efficiency.  It's about adding to vectors when iterating through them as range based for loops.  When this is done, the end iterator is invalidated and since range based for loops are short hand for:
for(auto b = begin(), auto e = end(); b != e; ++b)
{
    auto value = *b;
}
There are cases where all iterators can be invalidated so it may be worth making this just a simple for loop in any case.

dGame/EntityManager.cpp
EmosewaMC: Further discussion shows we cant actually pass by reference here at the least

Wincent01 (approved):
Looks good to me

EmosewaMC (approved):

PR #671, opened 2022-07-24
Split out Level progression component

aronwk-aaron:
from Character Component
This is to get to the Player forced movement Comp in a sane way

Tested leveling up
test saving a loading the level
tested that level up rewards are given
test preconditions
tested mission rewards at and under level cap


dGame/dComponents/CharacterComponent.cpp
```diff
@@ -80,8 +80,6 @@ CharacterComponent::~CharacterComponent() {
 }
 
 void CharacterComponent::Serialize(RakNet::BitStream* outBitStream, bool bIsInitialUpdate, unsigned int& flags) {
-	outBitStream->Write1();
-	outBitStream->Write(m_Level);
 	outBitStream->Write0();
```
aronwk-aaron, line 83: This is the Player Forced Movement, which we always write0 to.

PR #675, opened 2022-07-24
Make logger automatically put a newline

aronwk-aaron:
at the end of the line
remove all the newlines in log calls

made sure logs to console and files are still correct


aronwk-aaron:
I accidentally made a bunch of whitespace changes since there SO MUCH trailing whitespace for no good reason.
But given that this is really a non-funcitonal change, I don't really care to go back and remove that, better just to get it gone, imo

EmosewaMC (approved):

PR #677, opened 2022-07-25
Add .editorconfig and do whitespace cleanup

aronwk-aaron:


codeshaunted:
An editor config would be good. Would also be cool if we could also a clang-format and clang-tidy pass at some point.

aronwk-aaron:
Yeah, will look into clang stuff after this. But editorconfig will be a godsend do that there's some sense of consistence.
I have a tool that'll run it across every relevant file in the repo, and there will be a tiny bit of cleanup with enums that are aligned (components in dCommon)

Also this is the guide for c++ editorconfig if anyone wants to see the options https://docs.microsoft.com/en-us/visualstudio/ide/cpp-editorconfig-properties?view=vs-2019

PR #679, opened 2022-07-26
Fix integer underflow with Bricks Collected world summary.

### LINKED ISSUE #678: BUG: Zone statistic for Bricks Collected underflows due to mismatched Integer types

### DarkflameServer Version

All

### Issue Description

When selling bricks, we receive the data as an int32 but we _store_ the data as a uint64.  This can cause an underflow when the received value is negative since you can sell bricks to lower your Bricks Collected count.  

### Reproduction Steps

1) Sell bricks
2) Observe that if you sold more bricks than you earned in a world, the value underflows and becomes uint64_MAX internally but is int32_MAX when the client sees it.

### Expected Behavior

Collected Bricks Bottoms out at zero?  This behavior should ideally be confirmed with live if possible.

### Environment

All

PyEddy:
This fixes an issue where BricksCollected goes to an insane number after selling more bricks than you collected in the area

Fixes #678 

On the latest commit it would do this
image

As you can see the Bricks Collected number is going way too high and shouldn't be like that

This fix changes that and after testing the going back and fourth between worlds, logging out and logging back in then going back and fourth again, restarting the server then logging in then going back and fourth again, it will stay on the number it was initially from when you sold the bricks

For example I sold around 19 bricks in LEGO Club (I didn't collect any bricks in LEGO Club so it will go down to -19), in the latest commit if I launch to another world it will first say -19 however when I go back to LEGO Club and then launch to another world again, it will go to that insane number you see in the image above. With the fix, if I go and sell 19 bricks in LEGO Club, it will go down to -19 as expected and if I launch to another world and come back then launch off to a new world again, it will stay on -19 rather than show some insane number. As you can see below
image


EmosewaMC:
> This fixes an issue where BricksCollected goes to an insane number after selling more bricks than you collected in the area
> 
> Refer to issue #678

Please be more specific with the title, and description of the fix.  

Please also describe the tests you have done to prove this works.

PyEddy:
> > This fixes an issue where BricksCollected goes to an insane number after selling more bricks than you collected in the area
> > Refer to issue #678
> 
> Please be more specific with the title, and description of the fix.
> 
> Please also describe the tests you have done to prove this works. Also instead of refer to #678 say `Fixes ...` this will link the two together and close the issue when it is merged

Hopefully I made it look a lot better

PR #680, opened 2022-07-26
WSL Readme Section Changes

PyEddy:
As the title says, this will be changing some stuff in the readme section for WSL

Changes include 

- Added the Windows 11 RTM build 22000 to the WSL requirement for Windows 11 `WSL requires...`
- Added "Change WSL version" which is optional if the user wants to use WSL1 instead
- Added a note that lets users reading the readme know about the network issues that may occur when using WSL and that its not an issue to do with DLU

I think this PR is a good idea especially knowing some people use WSL and report network issues sometimes that DLU can't solve as its to do with Microsoft

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ cmake --build . --config Release
 ### Windows builds (WSL)
 This section will go through how to install [WSL](https://docs.microsoft.com/en-us/windows/wsl/install) and building in a Linux environment under Windows. WSL requires Windows 10 version 2004 and higher (Build 19041 and higher) or Windows 11 (Build 22000 and higher). 
 
-Please note that Microsoft sometimes ends up breaking WSL (there tends to be network issues when using WSL) and DLU won't be able to fix them so verify that its not WSL that is broken before reporting an issue.
+Please note that Microsoft sometimes ends up breaking WSL2 (there tends to be network issues when using WSL2) and DLU won't be able to fix them so verify that its not WSL that is broken before reporting an issue. Its also recommended that users use WSL1 instead as it's more stable. 
 
 **Open the Command Prompt application with Administrator permissions and run the following:**
 ```bash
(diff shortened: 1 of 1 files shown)


Jettford:
> Please note that Microsoft sometimes ends up breaking WSL (there tends to be network issues when using WSL) and DLU won't be able to fix them so verify that its not WSL that is broken before reporting an issue.

This is very non-descript and unhelpful to the average user. Please note the actual issue to do with WSL 2 and recommend use of WSL 1.

aronwk-aaron:
Honestly, if you're on windows, do native windows compiling. WSL is really just a dev tool to begin with, and if you're a dev you'll know how to save to not use it. It should not be used for long running play servers

PyEddy:
> > Please note that Microsoft sometimes ends up breaking WSL (there tends to be network issues when using WSL) and DLU won't be able to fix them so verify that its not WSL that is broken before reporting an issue.
> 
> This is very non-descript and unhelpful to the average user. Please note the actual issue to do with WSL 2 and recommend use of WSL 1.

Right will change the note to specify that WSL2 is having issues and add a recommendation for WSL1

EmosewaMC:
The only change out of that list that I think is worth noting is the recommending that users use WSL1 over WSL2 for stability.  The other changes I don't believe will provide much more clarity than is already there

PyEddy:
> The only change out of that list that I think is worth noting is the recommending that users use WSL1 over WSL2 for stability.  The other changes I don't believe will provide much more clarity than is already there

I mean if we are recommending WSL1 we will have to show them how to change WSL versions as I have done

I think everything is great imo but I can change things if you want

aronwk-aaron:
No reason that we should be telling users to downgrade to wsl1, wsl2 works fine, the issues with it are out of are control. as well as MSFT has stabilized wsl2 over the previous months.
Additionally, a windows setup guide has been made by a community member, and people should be using native windows instead of WSL (be it 1 or 2)

PR #682, opened 2022-07-27
Clean up format logs

codeshaunted:
Cleans up some older log calls that use string concatenation instead of format strings for security.


EmosewaMC (approved):
Lgtm

PR #683, opened 2022-07-28
Fix racing lap times

### LINKED ISSUE #463: BUG: First lap for races does not track time correctly

### Issue Description

When the first lap of a race is completed, the time that is stored on the server for that lap is incorrect
Here is the relevant code:
https://github.com/DarkflameUniverse/DarkflameServer/blob/ad52facef71e1349057512237ccbd0b157af19f9/dGame/dComponents/RacingControlComponent.cpp#L813

### Reproduction steps

1) Enter a race
2) Complete the first lap within a time trial (or have access to the server logs)
3) Observe that the time output to console does not match the time for the lap.

### Expected Behavior

Lap time matches (to the second) what the client had.

### Environment

WSL

EmosewaMC:
Fixes #463 

Just a small change to properly track lap times on the first and second lap for races

Tested with logs and the correct time was reported for the first, second and third lap.

### DIFF
--- a/dGame/dComponents/RacingControlComponent.cpp
+++ b/dGame/dComponents/RacingControlComponent.cpp
@@ -818,7 +818,7 @@ void RacingControlComponent::Update(float deltaTime) {
 
             // Reached the start point, lapped
             if (respawnIndex == 0) {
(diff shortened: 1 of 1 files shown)


aronwk-aaron (approved):
Off by one error lol
LGTM, watched you test it

PR #686, opened 2022-07-29
Fix missions re-ordering on reload

### LINKED ISSUE #360: BUG: Mission Tracker mission order changes after relog

### Issue Description

It looks like after a relog the order of missions on the mission tracker (the bubble in the top center of the UI) seems to change.

### Reproduction steps

- Start some missions
- Look at the order
- Relog

### Expected Behavior

The order should stay

### Environment

Ask Jett (Oog)

EmosewaMC:
Fixes #360 

This PR implements the missing functionality of the `o` attribute for missions saved in the charxml.  The `o` value indicates the order the player accepted the mission in.  This number in live packet captures started at 746 and always went up by 1 for each _mission_ accepted.  

Tested that a fresh character gets the correct `o` value in their charxml as well as an existing character setting the correct defaults since we cannot know the order these were accepted in, they will simply remain in the current order with newer missions appending to the end of the list as expected.  Tested that accepting a daily mission for the second time and the mission was correctly assigned a new, unique order ID corresponding to its newer spot in the list.


Wincent01:
Could that arbitrarily number maybe be the number of achievements? Just a guess -- lgtm regardless.

EmosewaMC:
> Could that arbitrarily number maybe be the number of achievements? Just a guess -- lgtm regardless.

image
Precisely correct.  Good catch to just check that real quick.  I'll make this a cached query and make this the starting value

PR #688, opened 2022-07-31
Remove parentheses from ADD reporter_id

codeshaunted:
Removes parentheses from the `ADD reporter_id` statement in [migrations/dlu/2_reporter_id.sql](https://github.com/DarkflameUniverse/DarkflameServer/compare/reporter_id_fix?expand=1#diff-cf744faf89bba96f349dfaded7e41503511067b9fb4d1bcd7ef5d5337ad8abc7). This statement will error with parentheses on stock MariaDB.


EmosewaMC (approved):

Xiphoseer:
<https://mariadb.com/kb/en/alter-table/> and <https://mariadb.com/kb/en/create-table/#column-definitions> suggests the issue is not that it has parentheses, but rather that they only enclose the name and not the type and default too. But removing them works either way, so not a blocker.

PR #689, opened 2022-07-31
Add Security Policy

Xiphoseer:
This PR adds a `SECURITY.md`, which GitHub picks up in its UI as the Security Policy for the repo. As mentioned in the proposed text, I created both `darkflame-security` as well as `darkflame-security-announce` for use as described.

If you are a project member or private server operator, please contact me to join either one of these groups.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -48,4 +48,4 @@ if you fail to comply with this embargo, you might be exluded from [receiving se
 Unfortunately we cannot provide bug bounties at this time.
 
 [darkflame-security]: mailto:darkflame-security@googlegroups.com
-[darkflame-security-announce]: mailto:darkflame-security-announce@googlegroups.com
+[darkflame-security-announce]: https://groups.google.com/g/darkflame-security-announce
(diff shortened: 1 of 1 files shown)


PyEddy:
This actually sounds like a good idea and I was wondering about what we should do if someone found an exploit/security vulnerablity in DLU, it would be dangerous if a discovered vulnerablity is disclosed to the public as some may use it for malicious intent

I'm glad that covers that by saying and I'm paraphrasing within 90 days of discovery of the vulnerablity, it cannot be disclosed to the public (assuming til its patched?)

SECURITY.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Embargo
+
+We propose a 90 day (approx. 3 months) embargo on security vulnerabilities. That is, we ask everyone not to disclose the vulnerabilty
+publicly until either:
+
+1. 90 days have passed from the time the first related email is sent to `darkflame-security@`
+2. a security advisory related to the vulnerability has been published by the project.
+
+if you fail to comply with this embargo, you might be exluded from [receiving security updates](#receiving-security-updates).
```
DarwinAnim8or, line 44: "if" should have a capital I here

DarwinAnim8or (approved):

PR #690, opened 2022-07-31
Add an update bash script for updating the Darkflame Server [no-ci]

PyEddy:
This script is intended for updating the Darkflame Server in Linux, it does all of what build.sh does but this has more commands on it as its starting from the first command you run when updating your Darkflame Server which is `git pull` then 
`git submodule update --init --recursive`

I feel like I should try to justify my reasoning for this, build.sh can be seen as a first use when building the server as if you are building the server for the first time you aren't going to be git pulling rather git cloning, update.sh can be seen as updating the server (assuming you have already your server set up and all it needs is to update to the latest changes) there after.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -1,0 +1,17 @@
+# Git Pull the recent changes from the repository
+git pull
+
+# check if there is a submodule update
+git submodule update --init --recursive
+
+# cd into the directory build
+cd build
+
+# run the cmake command
+cmake ..
+
+# run the make command
+make
+
+# run the migrations
+./MasterServer -m


Jettford:
Not sure I like the idea of just adding more scripts to solve "problems", maybe a single batch script with args could be an option if people **really** wanted something to run 1 command. (The first 2 commands in your added script can be one: `git pull --recurse-submodules`)

EmosewaMC:
More scripts are not the solution imo.  This also doesn't follow the different building solution for Mac.  

There should only be one script if there is one since they all do the same thing.

PyEddy:
After multiple discussions both here and on discord, it seems its best to keep it how it is and there is no need for an update.sh

Therefore I am closing the PR

Thanks for helping me reach a conclusion on this PR

PR #694, opened 2022-08-01
Replace the vsprintf used in the logger with vsnprintf.

Jettford:
This removes the requirement for a strlen implementation at the top of the function.


aronwk-aaron (approved):
LGTM

PR #695, opened 2022-08-01
Use UTF8ToUTF16 in more places

Xiphoseer:
As suggested in #658, this replaces `ASCIIToUTF16` with the new `UTF8ToUTF16` in a few places where it seems appropriate. I haven't tested this ingame yet, so would appreciate if someone does, incase we missed a spot where the input encoding is actually Latin-1 / windows-1521

### DIFF
--- a/dGame/dGameMessages/GameMessages.cpp
+++ b/dGame/dGameMessages/GameMessages.cpp
@@ -978,6 +978,8 @@ void GameMessages::SendSetNetworkScriptVar(Entity* entity, const SystemAddress& 
 		bitStream.Write(entity->GetObjectID());
 	bitStream.Write((uint16_t)GAME_MSG_SET_NETWORK_SCRIPT_VAR);
 
+	// FIXME: this is a bad place to need to do a conversion because we have no clue whether data is utf8 or plain ascii
+	// an this has performance implications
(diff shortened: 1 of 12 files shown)


dGame/dGameMessages/GameMessages.cpp
```diff
@@ -5086,7 +5088,8 @@ void GameMessages::HandleNotifyServerLevelProcessingComplete(RakNet::BitStream*
 	wss << "name=0:";
 	wss << character->GetName();
 
-	std::u16string attrs = GeneralUtils::ASCIIToUTF16(wss.str());
+	// FIXME: only really need utf8 conversion for the name, so move that up?
```
EmosewaMC, line 5091: Technically speaking this whole stringstream should be re-written to LDF instead of a hard coded stream like this. Would be more descriptive of the actual fix.

Jettford (approved):
Looks good!

dGame/dGameMessages/GameMessages.cpp
Xiphoseer: Good point!

aronwk-aaron: Just to note, there are several places that need to be updated to use proper LDF instead of hardcoded streams

aronwk-aaron (approved):
Looks great!

PR #696, opened 2022-08-01
Hide dependency compilation warnings

Xiphoseer:


EmosewaMC (approved):
This passes CI so no friction on my end.

PR #699, opened 2022-08-02
Repair chat filter file format for different architecture pointer sizes.

Jettford:


DarwinAnim8or:
Er-- why replace size_t with uint64_t? size_t makes it so that 32-bit systems would use uint32 and 64-bit uses 64-bit ints
I don't see why you would force it to uint64_t

Jettford:
The blacklist cannot be pushed to the main repo so the dcf had to be and that has to be supported across systems.

PR #708, opened 2022-08-04
Implement missing null checks on destruction of the navigation mesh

Jettford:
Add a few checks before calling free on some Detour objects


Jettford:
This is based off of the incorrect branch, recreating...

PR #709, opened 2022-08-04
Implement missing null checks on destruction of the navigation mesh

Jettford:
Add a few checks before calling free on some Detour objects


EmosewaMC (approved):
Didn't crags on Ubuntu local anymore

aronwk-aaron (approved):

PR #711, opened 2022-08-05
Fix checks for if the physics world is loaded

Jettford:
This PR fixes the check for if the physics world is loaded and therefore repairs the navigation of ships in the Gnarled Forest shooting gallery.


EmosewaMC (approved):
lgtm

aronwk-aaron (approved):

PR #724, opened 2022-08-09
Brickbuildfix port can now be changed through .env

Denperidge:


Wincent01:
The client is hardcoded to do a get request to port 80, so this wouldn't be so useful.

PR #725, opened 2022-08-12
Address Brick-By-Brick builds not properly saving and make migrations automatic

### LINKED ISSUE #190: Brick model processing does not work most often
### 

### Issue Description

After spending a lot of time brick building lately I can confirm in its current state it is unusable. There is a fair chance processing a brick model, whether small or moderately sized, will fail and the model will disappear (and yes I group models for efficiency). Editing a model that was processed earlier also has a fair chance of deleting the whole thing (although the models will remain in the client). This is too inconsistent to find decent workarounds. I found a method while making several models to form a house and process each one individually, which works and with less casualties, but after coming back later one model had faded away, and coming back after that the rest of the models had gone. Another user had also been unable to load in (see #156). I'd say either fix this or remove brick building altogether to avoid this sort of frustration.

### Reproduction steps

go to brick building mode, create some models, process them, and if successful, try to edit them. You will quickly see what I mean

### Expected Behavior

Model processing should have a very small chance of failing if at all, and editing a model should not cause it to be deleted.

### Environment

ubuntu server

### LINKED ISSUE #654: BUG: Brick-By-Brick saves of sufficient size do not save completely

### DarkflameServer Version

ec4ed8fa7e37d1a9457a17580273e3db79e4e818

### Issue Description

If a BBB is large enough, not all of it's data will be saved.
Some data is saved, but it get's cut off in the middle of the file.
The data can decompress successfully, so it's not like it only saved half of the compressed part

### Reproduction Steps

make a single model that takes a property score to over 40
save
save is corrupted and no BBB can load on that property now since one is corrupted

### Expected Behavior

Save get's saved completely

### Environment

Ubuntu 20.04, have seen it happen across multiple servers

EmosewaMC:
Fixes #654 
Fixes #190 

This PR fixes large Brick-By-Brick (hereafter referred to as B3) models not saving due to B3 models being sent in a compressed SD0 format and not just a Zlib compressed format.  Since we only ever use the compressed version when communicating with the client, there is no reason to de-compress concatenate the sd0 chunks, and re-compress them.  Hence this PR makes it so we instead just store the full sd0 the client sends so we can send it to other clients as they request them.  In the future we will need to de-compress the sd0 in order to normalize the position of the model, but this fixes the core functionality of B3 where you couldn't make big models at all.

The second change necessary here is a regression fix for all previous models.  Since we are removing the SD0 header writing, it is necessary to include a command line argument in order to fix the old models, since this update will only ever need to be run once, but we need to de-compress the current Zlib stored in the database to see if its actually valid first as before this PR, large models would not save all their chunks.  It is hence necessary to parse _all_ of the models, check if they are valid lxfml, and if so just prepend the SD0 header and chunk size.  If it is not valid lxfml the model will be deleted from the database since models that saved broken lxfml would cause B3 to break entirely on the client and none would render at all.  

Tested building big and small models alike as a Civilian player.  All models saved and were able to be loaded properly when a player visited the property, including models that edged the maximum lxfml size allowed by Civilians.

Tested commands as follows:
Since all migrations are now automatic, I started the server and no matter what state the server was in, all B3 builds and all database columns and rows were updated to reflect what they should be in the live version

A database with both large and small models truncates only broken ones.  Large ones are processed correctly and not truncated.
Valgrind reports no memory leaks with the new functionality.


dGame/dGameMessages/GameMessages.cpp
```diff
@@ -2468,6 +2431,8 @@ void GameMessages::HandleBBBSaveRequest(RakNet::BitStream* inStream, Entity* ent
 
 		Note, in the live client it'll still display the bricks going out as they're being used, but on relog/world change,
 		they reappear as we didn't take them.
+
+		TODO Apparently the bricks are supposed to be taken via MoveInventoryBatch?
```
aronwk-aaron, line 2435: not necessarily move inv batch, but there's a BBB brick inventory. and the batch move is used a lot for when bricks are clone

EmosewaMC: So this is only used for cloning?

aronwk-aaron: possibly? but not sure. It's definitely used for cloning and moving the brick that were cloned into the bricksInBBB inventory, but whatever bricks are in the inventory is what should be removed

move batch could be used for other things, just not sure off the top of my head

aronwk-aaron: Can be a separate PR with a config, but I just wanted to clear up the explanation of how it works since you left a comment about it and I didn't feel like I explained it before that well

CMakeVariables.txt
```diff
@@ -15,6 +15,6 @@ __dynamic=1
 # __include_backtrace__=1
 # Set __include_backtrace__ to 1 to includes the backtrace library for better crashlogs.
 # __compile_backtrace__=1
-# Set __compile_backtrace__ to 1 to compile the backtrace library instead of using system libraries. 
-__maria_db_connector_compile_jobs__=1
+# Set __compile_backtrace__ to 1 to compile the backtrace library instead of using system libraries.
+__maria_db_connector_compile_jobs__=
```
EmosewaMC, line 19: need to add the 1 back here.

dCommon/BrickByBrickFix.cpp
```diff
@@ -0,0 +1,192 @@
...
+#include <memory>
+#include <iostream>
+#include <sstream>
+
+#include "tinyxml2.h"
+
+#include "Database.h"
+#include "Game.h"
+#include "ZCompression.h"
+#include "dLogger.h"
```
EmosewaMC, line 12: Simplify logging and make it less verbose.  Check which headers are unneeded.

dCommon/BrickByBrickFix.cpp
```diff
@@ -0,0 +1,192 @@
...
+			std::unique_ptr<tinyxml2::XMLDocument> document = std::make_unique<tinyxml2::XMLDocument>();
+			if (!document) {
+				Game::logger->Log("BrickByBrickFix", "Failed to initialize tinyxml document.  Aborting.");
+				return 0;
+			}
+
+			if (document->Parse(completeUncompressedModel.c_str(), completeUncompressedModel.size()) == tinyxml2::XML_SUCCESS) {
+				Game::logger->Log("BrickByBrickFix", "Model %llu is a valid brick-by-brick model!", modelId);
+			} else {
+				// Change this to just look for </LXFML> near the end of the function.  Should rely on human interference for this...
```
EmosewaMC, line 80: meant to say "shouldn't rely on human interference".  This should be automated.

dDatabase/Database.cpp
```diff
@@ -83,3 +83,13 @@ sql::PreparedStatement* Database::CreatePreppedStmt(const std::string& query) {
 void Database::Commit() {
 	Database::con->commit();
 }
+
+bool Database::GetAutoCommit() {
+	// TODO This should not just access a pointer.  A future PR should update this
```
EmosewaMC, line 88: Check for null and throw if used before database is setup?

dGame/dGameMessages/GameMessages.cpp
```diff
@@ -2514,6 +2478,7 @@ void GameMessages::HandleBBBSaveRequest(RakNet::BitStream* inStream, Entity* ent
 				auto result = query.execQuery();
 
 				if (result.eof() || result.fieldIsNull(0)) {
+					free(sd0Data);
```
EmosewaMC, line 2481: replace the free and malloc with the corresponding c++ keywords (new delete)

dCommon/BrickByBrickFix.cpp
```diff
@@ -0,0 +1,192 @@
...
+	input[0] = 's';
+	input[1] = 'd';
+	input[2] = '0';
+	input[3] = 0x01;
+	input[4] = 0xFF;
+	*reinterpret_cast<uint32_t*>(input + 5) = chunkSize; // Write the integer to the character array
+}
+
+bool CheckSd0Magic(sql::Blob* streamToCheck) {
+	return streamToCheck->get() == 's' && streamToCheck->get() == 'd' && streamToCheck->get() == '0' && streamToCheck->get() == 0x01 && streamToCheck->get() == 0xFF;
```
EmosewaMC, line 191: This should not be 2 separate methods I dont think anymore.  If the model is not sd0 just update it to sd0 so there is only 1 command to run.  Perhaps this is wrong but this just seems like an unneeded step.

EmosewaMC (commented):
Leaving comments on what needs to be done to finish this PR at a minimum.  I currently have other stuff on my plate.

EmosewaMC:
mariadb having cert issues...  Not sure how to fix this.

EmosewaMC:
Should we have a way to prevent users from starting the server should they have not updated their database of brick models yet?  Or should i just get rid of the command altogether and just always run the updating code on master server start?

aronwk-aaron:
either is not a great choice, the worse one being running it every time, imo
how about checking the migrations table for an entry and then don't run in that entry exists and do if it doesn't, then make the entry to mark it as run.
since this is "technically" a migration of database data, just not one that can easily be done in sql
and could run every time, or just make the migrate command call this function when it is run

aronwk-aaron (dismissed):

thirdparty/CMakeMariaDBLists.txt
```diff
@@ -17,10 +17,10 @@ if(WIN32 AND NOT MARIADB_BUILD_SOURCE)
...
 	endif()
 
-	if(NOT EXISTS "${MARIADB_MSI_DIR}/mariadb-connector-cpp-1.0.1-win64.msi" )
+	if(NOT EXISTS "${MARIADB_MSI_DIR}/mariadb-connector-cpp-1.0.2-win64.msi" )
 		message("Downloading mariadb connector/c++")
-		file(DOWNLOAD https://dlm.mariadb.com/1683453/connectors/cpp/connector-cpp-1.0.1/mariadb-connector-cpp-1.0.1-win64.msi
-			"${MARIADB_MSI_DIR}/mariadb-connector-cpp-1.0.1-win64.msi"
+		file(DOWNLOAD https://dlm.mariadb.com/2531525/Connectors/cpp/connector-cpp-1.0.2/mariadb-connector-cpp-1.0.2-win64.msi
+			"${MARIADB_MSI_DIR}/mariadb-connector-cpp-1.0.2-win64.msi"
 			EXPECTED_HASH MD5=548e743fbf067d21d42b81d958bf4ed7)
```
Xiphoseer, line 24: Changing the version without changing the hash seems to have broken CI?

EmosewaMC: Good catch but that's not what's up.  It's an issue with mariadb error ing out on the download.  I need to change the download link back but we won't have windows ci until they fix that.

EmosewaMC: For reference, the hash that is returned here is just a null character hashed

EmosewaMC: Changes were reverted.  Still worth noting that all CI is broken for windows (and mac for whatever reason) until mariadb addresses the ssl issue (or we could put in a workaround)

Xiphoseer: Huh, interesting. What kind of SSL problem do they have? I don't see anything obvious on their end with https://dlm.mariadb.com/2531525/Connectors/cpp/connector-cpp-1.0.2/mariadb-connector-cpp-1.0.2-win64.msi, is there a tracking issue / status page on this somewhere?

aronwk-aaron: I've dug around and don't see any issue for it in their Jira, but that was a few days ago, I'll check again and make a ticket in there if there's isn't one

aronwk-aaron: Also, to be clear, the issue is that GitHub actions filter/block http requests, so this is just an issue here. And yeah finding the SSL info for that page is weird cause browsers are starting to hide that info from users now. But from what I can tell, there's either just no cert, a bad dummy cert, or an expired cert

Xiphoseer: Afaict from online tools like https://dnschecker.org/ssl-certificate-examination.php and https://redirectdetective.com, the dlm.mariadb.com link serves a 302 redirect to storage.googleapi.com and both have a valid certificate chain.

Is there some region based IP resolution or MITM going on that GH Actions are hitting? I had a case with Visual Studio in the past where there US servers worked but the european ones were broken.

dGame/dGameMessages/GameMessages.cpp
```diff
@@ -2401,63 +2401,25 @@ void GameMessages::HandleControlBehaviors(RakNet::BitStream* inStream, Entity* e
 }
 
 void GameMessages::HandleBBBSaveRequest(RakNet::BitStream* inStream, Entity* entity, const SystemAddress& sysAddr) {
-	/*
```
aronwk-aaron, line 2404: I will make a non-blocking protest against the removal of this artwork

aronwk-aaron (approved):

Jettford (approved):
Removal of the "notice" of the cave of dragons, is unacceptable. I will not stand for it.

PR #729, opened 2022-08-14
Added windows build commands  to build script

HailStorm32:
Added "intelligence" to the build script to run different build commands if on Windows.

Also, given this uses the OSTYPE variable, the ability to run different commands for different OS's could be added in the future.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -9,11 +9,6 @@ cmake ..
 if [[ "$OSTYPE" =~ ^msys ]]; then
 
 	cmake --build . --config Release
-	
-	# Move over binaries
-	if test -f "Release/MasterServer.exe"; then
-		mv Release/*.exe .
-	fi
 else
 
 	# Run make to build the project. To build utilizing multiple cores, append `-j` and the amount of cores to utilize, for example `make -j8`
--- a/build.sh
+++ b/build.sh
@@ -5,15 +5,8 @@ cd build
 # Run cmake to generate make files
 cmake ..
 
-# Windows build
-if [[ "$OSTYPE" =~ ^msys ]]; then
-
-	cmake --build . --config Release
-else
-
-	# Run make to build the project. To build utilizing multiple cores, append `-j` and the amount of cores to utilize, for example `make -j8`
-	make
-fi
+# To build utilizing multiple cores, append `-j` and the amount of cores to utilize, for example `cmake --build . --config Release -j8'
+cmake --build . --config Release
 
 # Run migrations
 ./MasterServer -m
--- a/build.sh
+++ b/build.sh
@@ -5,8 +5,20 @@ cd build
 # Run cmake to generate make files
 cmake ..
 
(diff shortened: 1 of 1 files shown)


EmosewaMC (changes requested):
Personally not a fan of this.  File moving should ideally be handled by installing the target via cmake, not through a shell script.  This kind of circumvents the build system by doing work outside of it.  

On a side note, this doesn't copy over any of the other executables or shared libraries.

Edit: You do this by checking if Master is built/exists?  We shouldn't be assuming files exist if they arent checked for, nor should a copy be done like this with *.exe.  I have also made the conclusion that you are moving the exes because they are in the Release directory.  If this is the case then this PR is a no-go.  We are already working on making the build system properly ensure the dlls are in the same directory as the exes.

HailStorm32:
Correct, my only reason for moving the binaries was because they couldn't run in the Release directory. However, you bring up a valid point in that the way I went about it wasnt ideal. 

My question then is this. Do you want me to remove the part the moves the binaries and just tell the user to do that manually, or move the movement of the binaries to cmake? 

As it stands now, building on Windows requires the movement of files. It's not just the dlls that its missing, but also the log folder and .ini files. Hence why I moved only the binaries, as it was less to move. I presume the rework of the build system will take this into account?

EmosewaMC:
Yes the fixing of the build system to actually work on native windows will be fixed in #670 .

build.sh
```diff
@@ -5,8 +5,15 @@ cd build
 # Run cmake to generate make files
 cmake ..
 
-# Run make to build the project. To build utilizing multiple cores, append `-j` and the amount of cores to utilize, for example `make -j8`
-make
+# Windows build
+if [[ "$OSTYPE" =~ ^msys ]]; then
+
+	cmake --build . --config Release
```
EmosewaMC, line 11: We do not build in release for our binaries.  We build in RelWithDebInfo.  Why exactly does make not work here?  This seems pointless at the moment

HailStorm32: I was just following what was on the README. For windows it says to use cmake instead of make. I can look into using make instead, if the README is outdated

EmosewaMC: Try using make instead if possible.

EmosewaMC: make looks to be a gnu exclusive command.

EmosewaMC: Instead of hard coding a check for windows I think it'll be more ideal to make a version of this script that doesn't need any checks or different commands for different OS's since we use cmake.

HailStorm32: So is the plan then to change from using `make` to using `cmake --build . --config Release` for all systems?

EmosewaMC: for the build script yes - keep it consistent across platforms so there is uniformity (and so we dont need to check for os type)

HailStorm32:
@EmosewaMC removed OS check and switched to cmake

PR #732, opened 2022-08-16
Selective saving for map and location

aronwk-aaron:
Don't save the map and char location info if we are in an instanced Zone

tested going to a prop and making sure everything else saved but I got put back in the origin world.

### DIFF
--- a/dGame/Character.cpp
+++ b/dGame/Character.cpp
@@ -301,20 +301,22 @@ void Character::SaveXMLToDatabase() {
 		character->SetAttribute("gm", m_GMLevel);
 		character->SetAttribute("cc", m_Coins);
 
+		auto zoneInfo = dZoneManager::Instance()->GetZone()->GetZoneID();
 		// lzid garbage, binary concat of zoneID, zoneInstance and zoneClone
-		if (Game::server->GetZoneID() != 0) {
-			auto zoneInfo = dZoneManager::Instance()->GetZone()->GetZoneID();
-			uint64_t lzidConcat = zoneInfo.GetCloneID();
-			lzidConcat = (lzidConcat << 16) | uint16_t(zoneInfo.GetInstanceID());
-			lzidConcat = (lzidConcat << 16) | uint16_t(zoneInfo.GetMapID());
-			character->SetAttribute("lzid", lzidConcat);
-			character->SetAttribute("lnzid", GetLastNonInstanceZoneID());
+		if ((zoneInfo.GetMapID() != 0) && !(1700 > zoneInfo.GetMapID() && zoneInfo.GetMapID() > 1600)) {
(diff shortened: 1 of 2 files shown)


dGame/dComponents/ControllablePhysicsComponent.cpp
```diff
@@ -11,6 +11,7 @@
 #include "CDClientManager.h"
 #include "EntityManager.h"
 #include "Character.h"
+#include "dZoneManager.h"
```
aronwk-aaron, line 14: My only concern is if this was the best way to be able to check if we are in a zone we don't want t save in from the component.

dGame/Character.cpp
```diff
@@ -301,20 +301,22 @@ void Character::SaveXMLToDatabase() {
...
-			lzidConcat = (lzidConcat << 16) | uint16_t(zoneInfo.GetMapID());
-			character->SetAttribute("lzid", lzidConcat);
-			character->SetAttribute("lnzid", GetLastNonInstanceZoneID());
-
-			//Darwin's backup:
-			character->SetAttribute("lwid", Game::server->GetZoneID());
-
-			// Set the target scene, custom attribute
-			character->SetAttribute("tscene", m_TargetScene.c_str());
+		if ((zoneInfo.GetMapID() != 0) && !(1700 > zoneInfo.GetMapID() && zoneInfo.GetMapID() > 1600)) {
```
EmosewaMC, line 306: lup worlds have a should save position flag in the database.  use this instead here.  This comparison is just comparing magic numbers and is not ideal.

EmosewaMC (changes requested):

dGame/Character.cpp
aronwk-aaron: oh yeah, there's the whole LUPZoneIDs table

dGame/Character.cpp
```diff
@@ -301,20 +301,22 @@ void Character::SaveXMLToDatabase() {
...
-			character->SetAttribute("lzid", lzidConcat);
-			character->SetAttribute("lnzid", GetLastNonInstanceZoneID());
-
-			//Darwin's backup:
-			character->SetAttribute("lwid", Game::server->GetZoneID());
-
-			// Set the target scene, custom attribute
-			character->SetAttribute("tscene", m_TargetScene.c_str());
+		if ((zoneInfo.GetMapID() != 0)) {
+			if (zoneInfo.GetCloneID() == 0) {
```
EmosewaMC, line 307: This can be simplified to if zoneInfo.getMapID != 0 && getCloneID == 0

EmosewaMC (changes requested):
Some simplification of logic gate.  This applies to both sets of changes

EmosewaMC (approved):

PR #735, opened 2022-08-17
Fix death planes

EmosewaMC:
Hotfix to move the position of the death barrier down 13.521004 units so we effectively only extend its hitbox in the -Y direction as opposed to the +Y direction, resolving an issue in Battle of Nimbus Station where the death plane was too tall


aronwk-aaron (approved):
love it

PR #738, opened 2022-08-18
Add imagimeter visibility script

aronwk-aaron:
Implements the imagimeter visibility script when completing the QB for the imagimeter
tested that it does what it's supposed to do


EmosewaMC (approved):
lgtm

Issue #746, opened 2022-08-24
Tracking Issue: Missing Scripts


aronwk-aaron:
This issue is to track missing scripts reported by the server per world.
This list is non-exhaustive, and will contain duplicates.
This will help to serve as a guide as to what scripts aren't implemented in general.

Venture Explorer
 - [ ] `scripts\ai\AG\L_AG_SHIP_SHAKE.lua` Currently lumped into `AgSpaceStuff` Split out?

Return to the Venture Explorer
 - [x] `scripts\ai\FV\L_FV_MAELSTROM_GEYSER.lua` #748 

Avant Gardens
 - [ ] `scripts\zone\AG\L_ZONE_AG.lua`
 - [ ] `scripts\02_server\Map\General\L_ENEMY_CLEAR_THREAT.lua`
 - [x] `scripts\02_server\Map\General\L_QB_SPAWNER.lua` #761 
 - [x] `scripts\ai\AG\L_AG_QB_Wall.lua` #761 
 - [ ] `scripts\ai\AG\L_AG_SENTINEL_GUARD.lua`

Avant Gardens Survival
 - [ ] `scripts\ai\AG\L_AG_SPIDER_BOSS_MESSAGE.lua`

Nimbus Station
 - [ ] `scripts\zone\NS\L_ZONE_NS.lua`
 - [ ] `scripts\ai\GF\L_GF_RACE_INSTANCER.lua` Needed to add double click and dragging functionality to racing
 - [ ] `scripts\ai\AG\CONCERT_STAGE.lua`
 - [ ] `scripts\ai\NS\L_NS_CAR_MODULAR_BUILD.lua`

Pet Cove
 - [x] `scripts\ai\MINIGAME\Objects\MINIGAME_BLUE_MARK.lua` #1349
 - [ ] `scripts\ai\PETS\PET_BLOCKER.lua`
 - [ ] `scripts\ai\PETS\PET_FLEA_MISSION.lua`
 - [ ] `scripts\ai\ACT\L_ACT_PET_INSTANCE_EXIT.lua`

Nimbus Station Shooting Gallery
 - [ ] `scripts\ai\MINIGAME\SG_NS\L_ZONE_SG_NS_SERVER.lua`
 - [ ] `scripts\ai\MINIGAME\SG_GF\SERVER\SG_CANNON_INSTANCE_ACTOR.lua`
[...]

PR #748, opened 2022-08-25
Add FV Geyser script

aronwk-aaron:
They show up and they hurt when hit by them.
This relates to #746 and resolves 2 places this script was being use, FV and RtVE


dScripts/FvMaelstromGeyser.h
```diff
@@ -0,0 +1,18 @@
...
+	void OnTimerDone(Entity* self, std::string timerName) override;
+
+
+private:
+	std::string m_StartSkillTimerName = "startSkill";
+	float m_StartSkillTimerTime = 2.0;
+	std::string m_KillSelfTimerName = "killSelf";
+	float m_KillSelfTimerTime = 5.5;
+	uint32_t m_SkillID = 831;
+	uint32_t m_BehaviorID = 15500;
```
EmosewaMC, line 17: Use const for all of these if possible.

aronwk-aaron: Bah, right, keep forgetting

aronwk-aaron: Will fix in both

EmosewaMC (approved):

PR #750, opened 2022-08-25
Force field effect Script

aronwk-aaron:
Force Field effect script that is used in LEGO Club and Starbase 3001.
Small issue of it not working due to there being no phantom collision coponent on the two objects that use this, but that's an issue to solve in another PR


aronwk-aaron:
> Can you test that this works with an OnTimerDone call or something first? I'd rather not submit this under the guise that it _might_ work or might not.

Can do

aronwk-aaron:
> Can you test that this works with an OnTimerDone call or something first? I'd rather not submit this under the guise that it _might_ work or might not.

Done and can confirm that the effect plays correctly

aronwk-aaron:
TODO: maually define physics sizes since we can't read hkx

EmosewaMC:
At the moment these scripts are unable to be used.  Their physics are too complex for our current system.

PR #755, opened 2022-08-28
Add blue mark script

aronwk-aaron:
possibly from crusty ancient relics, but it's on 5 objects used in game currently, so won't hurt to implement it even if it does nothing.

It does what it's supposed to and doesn't crash the server


EmosewaMC:
What does this fix?  How are you testing that functionality is correct?

aronwk-aaron:
> What does this fix? How are you testing that functionality is correct?

¯\_(ツ)_/¯
It possibly does nothing. but some client scripts (mostly deprecated ones) to look for that notification and do stuff. probably just legacy stuff that we honestly don't care about. but since it's not hurting anything to have it, why not.

EmosewaMC:
If you are unable to test it then this is a no-go for merging on my end.  There has to be a way to test the functionality and ensure its working correctly before asking people to use something untested

aronwk-aaron:
I tested that it broke nothing when going to each world that had it

EmosewaMC:
> I tested that it broke nothing when going to each world that had it

Thats not what matters.  What matters is the script does what it says it does, not that current stuff doesnt break.  You've got the latter covered, but there if you haven't seen what this changes and cant confirm it works then I cannot approve this.

aronwk-aaron:
After dinner more digging through scripts. It's looks like it's mainly legacy from the old racing system when colored markers were used. And would notify the minigame controller of it's location for certifian actions. Which we do propagate notifications to the receiver's scripts which is where I see this being used on older scripts.

May not provide any functionality now, but if someone wants to port DLU to an alpha version, may be useful to have.

Jettford:
Creating pull requests for any of those reasons are not good reasons, I insist that this not be merged. I can understand adding scripts that make a minor change to a feature that was maybe even bugged in live but was intended to be enabled but random scripts that may assist people if they want "port DLU to an alpha version" is just an excessive change and **even** in a situation where that might happen (of which I highly doubt will ever occur), that does still need to be tested for functionality at a bare minimum.

aronwk-aaron:
re-opening as this is used in FV seige

Jettford:
The object is used there but the functionality of this script is not.

PR #765, opened 2022-09-02
Fix Windows not working due to libbcrypt not working on Ninja

EmosewaMC:
Reverts back to file globing for libbcrypt to fix building on Windows Only

Windows compiles now.  Checking other platforms


EmosewaMC:
Compiles on Windows Only Configure locally and Ubuntu locally.  Once macos CI passes this is all set.

EmosewaMC:
Checking ARM before merge

EmosewaMC:
Compiles, runs and allows all servers to connect to clients.

PR #766, opened 2022-09-02
Script for aborting a wbl zone transfer

aronwk-aaron:
The client will sent this message if it wants to abort the zone transfer.
`FUN_00c46790` is the function in the client that will send an event to fire this server side

There will be a few more WBL scripts, which is why i started a section for it


EmosewaMC:
> `FUN_00c46790` is the function in the client that will send an event to fire this server side (from what I understand)

Why do we care what specific function from the client sends this?  All that matters is the server script works correctly and is tested to send the player to zone 1600.

aronwk-aaron:
> Why do we care what specific function from the client sends this? All that matters is the server script works correctly and is tested to send the player to zone 1600.

Just to explain why there isn't something else calling it from the server or another script. I.E. that it is something built into the client.
And yes, I manually called the script and it did the zone transfer correctly

EmosewaMC:
But does the object that uses this script function correctly?  Yes the code works but does the object that uses it actually call it?

EmosewaMC:
My only problem with that is you had to modify code in order to call that event and then put it back in this state...  Not sure how I feel about approving code that isn't tested in its final state

aronwk-aaron:
Scripts are the closet thing to source code we have from the live game. Take that as you wish

EmosewaMC:
> Scripts are the closet thing to source code we have from the live game. Take that as you wish

Im not talking about mirroring the live scripts, im talking about ensuring that the code in its final state has been tested and is working.

PR #767, opened 2022-09-03
Add LupGenericInteraction and Robot Citizen script

aronwk-aaron:
This is the coverage remaining RC scripts, the ones left are kinda weird so waiting on those.

Generic interact just plays an animation on interaction, and instead of making the same script like 6 times for the differently colored robot citizens, there's just one script and rfind is used to match all the scripts for them.

Tested that the animations work and look at works, however the pathing does not work.
Not sure if that's an issue with this or with moving platforms in general


dScripts/LupGenericInteract.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+#include "CppScripts.h"
+
+class LupGenericInteract : public CppScripts::Script {
+public:
+	void OnUse(Entity* self, Entity* user) override;
+private:
+	const uint32_t frakjawChestId = 16486;
```
EmosewaMC, line 8: Copy and paste error?

aronwk-aaron: Yes, ty. I usually copy another script she just that as a base. Got distracted with moving platform stuff and missed this

PR #769, opened 2022-09-04
ignore empty.lua and empty scripts

aronwk-aaron:
and return early if it's an ignored script
tested that ignored scripts are still ignored and not ignored scripts still load properly.

### DIFF
--- a/dScripts/CppScripts.cpp
+++ b/dScripts/CppScripts.cpp
@@ -294,12 +294,6 @@
 //Big bad global bc this is a namespace and not a class:
 InvalidScript* invalidToReturn = new InvalidScript();
 std::map<std::string, CppScripts::Script*> m_Scripts;
-const std::vector<std::string> m_IgnoredScripts = {
-	"",
-	"scripts\\02_server\\Enemy\\General\\L_SUSPEND_LUA_AI.lua",
-	"scripts\\02_server\\Enemy\\General\\L_BASE_ENEMY_SPIDERLING.lua",
-	"scripts\\empty.lua"
-};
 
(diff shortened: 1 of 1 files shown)


aronwk-aaron:
definitely open to improvements, not sure if searching the vector of strings is the most efficient way to do that, but form cursory searches and tests say this is good enough since the the vector that is being searched is finite and not large

EmosewaMC:
When using std::find use a list instead.

dScripts/CppScripts.cpp
```diff
@@ -308,6 +314,9 @@ CppScripts::Script* CppScripts::GetScript(Entity* parent, const std::string& scr
 
 	script = invalidToReturn;
 
+	// check for ignored scripts and return early
+	if (std::find(m_IgnoredScripts.begin(), m_IgnoredScripts.end(), scriptName) != m_IgnoredScripts.end()) return script;
```
EmosewaMC, line 318: We should not be returning early here.  If returning early then we would not cache these scripts for faster lookup for other objects.  See line 868

EmosewaMC (changes requested):

dScripts/CppScripts.cpp
aronwk-aaron: ah will update.

EmosewaMC: IMHO it's better to just keep the current else if standard we have for script lookups than to iterate over a list, creating unneeded overhead.  At the end of the day it's still just string comparisons so adding a list of strings just seems extra.

EmosewaMC (approved):

PR #784, opened 2022-10-22
fix reading respawn for older maps

aronwk-aaron:
this fixes reading in the respawn timer on some old maps


EmosewaMC (approved):
Under the assumption that these changes are tested to confirm respawn timers that are u32 are unchanged this is fine to merge.

aronwk-aaron:
> Under the assumption that these changes are tested to confirm respawn timers that are u32 are unchanged this is fine to merge.

yes, i checked this, sorry for forgetting to mentions this

Jettford (approved):

PR #787, opened 2022-10-25
NPC Pathing

aronwk-aaron:
This PR is for implementing NPC pathing


EmosewaMC:
holy mother of logs

dGame/dComponents/ControllablePhysicsComponent.cpp
```diff
@@ -33,6 +33,10 @@ ControllablePhysicsComponent::ControllablePhysicsComponent(Entity* entity) : Com
 	m_PickupRadius = 0.0f;
 	m_DirtyPickupRadiusScale = true;
 	m_IsTeleporting = false;
+	m_AttachedPath = entity->GetVarAsString(u"attached_path");
+	m_PathWaypoint = entity->GetVarAs<int>(u"attached_path_start");
+	m_PathSpeed = 2.5f;
```
Wincent01, line 38: The PhysicsComponent has a speed variabel. Check the movement ai component.

aronwk-aaron: is that the same one from the movementAI component? I know the controllable physics comp has a speed varaible, but it's not read in here. I plan on reading it in and using it evetually, but didn't want to go down that rabbit hole yet, since most of the have a defined speed on their path waypoint configs

dZoneManager/Zone.cpp
```diff
@@ -539,8 +539,13 @@ void Zone::LoadPath(std::ifstream& file) {
 					BinaryIO::BinaryRead(file, character);
 					value.push_back(character);
 				}
-				LDFBaseData* ldfConfig = LDFBaseData::DataFromString(parameter + "=" + value);
-				waypoint.config.push_back(ldfConfig);
+				LDFBaseData* ldfConfig = nullptr;
```
aronwk-aaron, line 542: Movement path waypoint configs to not have type data, so just make everything a string and let someone else deal with it when they need to use it

dScripts/SGCannon.cpp
```diff
@@ -287,38 +287,27 @@ void SGCannon::OnActivityTimerDone(Entity* self, const std::string& name) {
 			Game::logger->Log("SGCannon", "Spawning enemy %i on path %s", toSpawn.lot, path->pathName.c_str());
 
 			auto* enemy = EntityManager::Instance()->CreateEntity(info, nullptr, self);
-			EntityManager::Instance()->ConstructEntity(enemy);
+			if (enemy) {
```
aronwk-aaron, line 290: Entity construction handles setting up the movement paths now

dScripts/ActivityManager.cpp
```diff
@@ -124,8 +124,6 @@ void ActivityManager::ActivityTimerStart(Entity* self, const std::string& timerN
 	auto* timer = new ActivityTimer{ timerName, updateInterval, stopTime };
 	activeTimers.push_back(timer);
 
-	Game::logger->Log("ActivityManager", "Starting timer '%s', %f, %f", timerName.c_str(), updateInterval, stopTime);
```
aronwk-aaron, line 127: Removing spammy logs since I'm testing with AGS, as it has a bunch of paths that are meant to bring the enemies into the map towards the players, this will make AGS harder :smiling_imp:

dGame/Entity.cpp
```diff
@@ -317,15 +317,6 @@ void Entity::Initialize() {
 		m_Components.insert(std::make_pair(COMPONENT_TYPE_SOUND_TRIGGER, comp));
 	}
 
-	//Check to see if we have a moving platform component:
```
aronwk-aaron, line 320: Moved down to after things like ememies and pets get combat and movement AI, so that we can just update their movement AI to have paths and add movement AI to things that only have movement paths and

PR #789, opened 2022-10-26
Add automatic cdclient migration runner support and setup

EmosewaMC:
Add support to automatically migrate and update CDServers with new migrations.  Also adds support to simplify the setup process by simply putting the fdb in the res folder and letting the server convert it to sqlite.

This reduces the amount of back and forth when setting up a server and also allows us to fix bugs by using cdclient migrations as opposed to special case code.

It is likely up for debate whether we should handle converting the database to sqlite or require the user to do this.  Since we already require python and are cloning a repository which has tools to convert to sqlite from fdb.

Tested that starting a server with just the fdb in ./build/res properly copies and converts the database to sqlite and applies all migrations.  Went into Block yard and was able to open the mailbox (a required script in the migrations).
Tested that starting a server with the sqlite file has no changes aside from automatically applying the migrations that have not been run yet.
Tested that the migrations do not cause issues if run twice.  Since no columns are being added there was no issue planned here anyways but tested just in case.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -14,8 +14,8 @@
 	path = thirdparty/mariadb-connector-cpp
 	url = https://github.com/mariadb-corporation/mariadb-connector-cpp.git
 	ignore = dirty
-[submodule "thirdparty/docker-utils"]
-	path = thirdparty/docker-utils
+[submodule "thirdparty/utils"]
+	path = thirdparty/utils
 	url = https://github.com/lcdr/utils.git
 [submodule "thirdparty/LUnpack"]
 	path = thirdparty/LUnpack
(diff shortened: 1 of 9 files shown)


aronwk-aaron (dismissed):
Absolutely amazing work! this and the automated mysql database migrations will make setup a lot easier for newcomers and will help ease the existential dread of making changes via migrations.

My only comments are:
The readme needs to be updated with this info that this is handled automatically.
There could be a migrations table to avoid it running every time (non-blocking).

EmosewaMC:
> Absolutely amazing work! this and the automated mysql database migrations will make setup a lot easier for newcomers and will help ease the existential dread of making changes via migrations.
> 
> My only comments are: The readme needs to be updated with this info that this is handled automatically. There could be a migrations table to avoid it running every time (non-blocking).

Noted for Readme, there is a migrations table to avoid it running every time.  The migrations for cdserver are incredibly slow so this is imperative to exist

aronwk-aaron (approved):

Jettford (approved):

PR #795, opened 2022-10-29
Implement a shared config between servers

Jettford:


EmosewaMC (approved):
lgtm

EmosewaMC:
I tested that servers with scattered missing configs correctly used sharedconfig.ini as a backup, and since new servers would not have these configs in the first place, it can be proven that sharedconfig.ini will not cause any issues since the once in the other ini files effectively act as overrides should you for some reason want different databases for different servers.

aronwk-aaron (approved):

Wincent01 (approved):

PR #796, opened 2022-10-30
Change AMFArray getters to use Templates and fix CI halting when one matrix fails

EmosewaMC:
All tests still pass and these tests are the only place these functions are used at the moment.

Also fixed matrices cancelling each other when one fails.


aronwk-aaron (approved):

Xiphoseer (approved):
LGTM

PR #804, opened 2022-11-01
Make changes to allow Windows to run under a debug environment 

Jettford:
- Replace all interaction of std::string and sqlString.

### DIFF
--- a/dCommon/GeneralUtils.cpp
+++ b/dCommon/GeneralUtils.cpp
@@ -50,6 +50,7 @@ bool _IsSuffixChar(uint8_t c) {
 
 bool GeneralUtils::_NextUTF8Char(std::string_view& slice, uint32_t& out) {
 	size_t rem = slice.length();
+	if (rem == 0) return false;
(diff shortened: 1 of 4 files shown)


EmosewaMC (approved):
Looks fine to me.  Are there more of these in the codebase?  If there are more we may want to change them as well in the future just to be consistent and not have this issue pop-up in the future.

aronwk-aaron:
> Looks good to me. Are there more of these in the codebase? If there are more we may want to consider changing them as well in the future just to be consistent and not have this issue pop-up in the future.

I did a bried overview and there are still some in some distance corners of the code base that use std::string.
I can take a look at updating them and adding to this PR if wanted

Jettford:
There definitely are more but these seem to be the only ones causing issues. This is some definite black magic bugginess.

aronwk-aaron (approved):
since this is only a debug issue and does not effect release, I'll go ahead an approve it. If other places cause issues, we can just update them later

PR #806, opened 2022-11-02
Update README

### LINKED ISSUE #227: DOC: Enhancement

**I apologize** in advance that the issue uses the "Feature Request" template, the "Documentation Improvement" template returns an error code 400.

Link to my gist: `https://gist.github.com/leolion3/ccf654ab60c8e110c65ef948da6af461`

The documentation is too imprecise in explaining how things are done for simple users.
Even though I and you may agree that if someone has absolutely no experience with the command line they should probably join someone else's server instead of trying to do it themselves, a lot of people still want to do it and head over to discord asking for help.

After helping quite a lot of users on discord with setting up their own servers/clients, here are a couple things in my gist that I would suggest adding:
- Detailed MariaDB Setup, including database user creation, tables creation, granting access to the user to the database and to the `dbo.UNIX_TIMESTAMP` subroutine (required for the leaderboards table as I found out in #202 and access is denied by default)
- A better overview of the resulting build folder structure that needs to be validated
- Better ini file setup
- A perhaps as simple as possible explanation of the different network types (users kept confusing localhost/LAN/WAN setup steps) and what port forwarding and firewall rules need to be set. (Again I agree that if they have no experience they should probably not set up a server, however since they are going to do it anyway we might as well help them)
- Brick-by-Brick building could be fixed by supplying a non-existing Github link, such as my own `leolion3.github.io/404` which returns an error code 404
- Adding a guide to manually update user data in MySQL, and manual name approval
- Assembly animated sign not showing up fix
- Perhaps a "Known Errors" section detailing more user set-up errors and how to deal with them 

_No response_

_No response_

### Additional context

_No response_

### LINKED ISSUE #805: DOC: README is missing new documentation for packed client and the shared config

- [X] I have validated that the documentation is wrong or missing on the latest version of the main branch of DarkflameServer

### Location of the documentation

https://github.com/DarkflameUniverse/DarkflameServer

### Documentation problem

Just a note for us to add documentation.  We're already working on it and am just making an issue as a note.

### Suggested fix for documentation

title

EmosewaMC:
The README is a touch out of date, so we've re-written pretty much the whole thing.

Also address an issue where you could not use a database connection if it was not on the default port 3306.

Fixes #805
Fixes #227


HailStorm32:
Will this play nicely with clients that have the "old" AG survival fix applied, or will users need to unapply the client fix?

EmosewaMC:
> Will this play nicely with clients that have the "old" AG survival fix applied, or will users need to unapply the client fix?

No change with clients that have the fix.  The absolute worst case is clients without the fix still don't work sometimes.  Clients with the fix will always have 100% success rate.

aronwk-aaron (approved):

README.md
```diff
@@ -60,19 +60,7 @@ You can either run `build.sh` when in the root folder of the repository:
...
-mkdir -p build
-cd build
-
-# Run CMake to generate make files
-cmake ..
-
-# To build utilizing multiple cores, append `-j` and the amount of cores to utilize, for example `cmake --build . --config Release -j8'
-cmake --build . --config Release
-```
+Or manually run the commands used in [build.sh](build.sh):
```
luxaritas, line 63:
```suggestion
Or manually run the commands used in [build.sh](build.sh).
```

README.md
```diff
@@ -181,40 +168,38 @@ shasum -a 256 <file>
...
+- Extract the file `cdclient.fdb` from your packed client (if you have an unpacked client this step is skipped.)
+- `cdclient.fdb` will be converted to `CDServer.sqlite`.
 
-#### Client database
-* Move the file `res/cdclient.fdb` from the unpacked client to the `build/res` folder on the server.
-* The server will automatically copy and convert the file from fdb to sqlite should `CDServer.sqlite` not already exist.
-* You can also convert the database manually using `fdb_to_sqlite.py` using lcdr's utilities.  Just make sure to rename the file to `CDServer.sqlite` instead of `cdclient.sqlite`.
-* Migrations to the database are automatically run on server start.  When migrations are needed to be ran, the server may take a bit longer to start.
+* Optionally, you can also convert the database manually using `fdb_to_sqlite.py` using lcdr's utilities.  Just make sure to rename the file to `CDServer.sqlite` instead of `cdclient.sqlite`.  
+* Migrations to the database are automatically run on server start.  Migrations address issues with data in the database that would normally cause bugs.  When migrations are needed to be ran, the server may take a bit longer to start.
```
luxaritas, line 182: The bullet styles are mismatched, but more importantly there should be some indentation of some of these

luxaritas (commented):
As a side note, while setup has been significantly simplified, it still feels like the documentation of setup is still a bit unclear because the setup steps are a bit scattered between other information and don't necessarily follow a clear order. I'm not fully sure how much can be done to improve that, but it may be useful to have a section right at the top that lists out the core steps like:
- Clone
- Install dependencies
- Set up database
- Build
- Set config options
- Run

aronwk-aaron:
Need to note cmake 3.18+ requitement somewhere

EmosewaMC:
Going to separate and put the AGS fix in its own PR.

aronwk-aaron (approved):

aronwk-aaron (approved):

README.md
```diff
@@ -18,210 +18,294 @@ Darkflame Universe is licensed under AGPLv3, please read [LICENSE](LICENSE). Som
...
+		* [Linux, Ubuntu and others](#linux-ubuntu-or-others-cmake)
+	* [Packages](#packages)
+		* [Linux](#linux-build-packages)
+		* [Windows](#windows-packages)
+		* [MacOS](#macos-build-packages)
+	* [Database](#database)
+		* [Linux](#linux-mariadb)
+		* [Windows](#windows-mariadb)
+		* [MacOS](#macos-mariadb)
+	* [Choosing the right client version for your server](#choosing-the-right-clent-version-for-your-server)
```
luxaritas, line 40: Inconsistent whitespace (looks like a spaces vs tabs issue?)

README.md
```diff
@@ -18,210 +18,294 @@ Darkflame Universe is licensed under AGPLv3, please read [LICENSE](LICENSE). Som
...
+* [In-game commands](#in-game-commands)
+* [Credits](#credits)
+	* [Active Contributors](#active-contributors)
+	* [Darkflame Universe Team](#darkflame-universe-team)
+		* [Research and tools](#research-and-tools)
+		* [Community management](#community-management)
+		* [Former contributors](#former-contributors)
+		* [Logo](#logo)
+		* [Special thanks](#special-thanks)
+
```
luxaritas, line 68: These are not setup steps

README.md
```diff
@@ -18,210 +18,294 @@ Darkflame Universe is licensed under AGPLv3, please read [LICENSE](LICENSE). Som
...
-### Prerequisites
-#### Clone the repository
+Darkflame Universe is a server emulator and does not distribute any LEGO® Universe files. A separate game client is required to setup this server emulator and play the game, which we cannot supply. Users are strongly suggested to refer to the safe checksums listed [here](#lego®-universe-11064) to see if a client will work.
+
+## Steps to setup server
+* [Clone this repository](#clone-the-repository)
+* [Install dependencies](#install-dependencies)
+    * [cmake](#cmake)
+		* [Windows and MacOS](#windows-cmake)
+		* [Linux, Ubuntu and others](#linux-ubuntu-or-others-cmake)
```
luxaritas, line 31: In all cases, I'd omit adding line items for different system types. This makes the list of steps look significantly longer than it actually is, and doesn't really provide any benefit as an overview

README.md
```diff
@@ -18,210 +18,294 @@ Darkflame Universe is licensed under AGPLv3, please read [LICENSE](LICENSE). Som
...
+Then run the following command
 ```bash
 git clone --recursive https://github.com/DarkflameUniverse/DarkflameServer
 ```
-#### Python
 
-Some tools utilized to streamline the setup process require Python 3, make sure you have it installed.
+## Install dependencies
 
+### CMake
```
luxaritas, line 80: CMake is just another package. I'd just include cmake as a package as with the other packages (either as something to install via apt, download on windows, etc) and then add a note under that section to verify the cmake version and that on Linux if it isn't available in your package repos, to eg grab it from kitware

README.md
```diff
@@ -18,210 +18,294 @@ Darkflame Universe is licensed under AGPLv3, please read [LICENSE](LICENSE). Som
...
-# Run CMake to generate build files
-cmake .. -DOPENSSL_ROOT_DIR=/path/to/openssl
+sudo apt install openssl
 
-# Get cmake to build the project. If make files are being used then using make and appending `-j` and the amount of cores to utilize may be preferable, for example `make -j8`
-cmake --build . --config Release
+sudo apt install mariadb-server
+
+# On some systems, this package does not exist.
+sudo apt install zlib
```
luxaritas, line 152: You can turn this into a single apt command, which will make this significantly more condensed and not really lose you anything

README.md
```diff
@@ -18,210 +18,294 @@ Darkflame Universe is licensed under AGPLv3, please read [LICENSE](LICENSE). Som
...
-cd build
+brew install openssl
 
-:: Run CMake to generate make files
-cmake .. -DMARIADB_BUILD_SOURCE=ON
+brew install gcc
 
-:: Run CMake with build flag to build
-cmake --build . --config Release
+brew install mariadb
```
luxaritas, line 170: Similarly here, I imagine you can just run a single brew command

README.md
```diff
@@ -18,210 +18,294 @@ Darkflame Universe is licensed under AGPLv3, please read [LICENSE](LICENSE). Som
...
-# Make sure the install is up to date
-apt update && apt upgrade
-
-# Make sure the gcc, cmake, and build-essentials are installed
-sudo apt install gcc
-sudo apt install cmake
-sudo apt install build-essential
+sudo /etc/init.d/mysql start
 ```
+<font size="4">**You will need to run this command every time you restart your environment**</font>
```
luxaritas, line 181: Why do this vs `systemctl start mysql`?

README.md
```diff
@@ -18,210 +18,294 @@ Darkflame Universe is licensed under AGPLv3, please read [LICENSE](LICENSE). Som
...
+cmake .. -DOPENSSL_ROOT_DIR=/path/to/openssl
 
-After the server has been built there should be four `ini` files in the build director: `sharedconfig.ini`, `authconfig.ini`, `chatconfig.ini`, `masterconfig.ini`, and `worldconfig.ini`. Go through them and fill in the database credentials and configure other settings if necessary.
+# Get cmake to build the project. If make files are being used then using make and appending `-j` and the amount of cores to utilize may be preferable, for example `make -j8`
+cmake --build . --config Release
+```
 
-#### Migrations
+### Building on **all** other platforms
+You can either run `build.sh` when in the root folder of the repository:
```
luxaritas, line 266: Is this really accurate for Windows?

README.md
```diff
@@ -18,210 +18,294 @@ Darkflame Universe is licensed under AGPLv3, please read [LICENSE](LICENSE). Som
...
 
-* All that you need to do is create a database to connect to.  As long as the server can connect to the database, the schema will always be kept up to date when you start the server.
+```bash
+# Create the build directory, preserving it if it already exists
+mkdir -p build
+cd build
 
-#### Configuration
+# Run CMake to generate build files
+cmake .. -DOPENSSL_ROOT_DIR=/path/to/openssl
```
luxaritas, line 259: Is it possible to modify the bash script to try and automatically find the openssl location? Maybe not, but it _feels_ like this shouldn't be necessary

README.md
```diff
@@ -18,210 +18,294 @@ Darkflame Universe is licensed under AGPLv3, please read [LICENSE](LICENSE). Som
...
+You can either run `build.sh` when in the root folder of the repository:
 
-The database is automatically setup and migrated to what it should look like for the latest commit whenever you start the server.
+```bash
+./build.sh
+```
+
+Or manually run the commands used in [build.sh](build.sh).
+
+## Updating your server
```
luxaritas, line 274: Does this belong here? I feel like it maybe fits better in the user guide

README.md
```diff
@@ -18,210 +18,294 @@ Darkflame Universe is licensed under AGPLv3, please read [LICENSE](LICENSE). Som
...
+### Setting up the client dependencies
+Darkflame Universe can run with either a packed or an unpacked client as shown [here](#lego®-universe-11064)
+Navigate to `sharedconfig.ini` and fill in the `client_path` field with the location of your client.
 
+### Setting up the database
+Darkflame Universe utilizes a MariaDB database for account and character information.
+
+Initial setup can vary drastically based on which operating system or distribution you are running; there are instructions out there for most setups, follow those and come back here when you have a database up and running.
+
+Note that all that you need to do is create a database to connect to.  As long as the server can connect to the database, the schema will always be kept up to date when you start the server.
```
luxaritas, line 298: This seems like duplicate information given that it was already specified in the dependency section

README.md
```diff
@@ -18,210 +18,294 @@ Darkflame Universe is licensed under AGPLv3, please read [LICENSE](LICENSE). Som
...
 
-### Windows builds (WSL)
-This section will go through how to install [WSL](https://docs.microsoft.com/en-us/windows/wsl/install) and building in a Linux environment under Windows. WSL requires Windows 10 version 2004 and higher (Build 19041 and higher) or Windows 11.
-
-#### Open the Command Prompt application with Administrator permissions and run the following:
-```bash
-# Installing Windows Subsystem for Linux
-wsl --install
-```
+### Database
```
luxaritas, line 173: Maybe we could get rid of the subheadings under install dependencies entirely, and instead create a new section like "start database"? Also, don't you need to create an initial empty database?

README.md
```diff
@@ -230,204 +314,62 @@ Your build directory should now look like this:
...
-    * ...
 * ...
 
+Your `sharedconfig.ini` file should have:
+* The `client_path` option filled in with the location of your packed client.
+* A MariaDB database is setup with the credentials filled in
+
 ## Running the server
-If everything has been configured correctly you should now be able to run the `MasterServer` binary. Darkflame Universe utilizes port numbers under 1024, so under Linux you either have to give the binary network permissions or run it under sudo.
+If everything has been configured correctly you should now be able to run the `MasterServer` binary.  Darkflame Universe utilizes port numbers under 1024, so under Linux you either have to give the binary network permissions or run it under sudo.
```
luxaritas, line 326: Note that it's in the build folder?

README.md
```diff
@@ -230,204 +314,62 @@ Your build directory should now look like this:
...
 
 ### First admin user
 Run `MasterServer -a` to get prompted to create an admin account. This method is only intended for the system administrator as a means to get started, do NOT use this method to create accounts for other users!
 
-### Account Manager
-
-Follow the instructions [here](https://github.com/DarkflameUniverse/AccountManager) to setup the DLU account management Python web application. This is the intended way for users to create accounts.
+### Account management tool (Nexus Dashboard)
+Follow the instructions [here](https://github.com/DarkflameUniverse/NexusDashboard) to setup the DLU Nexus Dashboard web application. This is the intended way for users to create accounts.
```
luxaritas, line 332: Maybe note this is specifically in the case of multi-user setups, as it's not needed if you're just using it yourself and create a GM account as above?

README.md
```diff
@@ -230,204 +314,62 @@ Your build directory should now look like this:
...
+## Brick-By-Brick building
+Should you choose to do any brick building, you will want to have some form of a http server that returns a 404 error since we are unable to emulate live User Generated Content at the moment.  If you attempt to do any brick building without a 404 server running properly, you will be unable to load into your game.  Python is the easiest way to do this, but any thing that returns a 404 should work fine.
+* Note: the client hard codes this request on port 80.
+
+<font size="4">**If you do not plan on doing any Brick Building, then you can skip this step.**</font>
+
+### Allowing a user to build in Brick-by-Brick mode
+Brick-By-Brick building requires `PATCHSERVERIP=0:` and `UGCSERVERIP=0:` in the `boot.cfg` to point to a HTTP server which always returns `HTTP 404 - Not Found` for all requests. This can be most easily achieved by pointing both of those variables to `localhost` while having running in the background.
+```bash
+sudo python3 -m http.server 80
```
luxaritas, line 360: Is it work noting the need to install python, and to differentiate for Windows (which doesn't have sudo)?

luxaritas (changes requested):
Smattering of further adjustments I think would be helpful. This is more readable now, but I'm hoping we can work on condensing the instructions further so that it isn't as overwhelming.

As somewhat of an aside, it could be worth exploring splitting this into multiple "pages" instead of housing it all in the readme. You could use this approach to create separate guides for different OSes that's more concise and easier to follow (conventional for many pieces of server software), move detailed contextual information (eg, verifying client files) to separate pages, etc.

luxaritas (changes requested):
(Hit comment instead of request changes, formally noting that I want to see some stuff changed before merging)

README.md
HailStorm32: running the build script works on Windows

README.md
HailStorm32: > Why do this vs systemctl start mysql?

WSL doesn't support systemd, or at least it didn't last time I checked

luxaritas: Looks like it's available on newer WSL versions: https://devblogs.microsoft.com/commandline/systemd-support-is-now-available-in-wsl/. TBH neither systemd nor `/etc/init.d/mysql` are portable. It might be better to suggest `mysqld` or `mysqld_safe`, or even link to https://mariadb.com/kb/en/starting-and-stopping-mariadb-automatically/

aronwk-aaron (dismissed):

luxaritas (dismissed):
This looks MUCH better now! Thanks for taking the time to work on this. LGTM :)

aronwk-aaron (dismissed):

README.md
```diff
@@ -18,210 +18,181 @@ Darkflame Universe is licensed under AGPLv3, please read [LICENSE](LICENSE). Som
...
-git pull
-git submodule update --init --recursive
+sudo systemctl start mysql
+# If systemctl is not a known command on your distribution, try the following instead
+sudo service mysql start
 ```
-now follow the build section for your system
 
-## Setting up the environment
+<font size="4">**You will need to run this command every time you restart your environment**</font>
```
HailStorm32, line 130: If you change the above systemctl command to `sudo systemctl enable --now mysql` it will make it startup automatically on system restart.

Not sure about `sudo service mysql start`  though

README.md
```diff
@@ -18,210 +18,181 @@ Darkflame Universe is licensed under AGPLv3, please read [LICENSE](LICENSE). Som
...
-sudo apt install build-essential
 ```
 
-[**Follow the Linux instructions**](#linux-builds)
+## Database setup
+First you'll need to start MariaDB.
 
-### ARM builds
-AArch64 builds should work on linux and MacOS using their respective build steps. Windows ARM should build but it has not been tested
+For Windows the service is always running by default.
```
HailStorm32, line 116: Maybe make it clear somewhere in the Database setup that the MariaDB install needs the be on the default port `3306`. 

On Windows it allows the user to change the port on install, or will change it automatically if `3306` is in use. I've come across multiple users who either changed the default port on accident or on purpose and led to issues.

README.md
```diff
@@ -18,210 +18,181 @@ Darkflame Universe is licensed under AGPLv3, please read [LICENSE](LICENSE). Som
...
+This server has a few steps that need to be taken to configure the server for your use case.
+
+### Required Configuration
+Darkflame Universe can run with either a packed or an unpacked client.
+Navigate to `build/sharedconfig.ini` and fill in the following fields:
+* `mysql_host` (This is the IP address or hostname of your MariaDB server.  This is highly likely `localhost`)
+* `mysql_database` (This is the database you created for the server)
+* `mysql_username` (This is the user you created for the server)
+* `mysql_password` (This is the password for the user you created for the server)
+* `client_location` (This is the location of the client files.  This should be the folder path of a packed or unpacked client)
```
HailStorm32, line 185: Possibly make a note that the path to the client can't contain spaces. I had issues with the client folder name having a space on Windows.

README.md
```diff
@@ -230,204 +201,126 @@ Your build directory should now look like this:
...
 
 ### First admin user
 Run `MasterServer -a` to get prompted to create an admin account. This method is only intended for the system administrator as a means to get started, do NOT use this method to create accounts for other users!
 
-### Account Manager
+### Account management tool (Nexus Dashboard)
+**If you are just using this server for yourself, you can skip setting up Nexus Dashboard**
 
-Follow the instructions [here](https://github.com/DarkflameUniverse/AccountManager) to setup the DLU account management Python web application. This is the intended way for users to create accounts.
+Follow the instructions [here](https://github.com/DarkflameUniverse/NexusDashboard) to setup the DLU Nexus Dashboard web application. This is the intended way for users to create accounts.
```
HailStorm32, line 222: I think it would be best to further describe what ND does. ie change 

> This is the intended way for users to create accounts.

to 

> This is the intended way for users to create accounts and approve names/pets/properties

Or something along those lines. Just to make it clear that if they want to approve names and such, ND is how they would do so

README.md
```diff
@@ -230,204 +201,126 @@ Your build directory should now look like this:
...
 
-The admin level, or game master level, is specified in the `accounts.gm_level` column in the MySQL database. Normal players should have this set to `0`, which comes with no special privileges. The system administrator will have this set to `9`, which comes will all privileges. Admin level `8` should be used to give a player a majority of privileges without the safety critical once.
+While a character has a gmlevel of anything but `0`, some gameplay behavior will change. When testing gameplay, you should always use a character with a gmlevel of `0`.
 
-While a character has a gmlevel of anything but 0, some gameplay behavior will change. When testing gameplay, you should always use a character with a gmlevel of 0.
+# User guide
+Some changes to the client `boot.cfg` file are needed to play on your server.
```
HailStorm32, line 230: This line here seems a bit out of place given the current ordering of action items. Maybe move `Allowing a user to connect to your server` to before `Updating your server` or find a way to rephrase the line

HailStorm32 (commented):
Along with the suggested change to the README, did you also plan to add links to the source documentation in the README? I think it would be a good addition

README.md
luxaritas: Not sure if that's desirable - if they don't turn it off, it could use resources without them knowing. If it gets included, both should be given with that explanation

README.md
luxaritas: NB: When I tried (on Linux), I don't think spaces caused an issue. I didn't test thoroughly though

README.md
```diff
@@ -18,210 +18,181 @@ Darkflame Universe is licensed under AGPLv3, please read [LICENSE](LICENSE). Som
...
-cmake .. -DOPENSSL_ROOT_DIR=/path/to/openssl
-
-# Get cmake to build the project. If make files are being used then using make and appending `-j` and the amount of cores to utilize may be preferable, for example `make -j8`
-cmake --build . --config Release
+cmake --version
 ```
 
-### Windows builds (native)
-Ensure that you have either the [MSVC](https://visualstudio.microsoft.com/vs/) or the [Clang](https://github.com/llvm/llvm-project/releases/) (recommended) compiler installed. You will also need to install [CMake](https://cmake.org/download/). Currently on native Windows the server will only work in Release mode.
+If you are going to be using an ubuntu environment to run the server, you may need to get a more recent version of `cmake` than the packages available may provide.
```
Xiphoseer, line 78:
```suggestion
If you are going to be using an Ubuntu environment to run the server, you may need to get a more recent version of `cmake` than the packages available may provide.
```

README.md
```diff
@@ -18,210 +18,181 @@ Darkflame Universe is licensed under AGPLv3, please read [LICENSE](LICENSE). Som
...
-mkdir build
-cd build
+```bash
+# Remove the old version of CMake
+sudo apt purge --auto-remove cmake
 
-:: Run CMake to generate make files
-cmake .. -DMARIADB_BUILD_SOURCE=ON
+# Prepare for installation
+sudo apt update && sudo apt install -y software-properties-common lsb-release && sudo apt clean all
```
Xiphoseer, line 90: That's a typical stackoverflow thing. Ask people to install two new packages, just so the next command to run is slightly more generic.

IMO it would be better to just tell people to add a file to /etc/apt/sources.list.d and find out the ubuntu version name on their own (or tell them it's `jammy` now. iirc)

README.md
```diff
@@ -230,204 +201,126 @@ Your build directory should now look like this:
...
-While a character has a gmlevel of anything but 0, some gameplay behavior will change. When testing gameplay, you should always use a character with a gmlevel of 0.
+# User guide
+Some changes to the client `boot.cfg` file are needed to play on your server.
 
-## User guide
-A few modifications have to be made to the client.
+## Updating your server
+To update your server to the latest version navigate to your cloned directory
+```bash
+cd /path/to/DarkflameServer
```
Xiphoseer, line 235: It might make sense to remove the leading slash here not to confuse poeple they think they need to cd with an absolute path

Xiphoseer (approved):
LGTM, with some minor nits

README.md
HailStorm32: Hmm fair point, maybe added as a note or something

EmosewaMC: I too am not a fan of unknowingly background running applications so making this an opt in rather than an opt out would be ideal

README.md
EmosewaMC: I think it would be easier to make it a part of the database socket instead so people can change it if they want to

EmosewaMC:
> Along with the suggested change to the README, did you also plan to add links to the source documentation in the README? I think it would be a good addition

Done.  Its hastily thrown in there for now, a more detailed page will follow when I or others have time

aronwk-aaron (dismissed):

dDatabase/Database.cpp
```diff
@@ -30,15 +30,17 @@ void Database::Connect(const string& host, const string& database, const string&
...
+	const std::string TCP_PROTO = "tcp://";
     if (host.find(UNIX_PROTO) == 0) {
 		properties["hostName"] = "unix://localhost";
 		properties["localSocket"] = host.substr(UNIX_PROTO.length()).c_str();
     } else if (host.find(PIPE_PROTO) == 0) {
 		properties["hostName"] = "pipe://localhost";
 		properties["pipe"] = host.substr(PIPE_PROTO.length()).c_str();
     } else {
-		properties["hostName"] = host.c_str();
+		properties["hostName"] = (TCP_PROTO + host).c_str();
```
Xiphoseer, line 41: This seems like an unrelated change?

EmosewaMC: Not unrelated.  If you started MariaDB on any port other than 3306, than your server would fail to start.  This change addresses that issue and allows you to specify the port as is mentioned in the new README

EmosewaMC: c.f. https://github.com/DarkflameUniverse/DarkflameServer/blob/481bc037d53d3bab45f06bbcd1cf0cccb49f2418/README.md?plain=1#L186-L188

Xiphoseer: Yeah, sure, but adding `tcp://` to `hostName` seems like a bugged way of doing this, as per https://mariadb.com/docs/server/connect/programming-languages/cpp/connect. Something like `tcp://localhost:3306` is a *Compatibility Syntax Connection URI* according to that doc and we don't even use that parameter to `connect` because of claims it's buggy in the comment next to it.

What I'm getting at, is this is a code change and has different review criteria as a README, if you make a well-intentioned change in here.

EmosewaMC: I'm not entirely convinced as to why using `tcp://` for a tcp connection is bugged, its only bugged for the other two types because we never specify which one is being used to connect.  I will revert this and leave a note in the README to prepend `tcp://` if you plan to launch on a port besides 3306.  Is that ok?

Xiphoseer: Using `tcp` is the right thing, but in the first param of a `connect(uri, params)` call afaict.

Xiphoseer (approved):

aronwk-aaron (approved):

README.md
EmosewaMC: Will make an issue for this so we can get rid of this eventually.

PR #816, opened 2022-11-04
simplify  path fixing for packed vs unpacked

aronwk-aaron:
fix slashes for hasfile for unpacked client checking

Tested with unpacked client on linux
Tested with packed on linux


Jettford (changes requested):
This removes functionality?

EmosewaMC:
> This removes functionality?

He says hes tested with packed and unpacked clients - not sure how anything has changed?

Jettford:
He definitely has not tested enough.

Jettford (approved):

EmosewaMC:
Has chatting, character create, macros and testmap been tested?

EmosewaMC (approved):
lgtm

aronwk-aaron:
> Has chatting, character create, macros and testmap been tested?

yes, I did test all three of those, chatting and char create names were already good.
Testmap is the only thing that "HasFile" is used for since we don't care about what's in the file, just that it exists in the first place before we attempt to go to it

PR #818, opened 2022-11-05
Fix Pet Taming causing seg fault

EmosewaMC:
~~**THIS IS UNTESTED**!!!  Unit tests reveal this change worked however an integration test has not been run quite yet.  If someone would like to prop the build and test pet taming please let me know what is needed to address the issue~~

### DIFF
--- a/dCommon/dClient/AssetManager.cpp
+++ b/dCommon/dClient/AssetManager.cpp
@@ -67,6 +67,8 @@ bool AssetManager::HasFile(const char* name) {
 	auto fixedName = std::string(name);
 	std::transform(fixedName.begin(), fixedName.end(), fixedName.begin(), [](uint8_t c) { return std::tolower(c); });
 
+	// Special case for unpacked client have BrickModels in upper case
+	if (this->m_AssetBundleType == eAssetBundleType::Unpacked) GeneralUtils::ReplaceInString(fixedName, "brickmodels", "BrickModels");
 
 	std::replace(fixedName.begin(), fixedName.end(), '\\', '/');
 	if (std::filesystem::exists(m_ResPath / fixedName)) return true;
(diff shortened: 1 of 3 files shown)


aronwk-aaron:
Tested with packed and unpacked clients.
BrickModels is the only special case for not being fully lowercase, so just adding this one exclusion in the asset getter's makes a more centralized and seamless use of the AssetManager

aronwk-aaron (dismissed):

aronwk-aaron (approved):

EmosewaMC (commented):
lgtm

dCommon/dClient/AssetManager.cpp
```diff
@@ -92,17 +94,19 @@ bool AssetManager::GetFile(const char* name, char** data, uint32_t* len) {
...
 	std::replace(fixedName.begin(), fixedName.end(), '\\', '/'); // On the off chance someone has the wrong slashes, force forward slashes
-	auto realPathName = fixedName;
 
-	if (std::filesystem::exists(m_ResPath / realPathName)) {
+	// Special case for unpacked client have BrickModels in upper case
+	if (this->m_AssetBundleType == eAssetBundleType::Unpacked) GeneralUtils::ReplaceInString(fixedName, "brickmodels", "BrickModels");
+
+	if (std::filesystem::exists(m_ResPath / fixedName)) {
 		FILE* file;
 #ifdef _WIN32
```
Wincent01, line 103: I would prefer to go an abstraction level up to ifsteams here, avoids the ugly preprocessing. Doesn't require changing, just a note.

Wincent01 (approved):

dCommon/dClient/AssetManager.cpp
Jettford: I was doing this for speed reasons, no idea how much it actually matters.

PR #820, opened 2022-11-06
Add checks to AssetBuffers before they are used

aronwk-aaron:
to avoid crashing
Tested that this fixes crashing on non-existent macros.
When ahead and added checks where this is used to avoid accidental crashes.
Tested that those places are handled gracefully as well

### DIFF
--- a/dGame/dInventory/Item.cpp
+++ b/dGame/dInventory/Item.cpp
@@ -345,7 +345,7 @@ void Item::DisassembleModel() {
 	auto buffer = Game::assetManager->GetFileAsBuffer(lxfmlPath.c_str());
 
 	if (!buffer.m_Success) {
-		Game::logger->Log("Item", "Failed to load %s", lxfmlPath.c_str());
+		Game::logger->Log("Item", "Failed to load %s to disassemble model into bricks, check that this file exists", lxfmlPath.c_str());
 		return;
 	}
 
--- a/dGame/dInventory/Item.cpp
+++ b/dGame/dInventory/Item.cpp
@@ -344,6 +344,11 @@ void Item::DisassembleModel() {
 	std::string lxfmlPath = "BrickModels/" + GeneralUtils::SplitString(renderAssetSplit.back(), '.').at(0) + ".lxfml";
 	auto buffer = Game::assetManager->GetFileAsBuffer(lxfmlPath.c_str());
 
+	if (!buffer.m_Success) {
+		Game::logger->Log("Item", "Failed to load %s", lxfmlPath.c_str());
+		return;
+	}
+
 	std::istream file(&buffer);
 
 	result.finalize();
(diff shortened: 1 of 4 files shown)


dZoneManager/Zone.cpp
```diff
@@ -42,6 +42,12 @@ void Zone::LoadZoneIntoMemory() {
 	if (m_ZoneFilePath == "ERR") return;
 
 	AssetMemoryBuffer buffer = Game::assetManager->GetFileAsBuffer(m_ZoneFilePath.c_str());
+
+	if (!buffer.m_Success) {
+		Game::logger->Log("Zone", "Failed to load %s", m_ZoneFilePath.c_str());
+		return;
+	}
```
EmosewaMC, line 49: I'm not sure how I feel about letting the server continue running if the zone is unable to be loaded.  What are your thoughts on this throwing an exception instead since you kind of need the whole zone in order to play on the server.

This comment also applies to the lvl triggers below.  If we threw an exception at the very least we would have a known spot the exception would be located at.

EmosewaMC (dismissed):
Can you elaborate more on how you have tested the bounds of this PR?

While not intended, we should test if the files do not exist at all what would happen to the server.  
I also have some concerns about some of these checks allowing the server to keep running if files are unable to be found.

dGame/dUtilities/BrickDatabase.cpp
```diff
@@ -19,6 +19,11 @@ std::vector<Brick>& BrickDatabase::GetBricks(const std::string& lxfmlPath) {
 	}
 
 	AssetMemoryBuffer buffer = Game::assetManager->GetFileAsBuffer((lxfmlPath).c_str());
+
+	if (!buffer.m_Success) {
+		return emptyCache;
+	}
```
EmosewaMC, line 25: Perhaps we should also cache an empty cache instead of just leaving early here as well?  That way we arent repeatedly checking if the file exists when it didn't exist on server start.  What are your thoughts on this?

dGame/dInventory/Item.cpp
```diff
@@ -344,6 +344,11 @@ void Item::DisassembleModel() {
 	std::string lxfmlPath = "BrickModels/" + GeneralUtils::SplitString(renderAssetSplit.back(), '.').at(0) + ".lxfml";
 	auto buffer = Game::assetManager->GetFileAsBuffer(lxfmlPath.c_str());
 
+	if (!buffer.m_Success) {
+		Game::logger->Log("Item", "Failed to load %s", lxfmlPath.c_str());
```
EmosewaMC, line 348: I'd like to imagine we can be more descriptive of this error here and explain _why_ this fail was encountered

dGame/dUtilities/BrickDatabase.cpp
aronwk-aaron: Agreed, but was just following convention how how it was already being handled. Out of scope to change it when this is just to prevent potential crashes

dZoneManager/Zone.cpp
aronwk-aaron: Agreed, this should throw an error and say "Cannot Find File" 
Should also update the triggers message to say the same, but not throw an error

aronwk-aaron:
Addressed the feedback, however, this is a hacky solution. A proper solution is to make the buffer set the failbit set properly when the file doesn't exist, so that `file.good()` will properly work

EmosewaMC (approved):
lgtm

Jettford:
Feel free to merge, I don't have time to fix it properly at the moment

PR #821, opened 2022-11-06
Add player online status to Accounts table

HailStorm32:
Adds `is_online` column to accounts table, and updates said column for an account with a `1` when they logon, and a `0` when they disconnect or log off.

This doesn't drift and isn't affected by server crashes. Requires less computation to determine if a user is online, and is easier to pull from.

Has been running flawlessly for over 10 months on my own server.


EmosewaMC (changes requested):
It would be more advisable to apply these sort of fixes to the core problem which is the activity_log table not being updated at the right times instead of creating a whole new table and running 2x as many queries.

The spot you've added this can instead be used to send a message to ChatServer to mark a player offline forcibly on the off chance they are still online due to other issues.

migrations/dlu/7_isOnline.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE accounts ADD if not exists is_online tinyint(1) NOT NULL DEFAULT 0;
```
EmosewaMC, line 1: While this does compile and run with MariaDB, all our queries are MySQL compatible and this query is not compatible with MySQL.

A fix would be to remove `if not exists` from the query and just allow an error to show up when the migration is run since all migrations are idempotent and an error is fine when someone attempts to run add column migrations twice.

Jettford: There is also a `BOOLEAN` type in MySQL that should be used as this only has 2 states.

HailStorm32: > There is also a BOOLEAN type in MySQL that should be used as this only has 2 states.

`BOOLEAN` is just a synonym for `TINYINT` in MySQL, and are stored the same. 

<br>

> A fix would be to remove if not exists from the query and just allow an error to show up when the migration is run since all migrations are idempotent and an error is fine when someone attempts to run add column migrations twice.

Updated

HailStorm32:
> It would be more advisable to apply these sort of fixes to the core problem which is the activity_log table not being updated at the right times instead of creating a whole new table and running 2x as many queries.

I'm not adding a new table, just a new column to accounts. 

From the standpoint of getting the number of online players, you make less queries with this method than with using activity_log. When using activity_log, you have to make a query for every character in the database. With the way I am proposing, you only make one query to return the count of 1's in the is_online column. 
 
Assuming you update the player counter every minute for a display in a discord channel, and there are 10 total accounts, that's 100 queries in 10 minutes for the activity_log method, and only 10 for my method.

EmosewaMC:
You can also make a case switch statement to get the same information from activity log.  I fail to see how this is any sort of improvement.

EmosewaMC:
We have still decided to close this because the big theta running time is still N in both cases for fetching information from activity_log.  More advanced queries can resolve your issue as opposed to having the server do 2x as much work and adding a new column to maintain.

aronwk-aaron:
This is not a great approach, the data is there in the activity log, just use it. This is what I've done with nexus dash, albeit i can improve my query to make it faster. There is absolutely zero reason to add redundant data to the database since we already have this data stored elsewhere.
Since this would be used by an external program (like nexus dash) and is not being done by the game server, it's up to those programs to write proper queries to make use of the data that already exists.

Issue #822, opened 2022-11-06
BUILD: Docker installation no longer functional


jgkawell:
### Make sure you've done the following:

- [X] I have read the [installation guide](https://github.com/DarkflameUniverse/DarkflameServer/blob/main/README.md).

### DarkflameServer Version

9c58ea5c415bf64d7972fc5408d8ef4e554185e9

### Platform

Other

### Architecture

x86

### Error Logs

```log
Creating symlinks for client files
Creating symlinks for config files
Starting MasterServer
Fatal error -1
Stacktrace:
[00] ./MasterServer(_Z14CatchUnhandledi+0x42) [0x44d882]
[01] /usr/local/lib64/libstdc++.so.6(+0xb107a) [0x7f78e7f7e07a]
[02] /usr/local/lib64/libstdc++.so.6(+0xb10e5) [0x7f78e7f7e0e5]
[03] /usr/local/lib64/libstdc++.so.6(+0xb1338) [0x7f78e7f7e338]
[04] /usr/local/lib64/libstdc++.so.6(_ZSt24__throw_invalid_argumentPKc+0x39) [0x7f78e7f75171]
[05] ./MasterServer(_ZN9__gnu_cxx6__stoaIlicJiEEET0_PFT_PKT1_PPS3_DpT2_EPKcS5_PmS9_+0x93) [0x44b333]
[06] ./MasterServer(main+0x1a6) [0x442dc6]
[07] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xea) [0x7f78e7bd5d0a]
[08] ./MasterServer(_start+0x2a) [0x44510a]
```

jgkawell:
This is exclusively when running in Docker. I found the root of the issue being the change to the way the server recognizes the packed/unpacked client type and the new shared config file. I already have a fix for this and will have a PR up shortly. Wanted to have an issue up though so others with the issue can Google the log and find it.

PR #802, opened 2022-10-31
Add support for packed clients

Jettford:
- Implements pack file interface
- Automatically extracts cdclient for later conversion
- Implements pack file istream buffer so file reading can be done as normal
- Navmeshes were moved to the server root so they can now be automatically extracted

PR #823, opened 2022-11-06
Fixed Docker installation for new shared config and client type

jgkawell:
This resolves: #822

In PR #795, a shared config file was implemented to reduce duplicated values. This broke the Docker installation as it's configuration scripts weren't updated to use this new shared config. This PR updates the Docker setup scripts to make sure the new shared config is added and configured.

In PR #802, the server was updated to recognize the packed/unpacked type of client. This also broke the Docker installation as it doesn't have the required files for the recognition to work. This PR updates the Docker setup script to create a fake database file so the server recognizes the client as unpacked (the Docker installation always unpacks the client anyway).


docker/start_server.sh
```diff
@@ -8,6 +8,8 @@ function symlink_client_files() {
     ln -s /client/client/res/names/ /app/res/names
     ln -s /client/client/res/CDServer.sqlite /app/res/CDServer.sqlite
     ln -s /client/client/locale/locale.xml /app/locale/locale.xml
+    # need to create this file so the server knows the client is unpacked (see `dCommon/dClient/AssetManager.cpp`)
```
aronwk-aaron, line 11: why? this forces docker to only used unpacked client assets, when it should be able to use a packed client with no issues.

docker/start_server.sh
```diff
@@ -8,6 +8,8 @@ function symlink_client_files() {
...
+    touch /app/res/cdclient.fdb
```
EmosewaMC, line 12: Does cdclient.fdb get deleted after the conversion process?  Not sure why this is needed if the file is needed by the client so some clarification here would be great

EmosewaMC (approved):
1 quick question

aronwk-aaron (commented):
Just want to note, that there will be a few more additions to the server coming in the future which will largely the ways docker setup does things redundant and outdated.
I plan on ripping out the current docker setup and redoing it once the server changes are implemented, just due to the fact that a lot of it will be able to be simplified and streamlined instead of needing to sue external tools like lcdr-utils

docker/start_server.sh
jgkawell: It's still there, but it's just not linked to the location the server will run from. I tried linking it but it messed up and tried importing the db a second time. Didn't bother diving into why so I just created a basic file since the server just checks for existence.

docker/start_server.sh
jgkawell: Haven't done much with unpacking in a while, but if I remember correctly [this script](https://github.com/DarkflameUniverse/DarkflameServer/blob/main/docker/setup.sh) is unpacking the client if it hasn't been already. So from the server's perspective in Docker, the client will always be unpacked.

aronwk-aaron (approved):

PR #825, opened 2022-11-07
Fix server type advertisement in handshake with custom auth port

luxaritas:
Tested by setting auth port to 2001 and using nginx to redirect port 2001 to 1001. Successfully authenticated and loaded char.

### DIFF
--- a/dNet/AuthPackets.cpp
+++ b/dNet/AuthPackets.cpp
@@ -29,16 +29,16 @@ void AuthPackets::HandleHandshake(dServer* server, Packet* packet) {
 	inStream.Read(clientVersion);
 
 	server->GetLogger()->Log("AuthPackets", "Received client version: %i", clientVersion);
-	SendHandshake(server, packet->systemAddress, server->GetIP(), server->GetPort());
+	SendHandshake(server, packet->systemAddress, server->GetIP(), server->GetPort(), server->GetServerType());
 }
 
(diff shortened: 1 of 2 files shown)


Jettford (dismissed):

dNet/AuthPackets.cpp
```diff
@@ -29,16 +29,16 @@ void AuthPackets::HandleHandshake(dServer* server, Packet* packet) {
 	inStream.Read(clientVersion);
 
 	server->GetLogger()->Log("AuthPackets", "Received client version: %i", clientVersion);
-	SendHandshake(server, packet->systemAddress, server->GetIP(), server->GetPort());
+	SendHandshake(server, packet->systemAddress, server->GetIP(), server->GetPort(), server->GetServerType());
 }
 
-void AuthPackets::SendHandshake(dServer* server, const SystemAddress& sysAddr, const std::string& nextServerIP, uint16_t nextServerPort) {
+void AuthPackets::SendHandshake(dServer* server, const SystemAddress& sysAddr, const std::string& nextServerIP, uint16_t nextServerPort, const ServerType serverType) {
```
EmosewaMC, line 35: We could pass serverType by reference here instead of copying it over - just a small nit

EmosewaMC (approved):

Jettford (approved):

PR #826, opened 2022-11-07
cleanup and define all unknowns in zone reading

aronwk-aaron:
fix for movement path config reading
Tested that this has no functional changes. There may can be changes made since this data is more clearly defined.
This is also in prep for npc movement pathing
reference: https://github.com/lcdr/lu_formats/blob/master/files/luz.ksy

### DIFF
--- a/dZoneManager/Zone.cpp
+++ b/dZoneManager/Zone.cpp
@@ -358,11 +358,8 @@ SceneTransitionInfo Zone::LoadSceneTransitionInfo(std::istream& file) {
 void Zone::LoadPath(std::istream& file) {
 	Path path = Path();
 
-	uint32_t flags;
-	uint32_t pathType;
-	uint32_t pathBehavior;
-
 	BinaryIO::BinaryRead(file, path.pathVersion);
+
 	uint8_t stringLength;
 	BinaryIO::BinaryRead(file, stringLength);
 	for (uint8_t i = 0; i < stringLength; ++i) {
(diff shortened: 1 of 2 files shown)


dZoneManager/Zone.cpp
```diff
@@ -403,10 +389,17 @@ void Zone::LoadPath(std::istream& file) {
 			}
 		}
 	} else if (path.pathType == PathType::Property) {
-		int32_t unknown;
-		BinaryIO::BinaryRead(file, unknown);
+
+		int32_t propertyPathType;
+		BinaryIO::BinaryRead(file, propertyPathType);
+		path.property.pathType = PropertyPathType(propertyPathType);
```
EmosewaMC, line 395: I believe we could shorten this to BinaryIO::BinaryRead(file, path.property.pathType);
Does this work here?

dZoneManager/Zone.cpp
```diff
@@ -403,10 +389,17 @@ void Zone::LoadPath(std::istream& file) {
...
+
 		BinaryIO::BinaryRead(file, path.property.price);
-		BinaryIO::BinaryRead(file, path.property.rentalTime);
+
+		int32_t rentalTime;
+		BinaryIO::BinaryRead(file, rentalTime);
+		path.property.rentalTime = PropertyRentalTimeUnit(rentalTime);
```
EmosewaMC, line 401: Same as above

EmosewaMC (commented):
Couple questions so far

dZoneManager/Zone.cpp
aronwk-aaron: ahhh you're correct, was under the impression that due to the enum class that you have to declare it like this

EmosewaMC (approved):
small small comment

EmosewaMC (approved):

Jettford (approved):

PR #827, opened 2022-11-07
Use Property_Path name and description when claiming a property

aronwk-aaron:
Tested claiming props and they now have the correct default values for name and description that they would have in live


EmosewaMC (commented):
Does this address the issue of `small preoperty map` and `testmap 1451`?

aronwk-aaron:
> Does this address the issue of `small preoperty map` and `testmap 1451`?

yes, I checked against jamie' videos and those values contain the proper data.
it also contain the property path that we are getting from the DB, so we could technically eliminate a DB query.
And rental cost info is in there too, as well as the rent period.
May see about adding this stuff later but wanted to keep this PR a small scope

EmosewaMC (approved):
Let's try to remove that unused query you mentioned next since you said its unneeded.   lgtm

Jettford (approved):

PR #828, opened 2022-11-09
use path from zone instead of querying the db

aronwk-aaron:
Checked that all properties still load and that the data for the paths are the same


EmosewaMC (approved):

aronwk-aaron:
closing since I was testing and playing with this more I discovered it breaks bbb

aronwk-aaron:
Closing this PR as it's not possible to get the correct height (y) for the path to be at without nav meshes for all properties. Which all the property path waypoints have incorrect y values.

PR #829, opened 2022-11-09
Prevent adding movingplatform components to all entites with an attached_path

aronwk-aaron:
This is preparation for npc pathing. 
All entities that have an attached_path should not have moving platform components as they do now, so only add a moving platform comp if the path type is for a moving platform

### DIFF
--- a/dGame/Entity.cpp
+++ b/dGame/Entity.cpp
@@ -699,10 +699,10 @@ void Entity::Initialize() {
 		// else if we are a movement path
 		} else if (path->pathType == PathType::Movement) {
 			auto movementAIcomp = GetComponent<MovementAIComponent>();
-			if (!movementAIcomp){
-				// TODO: create movementAIcomp and set path
-			} else {
+			if (movementAIcomp){
 				// TODO: set path in existing movementAIComp
+			} else {
+				// TODO: create movementAIcomp and set path
 			}
 		}
 	}
--- a/dGame/Entity.cpp
+++ b/dGame/Entity.cpp
@@ -690,21 +690,21 @@ void Entity::Initialize() {
 	std::string pathName = GetVarAsString(u"attached_path");
 	const Path* path = dZoneManager::Instance()->GetZone()->GetPath(pathName);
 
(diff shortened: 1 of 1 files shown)


EmosewaMC:
How has this been tested?

dGame/Entity.cpp
```diff
@@ -696,6 +687,26 @@ void Entity::Initialize() {
 		m_Components.insert(std::make_pair(COMPONENT_TYPE_MOVEMENT_AI, new MovementAIComponent(this, moveInfo)));
 	}
 
+	std::string pathName = GetVarAsString(u"attached_path");
+	const Path* path = dZoneManager::Instance()->GetZone()->GetPath(pathName);
+
+	//Check to see if we have an attached path and assing the appropiate component to handle it:
```
EmosewaMC, line 693: small typo here :D.  `assing` vs `assign`

dGame/Entity.cpp
```diff
@@ -696,6 +687,26 @@ void Entity::Initialize() {
...
+			MovingPlatformComponent* plat = new MovingPlatformComponent(this, pathName);
+			m_Components.insert(std::make_pair(COMPONENT_TYPE_MOVING_PLATFORM, plat));
+		// else if we are a movement path
+		} else if (path->pathType == PathType::Movement) {
+			auto movementAIcomp = GetComponent<MovementAIComponent>();
+			if (movementAIcomp){
+				// TODO: set path in existing movementAIComp
+			} else {
+				// TODO: create movementAIcomp and set path
+			}
```
EmosewaMC, line 706: Since this code path is not used for now and is optimized away, what would be your thoughts on commenting it out for now and un-commenting it when this is implemented?

EmosewaMC (approved):
Couple questions

dGame/Entity.cpp
aronwk-aaron: sounds good

EmosewaMC (approved):

Jettford (approved):

PR #830, opened 2022-11-10
Implement undo action for pre-built models

EmosewaMC:
Brick building as of right now does not implement the undo action properly.  This commit addresses the issue with undoing button being non-functional server side and implements the GM needed for addressing further issues.

Implement GameMessage UnUseModel which is called when a model in BrickBuilding is UnUsed.  Important for UGC content down the line.  Final code has been tested as follows:
1. Placed a model in brick build
2. saved placed a brick
3. repeat 2 and 3 twice more for 6 total models
4. Place a new model in brick mode and then edit all 7 models into one brick model instance
5. Pressing undo returns the converted model to the inventory and properly discards the other 6 without crashing.  Intended live behavior is to store this in the inventory instead however behind the scenes work is needed to implement UGC models properly.
The undo action works for UGC models however that functionality I will leave unimplemented for now.


EmosewaMC:
A new unit test is attached for a now testable GameMessage

dCommon/eBlueprintSaveResponseType.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
```
aronwk-aaron, line 1: should this go ahead and be moved to the enum dir even though that PR isn't mergd?

aronwk-aaron (approved):
non-blocking question

dCommon/eBlueprintSaveResponseType.h
EmosewaMC: Personally I would say no since it would be more efficient to not have merge conflicts but I am leaning more towards taking care of it in that PR

dGame/dGameMessages/GameMessages.cpp
```diff
@@ -2353,16 +2380,40 @@ void GameMessages::HandleDeletePropertyModel(RakNet::BitStream* inStream, Entity
...
+	if (inventoryComponent) {
+		auto* inventory = inventoryComponent->GetInventory(eInventoryType::MODELS);
+		auto* itemToMove = inventory->FindItemById(previousItemID);
+
+		if (itemToMove) {
+			LOT previousLot = itemToMove->GetLot();
+			inventoryComponent->MoveItemToInventory(itemToMove, eInventoryType::MODELS_IN_BBB, 1, false);
+
+			auto* destinationInventory = inventoryComponent->GetInventory(eInventoryType::MODELS_IN_BBB);
+			if (destinationInventory) {
```
EmosewaMC, line 2398: If this is merged before names, this needs to be updated to look for an item by the previous subkey first, then by lot.

Wincent01 (approved):

PR #831, opened 2022-11-12
Quiet logs of non-existent lutriggers files unable to be loaded

aronwk-aaron:
Not every scene has triggers, that is normal
Quell the onslaught of meaningless logs
Tested that loading zones still works and triggers that do exist get loaded


EmosewaMC (approved):
We should have a more descriptive name for this PR so when it is merged we know what's changed.  Non-blocking

Jettford (approved):

PR #842, opened 2022-11-21
Add support to name and describe models

EmosewaMC:
Adds support to name and describe models.
Tested that User Generated Content models can be named and described and that the property does not get its name overwritten anymore. Regular models maintain their name and description across instances. This does not allow the name to persist when the model is picked up as this is a larger implementation. Tested that taming pets and spawning pets does not cause crashes or issues.

EntityManager changes are because the client expects the objectID of the User Generated Content model to have those flags set.

BaseReplicaData changes are because the client expects the name to be set in the construction of the entity.
Tested that if the requested name is not in the whitelist, the name reverts to its previous value, same with the description.

### DIFF
--- a/dGame/dGameMessages/GameMessages.cpp
+++ b/dGame/dGameMessages/GameMessages.cpp
@@ -28,6 +28,7 @@
 #include "GameConfig.h"
 #include "RocketLaunchLupComponent.h"
 #include "eUnequippableActiveType.h"
+#include "eModerationStatus.h"
 
 #include <sstream>
 #include <future>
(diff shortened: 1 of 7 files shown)


aronwk-aaron (dismissed):

dGame/dGameMessages/GameMessages.cpp
```diff
@@ -2167,31 +2170,41 @@ void GameMessages::HandleUpdatePropertyOrModelForFilterCheck(RakNet::BitStream*
 		inStream->Read(character);
 		name.push_back(character);
 	}
+	const std::string descriptionAsString = GeneralUtils::UTF16ToWTF8(description);
+	const std::string nameAsString = GeneralUtils::UTF16ToWTF8(name);
+	Game::logger->Log("GameMessages", "%llu %llu ", worldId, ugcId);
```
EmosewaMC, line 2175: remove logger

EmosewaMC:
Closing as this will be implemented alongside another crucial fix in #850

PR #843, opened 2022-11-21
Fix Avant Gardens Survival

EmosewaMC:
This addresses the Avant Gardens Survival bug and allows users to no longer need to edit their clients in order to play the minigame

Update README to reflect the change


EmosewaMC:
These changes do not conflict with users who have the patch, I have just tested that having the previous fix applied did not cause issues

Wincent01 (approved):

aronwk-aaron (approved):

PR #846, opened 2022-11-23
Add property Teleport behavior

EmosewaMC:
Will mark pr as ready for review when this has been tested


aronwk-aaron (commented):
~~Small note, this is supposed to play the LUP teleport animation where the door opens and the character walks through it~~
Ignore me, that's handled by the skill

aronwk-aaron (approved):

Jettford (approved):

PR #848, opened 2022-11-23
Make brick stacks live accurate

aronwk-aaron:
Increase brick stack size and don't make new stacks of bricks
Tested by adding bricks to the inventory and giving uint_max of a brick several times and only one stack was created


aronwk-aaron:
dissembling models go through this codepath to add their bricks to the inventory, so that is handles here as well

EmosewaMC (approved):

Wincent01 (approved):

PR #850, opened 2022-11-25
Add Model names and descriptions, save them to the inventory correctly

EmosewaMC:
This PR addresses 2 issues (yes i know two PRs in one but these issues are tied together)

The first half:
Adds support to name and describe models.
Tested that User Generated Content models can be named and described and that the property does not get its name overwritten anymore. Regular models maintain their name and description across instances. This does not allow the name to persist when the model is picked up as this is a larger implementation. Tested that taming pets and spawning pets does not cause crashes or issues.

EntityManager changes are because the client expects the objectID of the User Generated Content model to have those flags set.

BaseReplicaData changes are because the client expects the name to be set in the construction of the entity.
Tested that if the requested name is not in the whitelist, the name reverts to its previous value, same with the description.

Tested moving various items to and from the vault, all model related items moved without issue (but will be doing more extensive testing in a bit), will also test bound and unbound items and ensure they interact just fine moving to and from the vault.


EmosewaMC:
Split out the vault fixes into their own PR.  Will remove it from this one when that is merged.

EmosewaMC:
Closing due to being stale and having new knowledge

PR #851, opened 2022-11-25
Fix three-part names coming across as INVALID with custom client path

luxaritas:
This is due to AssetManager not being used, so the character names were being pulled from build/res, which may not exist.

Tested by running the server with a custom client path and creating a character, which successfully set the three-part name instead of setting it as INVALID

This is split out from #834 as it is a distinct bug fix critical for servers with a custom client path

### DIFF
--- a/dGame/UserManager.cpp
+++ b/dGame/UserManager.cpp
@@ -20,6 +20,7 @@
 #include "Entity.h"
 #include "EntityManager.h"
 #include "SkillComponent.h"
+#include "AssetManager.h"
 
 UserManager* UserManager::m_Address = nullptr;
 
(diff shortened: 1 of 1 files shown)


EmosewaMC:
Could you also test that chat whitelist works too please?

luxaritas:
Yep, works as expected

luxaritas:
Also please confirm that GetFIleAsBuffer and then converting to an istream is the correct approach here vs just constructing the paths as they were before but using `Game::assetManager->GetResPath()`

EmosewaMC (changes requested):
Yes you are correct, however I believe bounds checking similar to how we do it [here](https://github.com/DarkflameUniverse/DarkflameServer/blob/b17ba56af17f42f47dafdfbe56913caabf49c450/dZoneManager/Zone.cpp#L46) so if the file cannot be loaded for some reason, an error is logged and the server is unable to start.  This would also allow a very traceable error for if an issue happens with those files!

luxaritas:
Confirmed that if a relevant file is not available, the message is logged and the server raises an exception

EmosewaMC (approved):
Thanks for this!

Wincent01 (approved):

Issue #194, opened 2021-12-12
BUG: Xpresso Speedbuff retained after dying


Mezalion:
### 

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### Issue Description

When dying short before the XPResso Speedbuff expires the Buff is kept after death. Not transfered when changing worlds tho.

### Reproduction steps

Use Xpresso
Die shortly before the Timer Ends
Respawn -> Eternal XPresso Speed

### Expected Behavior

It should delete the Buff on Dying

### Environment

VPS Ubuntu 20.something 
LCDRs Unpacked Client

Mezalion:
Newly Discovered: After every death the Speed slightly increases.

averysumner:
Possibly related to #235.

ianRudnick:
This also seems to happen when using other speed boosts, like Daredevil flare guns skill 3. It also works in Nexus Tower: using the Paradox warps has the same effect as dying on increasing speed.

EmosewaMC:
fixed by #856 from my testing.  Let us know if the issue persists and the reproduction case.

PR #856, opened 2022-11-28
Add speedbase readling and writing to the level prograssion component and impli proper character versions for fixes

aronwk-aaron:
This implements using the speedbase in the charxml to properly update the speed multiplier
This also implements char XML versioning. Since we have some and are going to have more charxml fixes, we need to version them so that we don't pointlessly apply fixes on every char load.


Wincent01:
There at least were faults where a skeleton on Crux Prime could give you an enormous permanent speed-boost until the next session, make sure those are not preserved.

aronwk-aaron:
yeah, there are a few issues with speed boosts, I may fix them here if it's not too much code change. otherwise, I may look into them in another PR. I know I also need to fix speed multiplier not being able to be slower than 1, from previous conversations that you and I have had

EmosewaMC:
Personally not a fan of the if level greater than 20 thing.  This should be a 1 time change to base speed on a level up since it's an adjustment to base speed and nothing more

aronwk-aaron:
Tested that speedboosts work the same as they have before
Tested that changing the character version doesn't affect anything and upgradeing charxml's work properly and it's only done once

This does not fix speed boosts stacking permanently, that's a larger issue with the behavior system

dGame/dBehaviors/SpeedBehavior.cpp
```diff
@@ -7,85 +7,47 @@
...
 	const auto current = controllablePhysicsComponent->GetSpeedMultiplier();
-
 	controllablePhysicsComponent->SetSpeedMultiplier(current + ((m_RunSpeed - 500.0f) / 500.0f));
-
 	EntityManager::Instance()->SerializeEntity(target);
 
 	if (branch.duration > 0.0f) {
 		context->RegisterTimerBehavior(this, branch);
 	} else if (branch.start > 0) {
-		controllablePhysicsComponent->SetIgnoreMultipliers(true);
```
EmosewaMC, line 35: Out of curiosity, why was this removed?

dGame/dBehaviors/SpeedBehavior.cpp
```diff
@@ -7,85 +7,47 @@
...
 
+void SpeedBehavior::Timer(BehaviorContext* context, BehaviorBranchContext branch, LWOOBJID second) {
+	End(context, branch, second);
+}
+
 void SpeedBehavior::Load() {
 	m_RunSpeed = GetFloat("run_speed");
-
-	if (m_RunSpeed < 500.0f) {
-		m_RunSpeed = 500.0f;
```
EmosewaMC, line 87: Any particular reason this was removed?

dGame/dBehaviors/SpeedBehavior.cpp
```diff
@@ -7,85 +7,47 @@
...
 	auto* controllablePhysicsComponent = target->GetComponent<ControllablePhysicsComponent>();
-
-	if (controllablePhysicsComponent == nullptr) {
-		return;
-	}
+	if (!controllablePhysicsComponent) return;
 
 	const auto current = controllablePhysicsComponent->GetSpeedMultiplier();
-
 	controllablePhysicsComponent->SetIgnoreMultipliers(false);
```
EmosewaMC, line 41: This line is not done in the Timer function, have you made sure this doesn't cause any issues setting this when it was not in the past?

dGame/dBehaviors/SpeedBehavior.h
```diff
@@ -13,6 +13,8 @@ class SpeedBehavior final : public Behavior
 
 	void Handle(BehaviorContext* context, RakNet::BitStream* bitStream, BehaviorBranchContext branch) override;
 
+	void Calculate(BehaviorContext* context, RakNet::BitStream* bitStream, BehaviorBranchContext branch) override;
```
EmosewaMC, line 16: How was Calculate tested for functionality?

dGame/dComponents/LevelProgressionComponent.cpp
```diff
@@ -7,6 +7,8 @@
 LevelProgressionComponent::LevelProgressionComponent(Entity* parent) : Component(parent) {
 	m_Parent = parent;
 	m_Level = 1;
+	m_SpeedBase = 500.0f;
+	m_CharacterVersion = 1;
```
EmosewaMC, line 11: What are your thoughts on making an enum for character versions so that instead of these being arbitrary values, the versions would have some kind of value associated with them?

dWorldServer/WorldServer.cpp
```diff
@@ -1002,9 +1003,18 @@ void HandlePacket(Packet* packet) {
...
 
-				c->SetRetroactiveFlags();
-
-				player->RetroactiveVaultSize();
+				// Do charxml fixes here
+				auto* levelComponent = player->GetComponent<LevelProgressionComponent>();
+				if (!levelComponent) return;
+
+				if (levelComponent->GetCharacterVersion() == 1) {
+					Game::logger->Log("WorldServer", "Upgrading Character version to 2");
```
EmosewaMC, line 1011: What are your thoughts on adding some more information here so we know what character we are upgrading the version of?

EmosewaMC (commented):
Just a few questions and curiosities

dGame/dBehaviors/SpeedBehavior.cpp
aronwk-aaron: To allow slower speeds, which had multiple checks to make it not happen, even though it should be possible. this was originally because the wormholer animations would break, but that doesn't happen anymore.
there is one check for slower speeds left in the controllable physics component, this is due the infinite speed glitch that is present, since the infinite can be both positive and negative

dGame/dBehaviors/SpeedBehavior.h
aronwk-aaron: this was tested with the pirate parrot crash script, which is what originally caused me to look into this since it's supposed to slow the player down when sneaking past the parrot guards to the pirate treasure. and then speed the player back up when exiting that area.

dGame/dBehaviors/SpeedBehavior.cpp
aronwk-aaron: It's just a fail safe to enable the update in serialization, probably not needed in the first place, but it was there. This is the only place in the codebase that `SetIgnoreMultipliers` is used so it shouldn't be being blocked by anything else outside of the controllable physics serialization checks.

dGame/dComponents/LevelProgressionComponent.cpp
aronwk-aaron: That sounds good, can discuss this more off GH

EmosewaMC (approved):

dGame/dBehaviors/SpeedBehavior.cpp
aronwk-aaron: this was the case. It stopped things from working when they should work

aronwk-aaron:
Retested speed boosts again.
The changes to the speed behavior now prevent it from stacking infinitely.
However, there is still a bug (or rather unimplemented features) around being interrupted while attacking that can cause a speed boost to not be removed cause the end does not trigger.

EmosewaMC (dismissed):

aronwk-aaron:
fixes #194

aronwk-aaron:
> There at least were faults where a skeleton on Crux Prime could give you an enormous permanent speed-boost until the next session, make sure those are not preserved.

This is also fixed now

Jettford (approved):

EmosewaMC (approved):

PR #862, opened 2022-12-01
Fix crash in BasicAttackBehavior

### LINKED ISSUE #430: BUG: Ice Shurikens crashes WorldServer

### Issue Description

I found a bug and I don't know if it is just using the if ninja ice shurikens or others as well, but it causes the client to crash back to the login screen when throwing and hitting either nothing or the landscape not smashables or enemies and when you sign back in it reverts you to a previous location and doesn't save the progress.

### Reproduction steps

Equip Ice Shurikens and throw hit landscape or miss

### Expected Behavior

For the client to not crash and act as a miss like other ranged weapons

### Environment

WSL

EmosewaMC:
Address the BasicAttackBehavior reading out of bounds memory and reading bits that didnt exist, which occasionally caused crashes and also caused the behavior to do undefined behavior due to the bad reads.

Tested that attacking a wall anywhere with a projectile now does not crash the game.  Tested with logs that the behavior correctly returned when there were no allocated bits or returned when other unserialized states were not met.

No observed issues from about 10 minutes of attacking things, both grouped and not grouped.

Fixes #430

### DIFF
--- a/dGame/dBehaviors/BasicAttackBehavior.cpp
+++ b/dGame/dBehaviors/BasicAttackBehavior.cpp
@@ -14,43 +14,65 @@ void BasicAttackBehavior::Handle(BehaviorContext* context, RakNet::BitStream* bi
 		auto* destroyableComponent = entity->GetComponent<DestroyableComponent>();
 		if (destroyableComponent != nullptr) {
 			PlayFx(u"onhit", entity->GetObjectID());
-			destroyableComponent->Damage(this->m_maxDamage, context->originator, context->skillID);
+			destroyableComponent->Damage(this->m_MaxDamage, context->originator, context->skillID);
 		}
 
-		this->m_onSuccess->Handle(context, bitStream, branch);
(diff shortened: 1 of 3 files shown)


aronwk-aaron (approved):
LGTM!

Wincent01 (approved):
Looks good. Would be great to have these kinds of checks in more places. I see you've removed the commented out code for homing projectiles, that implementation doesn't work correctly, but it is something which is missing, just to note.

EmosewaMC:
> Looks good. Would be great to have these kinds of checks in more places. I see you've removed the commented out code for homing projectiles, that implementation doesn't work correctly, but it is something which is missing, just to note.

Ah interesting!  Ill add a TODO there so it is clear there is work to be done there for homing projectiles.  I saw the large commented out code block and just decided to bin it since that code was commented out due to either not working or being a temp fix.

aronwk-aaron (approved):

Wincent01 (approved):

PR #863, opened 2022-12-01
Add bandwidth limit

EmosewaMC:
I havent tested this yet but @DarwinAnim8or said just uncomment it so yeah ill test in morning unless someone else does

### DIFF
--- a/dNet/dServer.cpp
+++ b/dNet/dServer.cpp
@@ -181,7 +181,7 @@ bool dServer::Startup() {
 	if (mIsInternal) {
 		mPeer->SetIncomingPassword("3.25 DARKFLAME1", 15);
 	} else {
-		//mPeer->SetPerConnectionOutgoingBandwidthLimit(800000); //100Kb/s
+		mPeer->SetPerConnectionOutgoingBandwidthLimit(800000); //100Kb/s
 		mPeer->SetIncomingPassword("3.25 ND1", 8);
 	}
 
--- a/dNet/dServer.cpp
+++ b/dNet/dServer.cpp
@@ -2,6 +2,7 @@
 #include "dServer.h"
 #include "dNetCommon.h"
 #include "dLogger.h"
+#include "dConfig.h"
 
 #include "RakNetworkFactory.h"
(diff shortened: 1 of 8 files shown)


Wincent01:
Maybe make this configurable and allow for it to be turned off.

EmosewaMC:
> Maybe make this configurable and allow for it to be turned off.

Do you know what issues it causes having this limit be higher than it was in live, if it caused any issues to begin with?

dNet/dServer.cpp
```diff
@@ -191,6 +193,11 @@ bool dServer::Startup() {
 	return true;
 }
 
+void dServer::UpdateBandwidthLimit() {
+	auto newBandwidth = mConfig->GetValue("maximum_outgoing_bandwidth");
+	mPeer->SetPerConnectionOutgoingBandwidthLimit(!newBandwidth.empty() ? std::stoi(newBandwidth) : 0);
```
aronwk-aaron, line 198: Does 0 mean unlimited?

aronwk-aaron (dismissed):

dNet/dServer.cpp
EmosewaMC: yes - 0 is unlimited by default as per RakNet documentation

aronwk-aaron: cool beans

aronwk-aaron (dismissed):

DarwinAnim8or (commented):
Code looks fine, but the limit was actually 10kbps, not 100 like it was set in the commented out code. 

I had increased it tenfold to deal with the same Replica stuff I had mentioned before, but that didn't work so I just disabled it all together.

EmosewaMC:
> Code looks fine, but the limit was actually 10kbps, not 100 like it was set in the commented out code.
> 
> I had increased it tenfold to deal with the same Replica stuff I had mentioned before, but that didn't work so I just disabled it all together.

Interesting!  I'll make that change

DarwinAnim8or (approved):

aronwk-aaron (approved):

PR #865, opened 2022-12-02
Add Delete Inventory Slash Command

EmosewaMC:
Tested that the specified inventory has all its items deleted.
Tested that specifying an invalid inventory causes the command to output a chat message.

Enum changes in eInventoryType are only name changes, no inventory type numbers have been altered.  I have simply completed the list of inventory types so that in the future, the others can be used for reflection

Unsure if this method of reflection is good, this is just the way I thought to do this as.


dCommon/dEnums/eInventoryType.h
```diff
@@ -0,0 +1,59 @@
...
+	ITEMS = 0,
+	VAULT_ITEMS,
+	BRICKS,
+	MODELS_IN_BBB,
+	TEMP_ITEMS,
+	MODELS,
+	TEMP_MODELS,
+	BEHAVIORS,
+	PROPERTY_DEEDS,
+	BRICKINBBB,
```
aronwk-aaron, line 21: `BRICKS_IN_BBB` to match the above `MODELS_IN_BBB`

dGame/Character.cpp
```diff
@@ -15,6 +15,7 @@
 #include "Zone.h"
 #include "ChatPackets.h"
 #include "Inventory.h"
+#include "InventoryComponent.h"
```
aronwk-aaron, line 18: why are there a bunch of header additions of InventoryComponent and CDClientManager?

aronwk-aaron (changes requested):
One change request and a non-blocking question.

dGame/Character.cpp
EmosewaMC: Because I removed the InventoryComponent include in GameMessages.h to reduce the number of pre-compiled headers.  This caused several dozen files to no longer compile since they were no longer included anymore.

dCommon/dEnums/eInventoryType.h
EmosewaMC: sounds good

aronwk-aaron (dismissed):
LGTM

aronwk-aaron (dismissed):

DarwinAnim8or (approved):

aronwk-aaron (approved):

Issue #866, opened 2022-12-03
ENH: CDServer migration improvements


EmosewaMC:
### Is your feature request related to a problem?

Store the migration history for the CDServer in the CDServer itself so if it is replaced, migrations are run again for that cdserver.

Also run all migrations in order.

### Describe the solution you'd like

above

### Repository breaking implications

none

### Describe alternatives you've considered

a flag to always run migrations ig

PR #867, opened 2022-12-03
Move CDServer migration history table

Wincent01:
This pull request seeks to add functionality to move the cdserver migration history from the main database to the cdserver database.

Has been tested and is backwards compatible with the current system. The server will copy over cdserver migrations history table entries from the main database to the cdserver database, and any new migrations will go into the cdserver database.

This will make it easier to switch out the cdserver database for the purpose of modding and alike.

Fixes:  #866


EmosewaMC (approved):
lgtm - only nit is using int instead of int32_t but for an int im not sure how much difference there is there

EmosewaMC (approved):

aronwk-aaron (approved):

PR #871, opened 2022-12-05
Add change idle flags behavior and GM

aronwk-aaron:
Tested that this triggers appropriately when using things that have this behavior.
Weapons include (but are not limited to) the charge up attacks for the wormholer and the spinjutsu weapons


dGame/dGameMessages/GameMessages.h
```diff
@@ -373,6 +374,15 @@ namespace GameMessages {
...
 	void SendDisplayChatBubble(LWOOBJID objectId, const std::u16string& text, const SystemAddress& sysAddr);
 
+	/**
+	 * @brief
+	 *
+	 * @param objectId ID of the entity to set idle flags
+	 * @param FlagsOn Flag to turn on
+	 * @param FlagsOff Flag to turn off
+	 */
+	void SendChangeIdleFlags(LWOOBJID objectId, eAnimationFlags FlagsOn, eAnimationFlags FlagsOff, const SystemAddress& sysAddr);
```
EmosewaMC, line 384: These default to zero for what its worth per [the struct](https://lcdruniverse.org/lu_packets/src/lu_packets/world/gm/client/mod.rs.html#1363-1368)

EmosewaMC (approved):
How does this interact if you have multiple idle flags active at once?  Is the behavior as expected?

dGame/dGameMessages/GameMessages.h
aronwk-aaron: Ty will make it default to 0

aronwk-aaron:
> How does this interact if you have multiple idle flags active at once?  Is the behavior as expected?

There's not any combination of skills that allow it to be applied multiple at the same time. This is exclusively used on two handed weapons as far as I can tell. I'll double check that later though

EmosewaMC (approved):

Wincent01 (approved):

PR #872, opened 2022-12-05
Implement FDB to SQLite

EmosewaMC:
Implement a class to convert an FDB file to SQLite, removing any need for a python dependency for setting up and running a server (the python dependency removal will be in #806)

Tested that from the build directory and from 1 directory up from build, that the server starts, and converts the fdb file in the res path to CDServer.sqlite and properly applies the migrations.

ran a `SELECT * FROM BehaviorParameter query;` with LCDR's tool and the new converter and there was no difference between either of two with the table results

Tested that migrations ran as expected and I was able to load into the game and check my mail (requires migrations to be run to succeed)

### DIFF
--- a/dCommon/FdbToSqlite.cpp
+++ b/dCommon/FdbToSqlite.cpp
@@ -1,0 +1,248 @@
+#include "FdbToSqlite.h"
+
+#include <map>
+#include <fstream>
+#include <cassert>
+#include <iomanip>
+
+#include "BinaryIO.h"
+#include "CDClientDatabase.h"
+#include "GeneralUtils.h"
+#include "Game.h"
+#include "dLogger.h"
+
+#include "eSqliteDataType.h"
+
+std::map<eSqliteDataType, std::string> FdbToSqlite::Convert::sqliteType = {
+			{ eSqliteDataType::none, "none"},
+			{ eSqliteDataType::int32, "int32"},
+			{ eSqliteDataType::real, "real"},
+			{ eSqliteDataType::text_4, "text_4"},
+			{ eSqliteDataType::int_bool, "int_bool"},
+			{ eSqliteDataType::int64, "int64"},
+			{ eSqliteDataType::text_8, "text_8"}
+};
+
+FdbToSqlite::Convert::Convert(std::string basePath) {
+	this->basePath = basePath;
+}
+
+bool FdbToSqlite::Convert::ConvertDatabase() {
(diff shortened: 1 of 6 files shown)


EmosewaMC:
This also works on windows after further testing

aronwk-aaron (approved):

Jettford (approved):

dCommon/FdbToSqlite.cpp
```diff
@@ -0,0 +1,248 @@
...
+			size_t position = 0;
+
+			// Need to escape quote with a double of ".
+			while (position < stringValue.size()) {
+				if (stringValue.at(position) == '\"') {
+					stringValue.insert(position, "\"");
+					position++;
+				}
+				position++;
+			}
```
Xiphoseer, line 221: should use the sqlite functions for doing escaping, imho

dCommon/FdbToSqlite.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+	int32_t emptyValue{};
+	int32_t intValue{};
+	float_t floatValue{};
+	std::string stringValue{};
+	int32_t boolValue{};
+	int64_t int64Value{};
+	bool insertedFirstEntry = false;
+	std::stringstream insertedRow;
+	insertedRow << "INSERT INTO " << tableName << " values (";
```
Xiphoseer, line 188: The way assembly does this is instantiate a prepared statement per table after creating it and than calling that repeatedly for different params. Should speed up the process for larger tables because you do way less string juggling.

dCommon/FdbToSqlite.cpp
```diff
@@ -0,0 +1,248 @@
...
+	BinaryIO::BinaryRead(fdb, value);
+
+	fdb.seekg(prevPosition);
+	return value;
+}
+
+std::string FdbToSqlite::Convert::ReadString() {
+	int32_t prevPosition = SeekPointer();
+
+	auto readString = BinaryIO::ReadString(fdb);
```
Xiphoseer, line 70: I need to re-check the implementation, but this may break on e.g. technic parts, which use a latin1 `°` in their names while sqlite usually expects assumes utf8.

Xiphoseer (commented):
This looks like a solid way of getting the job done. There are a number of details about original FDBs this could use (ie aligned reads) that we can discuss in Discord. Otherwise, except for the-checking the encoding and a strong recommendation to not do escaping by hand, this LGTM.

PR #874, opened 2022-12-06
Remove GameConfig

EmosewaMC:
Fully remove the Game::config class and move the variable that the class used to the worldconfig file.


aronwk-aaron (approved):

Wincent01:
The purpose with this config class was to have some settings accessible by GMs. They shouldn't have access to things like database credentials, which this change gives them.

It was very lightly used, so a solution is to just remove the commands.

Wincent01 (dismissed):
See above comment

aronwk-aaron (dismissed):

Jettford (dismissed):

Jettford (approved):

aronwk-aaron (approved):

PR #877, opened 2022-12-06
Resolve warnings, change config init order and remove unused Game variables for all servers

EmosewaMC:
Initialize dConfig first, before logger so we know whether or not to log to console

This also incidentally fixes an issue brought up by @luxaritas about master not logging some initial server info to the console

Initialize namespace Game variables to nullptr so they are a known value if accessed before initialization. Removed unused Game variables

Replaced config with a pointer instead of referencing something on the stack.  Saves us memory used on the stack since this is used until the program shutsdown.

Assign return values to system calls to silence warnings.

Tested that the server still compiles, runs and allows me to load into the game.

### DIFF
--- a/dAuthServer/AuthServer.cpp
+++ b/dAuthServer/AuthServer.cpp
@@ -40,8 +40,8 @@ int main(int argc, char** argv) {
 	if (!Game::logger) return EXIT_FAILURE;
 
 	//Read our config:
-	Game::config = new dConfig("authconfig.ini");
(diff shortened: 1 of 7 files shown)


aronwk-aaron (approved):

luxaritas:
Alternatively, maybe it's worth verifying at least the server-specific config is present in config init, otherwise bailing

EmosewaMC:
Good suggestions.  I'll make the improvements

DarwinAnim8or (approved):

PR #880, opened 2022-12-08
Remove unneeded commands

EmosewaMC:
These are unneeded now since the missions now work.

I havent tested anything here but if its really needed ill test that you cant use the commands or something


aronwk-aaron (changes requested):
Remove from the document as well

aronwk-aaron (approved):

aronwk-aaron:
you're welcome

Wincent01 (approved):

PR #881, opened 2022-12-08
Allow player skills to be interrupted

aronwk-aaron:
This fixes behaviors not ending when the client sends an interrupt to the player.
This will infinite speed stacking when it was additive to the multiplier (or rather, permanent speed boost once that PR is merged what removes stacking speed boosts).

Tested several different scenarios behaviors for the player being interrupted to make sure they are handled sanely

### DIFF
--- a/dGame/dComponents/SkillComponent.cpp
+++ b/dGame/dComponents/SkillComponent.cpp
@@ -181,17 +181,11 @@ void SkillComponent::Reset() {
 }
 
 void SkillComponent::Interrupt() {
-	if (m_Parent->IsPlayer()) return;
-
+	// TODO: need to check immunities on the destroyable component, but they aren't implemented
 	auto* combat = m_Parent->GetComponent<BaseCombatAIComponent>();
+	if (combat != nullptr && combat->GetStunImmune()) return;
 
-	if (combat != nullptr && combat->GetStunImmune()) {
-		return;
-	}
-
-	for (const auto& behavior : this->m_managedBehaviors) {
-		behavior.second->Interrupt();
-	}
+	for (const auto& behavior : this->m_managedBehaviors) behavior.second->Interrupt();
 }
 
 void SkillComponent::RegisterCalculatedProjectile(const LWOOBJID projectileId, BehaviorContext* context, const BehaviorBranchContext& branch, const LOT lot, const float maxTime,


Wincent01:
I'm not sure entirely how the interrupt system works. How does the client expect an interrupt? Is it the case that the client has already been handling interrupts on its end and this is server parity; or is it entirely new for the client to receive interrupts? In the ladder case, is it enough to cancel any new syncs for the client on its side, like visually, recognize that an interrupt has occurred; just so it doesn't continue attempting to execute the skill with the server just removing its context.

aronwk-aaron:
> I'm not sure entirely how the interrupt system works. How does the client expect an interrupt? Is it the case that the client has already been handling interrupts on its end and this is server parity; or is it entirely new for the client to receive interrupts? In the ladder case, is it enough to cancel any new syncs for the client on its side, like visually, recognize that an interrupt has occurred; just so it doesn't continue attempting to execute the skill with the server just removing its context.

This allows for server parity for interrupts for players.

The server syncs the interrupt skill with the client that tells it what to interrupt and which immunities to check and only interrupt on certain types of skills.
If something can be interrupted and some skill tries to interrupt it, the client tries to tell the server to interrupt the skills, so the server knows to stop them as well.

I observed this secnario when tinkering with fixing the speed behavior:
If the client interrupted a speed behavior, then the client will correctly go back to normal walking speed (Maybe based on the speedbase in the levelcomp) until the server (which at this point disagrees with the client, since the speed behavior was never ended/interrupted, so the multiplier never get's corrected) serializes the controllable physics comp again and updates the speed multiplier on the client side, thus making it the incorrect server value.

This is the simplest fix that can currently be implemented.
a more robust fix would be to implement immunities and then only interrupt certain types of nodes (attack/charge/block) based upon the parameters in the interrupt behavior.
The target value in the interrupt behavior also (from what I have deduced, based on context clues from the behavior trees) should be the branch.target if true and the caster if false.
Additionally, there are several unknowns in the interrupt bitstream reading and writing that should also probably be figured out. But that will need to be a larger and more complex fix.

EmosewaMC (commented):
I approve this, but only cautiously.  We should keep an eye on this for edge cases and other issues this may cause.  If there are any, are you going to be able to help diagnose and resolve them?

aronwk-aaron:
Yep!

EmosewaMC (approved):

Wincent01 (approved):
I am cautious, please remain vigilant. I know the current interrupt behavior structure is wrong, and may and have in the past cause serialization faults. Would be unfortunate if some skills started to silently fail server-side. I do trust your testing, but do remain vigilant.

PR #883, opened 2022-12-12
Fix cannon super charge speed

EmosewaMC:
This should make the cannon actually super charge its attack speed as well, as it did in live.

This is untested but if someone wants to test this feel free to do so.  Until then I will comment when this is tested.


EmosewaMC:
This doesnt quite fix what I thought it fixes, but it makes the serialization at least correct from testing.

aronwk-aaron (approved):

Wincent01 (approved):

Wincent01:
Lgtm. One step towards filling in the many gaps in this instance.

PR #885, opened 2022-12-13
Log some recvfrom errors on linux

Xiphoseer:


aronwk-aaron (approved):

EmosewaMC:
Ideally this should be tested at least for functionality before merging just to ensure the code does what we expect it to do.

aronwk-aaron:
I can see about spinning this up on a server in my homelab and letting it run for a week to see if the issue happens and hits this code, unless there is a way that we could manually recreate the issue

Xiphoseer:
I thought about just adding two binaries with RakPeers, let them do some sort of ping roundtrip and then run it once with a forced exit and once with a crash, that should test a lot of the relevant code paths.

We could also send an empty udp packet by hand and see what happens, maybe.

Wincent01 (approved):

EmosewaMC:
Let's keep an eye on this for the next couple weeks.  If there aren't any issues past that then we should be all fine.

PR #886, opened 2022-12-14
Implement the Imaginite Backpack and Shard armor scripts

### LINKED ISSUE #376: BUG: Shard armor and Imaginite Pack do not grant the respective bonuses

### Issue Description

These items are supposed to restore armor/imagination after 5 hits but currently do not do so.

### Reproduction steps

1) equip one of the two items.
2) take 5 hits

### Expected Behavior

Gain armor or imagination after the hits

### Environment

server ubuntu
client windows 10

EmosewaMC:
fixes #376

Implements the scripts for the items above.  

Tested that equipping and unequipping these items properly adds/removes them from the list of subscribed scripts.
Tested that taking 5 hits properly triggers the skill value and resets when a new backpack is equipped that uses the same system.

I chose to go with a base class and inheritance so that I didn't need 3 copy and pasted scripts doing the exact same thing and only needed to provide a header implementation of scripts that only differ by a single value.


EmosewaMC:
Only thing I am maybe not a fan of is how I am passing the skill and behaviors down through the constructor delegation/chaining.  Not sure how to do this better without magic numbers.  If there is a better way please let me know

Wincent01:
Would it work to move this logic with subscribing to item scripts to the inventory component? Feels unnecessary for all entities to have this extra member without it being used.

EmosewaMC:
> Would it work to move this logic with subscribing to item scripts to the inventory component? Feels unnecessary for all entities to have this extra member without it being used.

That could be arranged.  I'll look into moving that.

dScripts/EquipmentTriggers/CoilBackpackBase.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include "Entity.h"
+#include "SkillComponent.h"
+
+void CoilBackpackBase::OnFactionTriggerItemEquipped(Entity* itemOwner, LWOOBJID itemObjId) {
+	itemOwner->Subscribe(itemObjId, this);
+	itemOwner->SetVar<uint8_t>(u"coilCount", 0);
+}
+
+void CoilBackpackBase::OnHitOrHealResult(Entity* self, Entity* attacker, int32_t damage) {
```
aronwk-aaron, line 11: should be `notifyHitOrHealResult`

To be consistent with the lua scripts and not conflict with a normal `OnHitOrHealResult` if there are ever two in the same script

aronwk-aaron (dismissed):

EmosewaMC:
All feedback is addressed - leaving this as a comment since I didn't yet.

aronwk-aaron (approved):

Wincent01 (approved):

PR #887, opened 2022-12-15
Improve server shutdown process

EmosewaMC:
Removes a global in a few servers and also properly shutsdown chat and auth when master shuts down.
Properly exit based on the path taken to shutdown master.

Tested that I can run MasterServer, send SIGINT and all servers properly shutdown.  There is no longer a delay when shutting down master between auth and chat shutting down as well.

Tested that shutting down through sigint or sigterm returns -1
Tested that a segfault exits the program properly
Need to test that players who are trying to connect while master is shutting down are not able to spawn more child worlds.


aronwk-aaron (approved):
LGTM

aronwk-aaron (approved):
re-approving

DarwinAnim8or (approved):

PR #890, opened 2022-12-16
Implement lower cap datagram / MTU size

EmosewaMC:
LEGO Universe uses a lower datagram / MTU size so we are also going to use this.  May also address other issues, we'll see.

Tested that the server compiles and boots.  

Will test that I can login with a substantially sized Create Character packet in the morning

### DIFF
--- a/dNet/dServer.cpp
+++ b/dNet/dServer.cpp
@@ -67,7 +67,7 @@ dServer::dServer(const std::string& ip, int port, int instanceID, int maxConnect
 	} else { mLogger->Log("dServer", "FAILED TO START SERVER ON IP/PORT: %s:%i", ip.c_str(), port); return; }
 
 	mLogger->SetLogToConsole(prevLogSetting);
-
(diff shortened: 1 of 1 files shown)


aronwk-aaron (approved):

aronwk-aaron:
Only questions is, Should this be set on all servers and not just world server?
Would make sense if that was the case, but just want to ask the stupid question incase no one has thought of it

EmosewaMC:
> Only questions is, Should this be set on all servers and not just world server? Would make sense if that was the case, but just want to ask the stupid question incase no one has thought of it

This is being done in dServer.  All dServers will have their MTU size set to this value, not just World

aronwk-aaron:
ty, i'm blind

Wincent01:
As with the rate limiter, a precompiler value with a default would be ideal imo.

Jettford (approved):

EmosewaMC:
> As with the rate limiter, a precompiler value with a default would be ideal imo.

Will do in a future pr.  I created #896 for that.

PR #891, opened 2022-12-17
Implement a server res directory for server required client files

EmosewaMC:
Tested that a sqlite file in the res path is correctly copied to `resServer`
Tested that if no sqlite file exists in the client res directory, a new one is created in `resServer`
Tested that is the sqlite file exists in resServer but not in the client res path, the file is not converted again


HailStorm32:
The README will also need to be updated if you are renaming the res folder

EmosewaMC:
> The README will also need to be updated if you are renaming the res folder

Waiting on the other readme update and if that is merged first ill update it there.

Jettford:
This will make simply dragging and dropping binaries into a client in the future more difficult, I do not agree with this change however half of that is because there is no reason for these changes listed here. Please outline reasons for this change.

luxaritas:
I pushed for this. Best practice for daemons is that they only write to directories which do not contain any other content aside from what it contains. The rationale is twofold:
1) Being intentional about not modifying input file locations. The assumption is that the client path is being used solely to read from, and having files written there. If the server needs to create additional files, they should be created somewhere it "owns", unless the explicit purpose is to generate those files in the manner the user asked.
2) Server hardening. As a manner of best practice, daemons should only have write access to what they strictly need to. Currently, you need to allow writes into build and res, which could allow for unintended writes of application files given another vulnerability. This provides defense-in-depth.

Regardless, I don't see why this has to make shipping binaries harder. The server can default create serverRes/CDServer.SQLite (including the parent dir) as well as it can res/CDServer.sqlite

Jettford:
You are missing the point, if you were to say drop the binaries into a client the res is called res, it won't know where to source the cdclient.fdb to convert.

Jettford:
Ok after re-going through the source I think I actually understand now but this just raises a different question of why does it matter if the server has R and W on client files?

luxaritas:
It's not critical, but I still maintain that as a matter of practice we shouldn't be modifying someone's client in any way. The CDServer is a server file. Keep it in a server-controlled area.

aronwk-aaron (approved):

EmosewaMC:
> The README will also need to be updated if you are renaming the res folder

This is not needed anymore since no user needs to interact with a server res directory with the new readme structure

dMasterServer/MasterServer.cpp
```diff
@@ -144,7 +144,8 @@ int main(int argc, char** argv) {
 	MigrationRunner::RunMigrations();
 
 	// Check CDClient exists
-	if (!std::filesystem::exists(Game::assetManager->GetResPath() / "CDServer.sqlite")) {
+	if (!std::filesystem::exists(Game::assetManager->GetResPath() / "CDServer.sqlite") &&
```
Xiphoseer, line 147: might make sense to extract those exists calls into variables if you use them multiple times.

dMasterServer/MasterServer.cpp
```diff
@@ -154,15 +155,19 @@ int main(int argc, char** argv) {
 
 		Game::logger->Log("WorldServer", "Found cdclient.fdb.  Converting to SQLite");
 
-		if (FdbToSqlite::Convert(Game::assetManager->GetResPath().string()).ConvertDatabase() == false) {
+		if (FdbToSqlite::Convert(Game::assetManager->GetResPath().string(), (BinaryPathFinder::GetBinaryDir() / "resServer").string()).ConvertDatabase() == false) {
 			Game::logger->Log("MasterServer", "Failed to convert fdb to sqlite");
 			return EXIT_FAILURE;
 		}
+	} else if (!std::filesystem::exists(BinaryPathFinder::GetBinaryDir() / "resServer" / "CDServer.sqlite")) {
```
Xiphoseer, line 162: I think this condition is a bit weird to reason about. Yes, if both don't exist, then it would have gone to the first if case, so if the sqlite in resServer does not exist, and we get here there must be a sqlite in res to move over.

But considering, that this is essentially a migration, I think this should just be a positive check for existance, nested under a simple "resServer sqlite missing" before converting from fdb

dMasterServer/MasterServer.cpp
```diff
@@ -154,15 +155,19 @@ int main(int argc, char** argv) {
...
+		// If the file doesn't exist in the new CDServer location, move it there
+		Game::logger->Log("MasterServer", "CDServer.sqlite is not located at resServer, but is located at res path.  Moving file...");
+		std::filesystem::copy_file(Game::assetManager->GetResPath() / "CDServer.sqlite", BinaryPathFinder::GetBinaryDir() / "resServer" / "CDServer.sqlite");
```
Xiphoseer, line 165: Add a comment that this copies instead of move because the folder may be read only

Xiphoseer (dismissed):

dMasterServer/MasterServer.cpp
EmosewaMC: Thats fair - ill add those

EmosewaMC:
Tested having fdb and no sqlite, having fdb and sqlite in the old directory, having sqlite in new directory with packed and unpacked clients.  All resulted in cdserver being in resServer

aronwk-aaron (approved):

dMasterServer/MasterServer.cpp
```diff
@@ -143,26 +143,38 @@ int main(int argc, char** argv) {
...
-
-		if (FdbToSqlite::Convert(Game::assetManager->GetResPath().string()).ConvertDatabase() == false) {
-			Game::logger->Log("MasterServer", "Failed to convert fdb to sqlite");
-			return EXIT_FAILURE;
+	const bool cdServerExists = std::filesystem::exists(BinaryPathFinder::GetBinaryDir() / "resServer" / "CDServer.sqlite");
+	const bool oldCDServerExists = std::filesystem::exists(Game::assetManager->GetResPath() / "CDServer.sqlite");
+	const bool fdbExists = std::filesystem::exists(Game::assetManager->GetResPath() / "cdclient.fdb");
+
+	if (!cdServerExists) {
+		if (!oldCDServerExists) {
```
Xiphoseer, line 151: Can you swap the branches and remove the `!`?

Xiphoseer (approved):

dMasterServer/MasterServer.cpp
EmosewaMC: yes - will do so in the morning

Xiphoseer (dismissed):
LGTM

dMasterServer/MasterServer.cpp
```diff
@@ -143,26 +143,38 @@ int main(int argc, char** argv) {
...
+				Game::logger->Log("WorldServer",
+					"%s could not be opened. Please move cdclient.fdb to %s",
+					(Game::assetManager->GetResPath() / "cdclient.fdb").c_str(),
+					(Game::assetManager->GetResPath().c_str()));
+				FinalizeShutdown();
+			}
+			Game::logger->Log("WorldServer", "Found cdclient.fdb.  Converting to SQLite");
+			if (FdbToSqlite::Convert(Game::assetManager->GetResPath().string(), (BinaryPathFinder::GetBinaryDir() / "resServer").string()).ConvertDatabase() == false) {
+				Game::logger->Log("MasterServer", "Failed to convert fdb to sqlite.");
+				return EXIT_FAILURE;
```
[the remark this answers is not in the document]
EmosewaMC: Mistake on my end.  Its supposed to return FinalizeShutdown(); so destroy all objects that exist

aronwk-aaron (approved):

Xiphoseer (dismissed):

aronwk-aaron (approved):

Xiphoseer (approved):

PR #892, opened 2022-12-17
Resolve some string related issues for Windows Debug

Jettford:


EmosewaMC:
Ah right we cant use strings like that in debug.  I still wonder why we need to do this in the first place

EmosewaMC (approved):

DarwinAnim8or (approved):
Should work, goddamn windows and it's prehistoric *mic cut*

PR #893, opened 2022-12-17
Rewritten XML loop that was crashing.

Jettford:
I am uncertain why it was crashing however the loop seemed a bit sketch so I just rewrote it and it resolved the issue on my end.


Jettford:
never mind still happens lol

PR #897, opened 2022-12-18
Implement DarkInspiration node

EmosewaMC:
Will test in a few minutes.

Implements the dark inspiration behavior node.  Since no skills attached to it are cast at the minute, this is effectively a no-op, but is needed for their implementation.

Will be testing that shinobi rank 3 properly executes its behavior.


dGame/dBehaviors/DarkInspirationBehavior.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "Entity.h"
+#include "DestroyableComponent.h"
+#include "EntityManager.h"
+#include "BehaviorContext.h"
+
+void DarkInspirationBehavior::Handle(BehaviorContext* context, RakNet::BitStream* bitStream, const BehaviorBranchContext branch) {
+	auto* target = EntityManager::Instance()->GetEntity(branch.target);
+
+	if (target == nullptr) {
+		Game::logger->Log("DamageReductionBehavior", "Failed to find target (%llu)!", branch.target);
```
aronwk-aaron, line 13: Wrong log file name, also, maybe make debug?

aronwk-aaron (commented):
One comment, that's all

dGame/dBehaviors/DarkInspirationBehavior.cpp
EmosewaMC: ah right, just copy and paste error

aronwk-aaron (approved):

dGame/dInventory/ItemSetPassiveAbility.cpp
```diff
@@ -293,8 +293,8 @@ void ItemSetPassiveAbility::OnEnemySmshed() {
 		break;
 	}
 	case ItemSetPassiveAbilityID::ShinobiRank3: {
-		if (equippedCount < 4) return;
-		destroyableComponent->Imagine(3);
+		if (equippedCount < 4 || !target) return;
+		skillComponent->CalculateBehavior(695, 11399, target->GetObjectID());
```
Wincent01, line 297: Unfortunate hardcoding. Is there some system behind this which would explain why this is necessary, are all set skills behaviors?

EmosewaMC: There is, but to keep this PR small and just as a proof of concept that the behavior itself works, I only chose to replace this one specifically with the behavior like this.  The actual implementation will be far cleaner

EmosewaMC: I can go into more detail if you want for how this is supposed to be implemented, whether you want all of these to be fixed at once, leave this alone for now and just implement the node, or implement the node and this one check.  any I am fine with

Jettford (approved):

PR #898, opened 2022-12-18
Fix overread in projectile behavior and address broken stuns

EmosewaMC:
When this reading existed, the logic failed with the number of unread bits being exactly 64.  A NiPoint3 is 96 bits (3 floats) so this reading is not correct and caused the plunger gun in Avant Gardens to not work.

Tested that the plunger gun now correctly destroys targets.

### DIFF
--- a/dGame/dBehaviors/ProjectileAttackBehavior.cpp
+++ b/dGame/dBehaviors/ProjectileAttackBehavior.cpp
@@ -31,14 +31,6 @@ void ProjectileAttackBehavior::Handle(BehaviorContext* context, RakNet::BitStrea
 		return;
 	}
 
-	if (m_useMouseposit) {
-		NiPoint3 targetPosition = NiPoint3::ZERO;
-		if (!bitStream->Read(targetPosition)) {
-			Game::logger->Log("ProjectileAttackBehavior", "Unable to read targetPosition from bitStream, aborting Handle! %i", bitStream->GetNumberOfUnreadBits());
-			return;
-		};
-	}
-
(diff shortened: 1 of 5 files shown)


aronwk-aaron (approved):

aronwk-aaron (dismissed):

dGame/dBehaviors/ProjectileAttackBehavior.cpp
```diff
@@ -31,14 +31,6 @@ void ProjectileAttackBehavior::Handle(BehaviorContext* context, RakNet::BitStrea
...
 	}
 
-	if (m_useMouseposit) {
-		NiPoint3 targetPosition = NiPoint3::ZERO;
-		if (!bitStream->Read(targetPosition)) {
-			Game::logger->Log("ProjectileAttackBehavior", "Unable to read targetPosition from bitStream, aborting Handle! %i", bitStream->GetNumberOfUnreadBits());
-			return;
-		};
-	}
-
```
Wincent01, line 41: Is this not part of the structure? Check that Cannon Cove still works.

EmosewaMC: This is apart of the structure after further investigation.  It looks to be based off of projectile_type instead of m_useMouseposit instead however as alongwith the plunger gun, lot 4799 also failed on main, which has a projectile_type of 2 as opposed to Cannon Coves 1.

Tested that 4799 and the plunger gun and the shooting gallery all properly use their skills and do not fault.

aronwk-aaron (approved):

Wincent01 (approved):

PR #900, opened 2022-12-19
Fix trading taking the wrong item

EmosewaMC:
When trading, we need to take the item by ID and not by lot.  This caused the wrong item to be taken.

Tested that trading properly takes items by ID and not by lot.

### DIFF
--- a/dGame/TradingManager.cpp
+++ b/dGame/TradingManager.cpp
@@ -121,20 +121,34 @@ void Trade::Complete() {
 
 	if (inventoryA == nullptr || inventoryB == nullptr || characterA == nullptr || characterB == nullptr || missionsA == nullptr || missionsB == nullptr) return;
 
-	characterA->SetCoins(characterA->GetCoins() - m_CoinsA + m_CoinsB, eLootSourceType::LOOT_SOURCE_TRADE);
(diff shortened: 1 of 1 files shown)


aronwk-aaron (approved):

Wincent01:
You say you're aborting but nothing happens to support that, no return statement etc. Maybe before doing any transaction check that it is legal.

Wincent01 (changes requested):
See above comment

EmosewaMC:
> See above comment

lol, guess I missed that.  Apologies

EmosewaMC:
Going to reinforce coins as well.

EmosewaMC:
Repeated the same tests, players who trade exactly enough coins or who trade exactly how many items they have can still trade, but if you somehow go over the amount in a stack, then the trade does not happen.

dGame/TradingManager.cpp
```diff
@@ -103,6 +103,7 @@ void Trade::SetAccepted(LWOOBJID participant, bool value) {
 		}
 
 		Complete();
+		TradingManager::Instance()->CancelTrade(m_TradeId);
```
aronwk-aaron, line 106: Why is this here?

EmosewaMC: Because previously this was at the bottom of complete, however we may not always hit that.

Wincent01 (approved):

aronwk-aaron (approved):

PR #908, opened 2022-12-21
Add MTU config option

### LINKED ISSUE #896: ENH: Implement MTU config

MTU is currently hardcoded as 1228, but should be configurable.

Add to sharedconfig.ini this value

none

none

### Additional context

none

EmosewaMC:
Fixes #896 

Tested that if the value is not there, the default live value is used.  If the value is there then this config option takes effect.

### DIFF
--- a/dGame/dUtilities/SlashCommandHandler.cpp
+++ b/dGame/dUtilities/SlashCommandHandler.cpp
@@ -1762,6 +1762,7 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 
 			scriptedActivityComponent->ReloadConfig();
 		}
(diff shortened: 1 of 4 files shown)


aronwk-aaron:
If you set this to like 1 will it be that low or bottom out at 512, or actually be 1?

EmosewaMC:
> If you set this to like 1 will it be that low or bottom out at 512, or actually be 1?

RakPeer will bottom the value out at 512 and cap it at 1492.

aronwk-aaron (dismissed):

aronwk-aaron (approved):

Wincent01 (approved):

PR #912, opened 2022-12-22
Qb team credit

aronwk-aaron:
When a QB is done when someone is on a team, given every member mission/achievement credit for doing so

### DIFF
--- a/dGame/dComponents/RebuildComponent.cpp
+++ b/dGame/dComponents/RebuildComponent.cpp
@@ -464,12 +464,20 @@ void RebuildComponent::CompleteRebuild(Entity* user) {
 
 	auto* builder = GetBuilder();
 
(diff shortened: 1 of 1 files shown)


EmosewaMC:
What kind of testing have you done?

aronwk-aaron:
Tested that it works for one player by themself, tested that it works two players on a team

EmosewaMC (approved):

Wincent01:
Are you sure this is live accurate?

EmosewaMC:
> Are you sure this is live accurate?

100%.  Confirmed with live packet captures that a _team_ would get credit for a build completion, but only the _builder_ would get the loot drops.  Can show you the packets I used to deduce this later if needed.

Wincent01 (approved):

PR #914, opened 2022-12-22
Fix baseEnemyApe stuns and fix IdleFlags serialization

### LINKED ISSUE #905: BUG: Ape ground pound after awaking from stun has no animation

### DarkflameServer Version

f2fa81b

### Issue Description

This has been the case forever. After depleting an ape's armor and attacking it at close range, when it gains armor back and awakens from the stun it will ground pound the player without displaying the animation, which is a surprise attack to anyone who isn't aware of this. Doesn't happen every time, but most of the time.

### Reproduction Steps

Attack an ape at close range as it begins to wake up from the stun and its armor has been partially restored.

### Expected Behavior

I'm going to assume that since the attack happens right after the typical stun animation, that there's an attack delay that hasn't been put into place for ground pounds.

### Environment

ubuntu server

EmosewaMC:
Live LUA script called CancelSkillCast, which I presume is to stop the ape from casting its skills since the ape cannot have any of their skills interrupted.

Also addresses an issue where IdleFlags was not being serialized correctly.  Idle flags have defaults as seen in Reverse Engineered source code and as is documented [here](https://lcdruniverse.org/lu_packets/src/lu_packets/world/gm/client/mod.rs.html#1363-1368)

Tested that no longer do their ground pound silently upon getting out of stun, even if they were stunned while winding the skill up.

Fixes #905

### DIFF
--- a/dGame/dGameMessages/GameMessages.cpp
+++ b/dGame/dGameMessages/GameMessages.cpp
@@ -3930,7 +3930,9 @@ void GameMessages::SendChangeIdleFlags(LWOOBJID objectId, eAnimationFlags flagsO
 
 	bitStream.Write(objectId);
 	bitStream.Write(GAME_MSG::GAME_MSG_CHANGE_IDLE_FLAGS);
+	bitStream.Write<bool>(flagsOff != eAnimationFlags::IDLE_NONE);
 	if (flagsOff != eAnimationFlags::IDLE_NONE) bitStream.Write(flagsOff);
+	bitStream.Write<bool>(flagsOn != eAnimationFlags::IDLE_NONE);
 	if (flagsOn != eAnimationFlags::IDLE_NONE) bitStream.Write(flagsOn);
 
 	SEND_PACKET_BROADCAST;
--- a/dGame/dGameMessages/GameMessages.cpp
+++ b/dGame/dGameMessages/GameMessages.cpp
@@ -3924,14 +3924,14 @@ void GameMessages::SendDisplayChatBubble(LWOOBJID objectId, const std::u16string
 }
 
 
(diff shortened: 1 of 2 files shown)


aronwk-aaron (approved):

dGame/dGameMessages/GameMessages.cpp
```diff
@@ -3924,14 +3924,14 @@ void GameMessages::SendDisplayChatBubble(LWOOBJID objectId, const std::u16string
...
+void GameMessages::SendChangeIdleFlags(LWOOBJID objectId, eAnimationFlags flagsOn, eAnimationFlags flagsOff, const SystemAddress& sysAddr) {
 	CBITSTREAM;
 	CMSGHEADER;
 
 	bitStream.Write(objectId);
 	bitStream.Write(GAME_MSG::GAME_MSG_CHANGE_IDLE_FLAGS);
-	bitStream.Write(FlagsOff);
-	bitStream.Write(FlagsOn);
+	if (flagsOff != eAnimationFlags::IDLE_NONE) bitStream.Write(flagsOff);
+	if (flagsOn != eAnimationFlags::IDLE_NONE) bitStream.Write(flagsOn);
```
aronwk-aaron, line 3934: This may be incorrect. The default is 0, which is handled in the behavior.
Not "only send if you're not 0"

aronwk-aaron (changes requested):

dGame/dGameMessages/GameMessages.cpp
EmosewaMC: this is not what is shown in the client.  The client has a read bit flag, followed by a read bits if the read bit was true.  This happens twice.

aronwk-aaron (approved):

Wincent01 (approved):

PR #922, opened 2022-12-24
remove memory leaks in hydrant scripts

aronwk-aaron:
i am plumber, i fix leaks


Wincent01 (approved):
"Remove" makes this fault sound  all intentional :P

aronwk-aaron:
> "Remove" makes this fault sound all intentional :P

more like me not knowing what I'm doing lol

EmosewaMC (commented):
I'm confused, won't this cause undefined behavior in entity settings is this ldf data is accessed?

aronwk-aaron:
Leaks were else where

PR #924, opened 2022-12-25
Improve Interrupt Behavior

EmosewaMC:
The motivation for this PR is to improve stuns so we dont just blanket stun all behaviors of an entity, but rather only behaviors that should be interrupted and that are not able to be blocked.  The first unknown definition comes from a GM call in the client to check immunity and if that is true, write 1 otherwise write 0.  The same goes for the second unknown.  This was the isBlocking flag as the client does a GetStatus check and if they are blocking and this behavior can interrupt blocks, interrupt the block and write 0, otherwise write 1.  The last bit is always a zero in my testing and is always written and I believe to be apart of a deprecated serialization system for the behavior in that it used to serialize what skillUIds were interrupted, but the client never logs this information and it always seems to be false.
There is only so much I can think of to test, so if you have special test cases where currently interrupts dont work that this may fix, let me know.  Otherwise all functionality appears to be good, if not a bit better on the surface.


dGame/dBehaviors/BehaviorContext.cpp
```diff
@@ -260,16 +260,32 @@ bool BehaviorContext::CalculateUpdate(const float deltaTime) {
...
+			}
+		}
 		keptSync.push_back(entry);
 	}
-
 	this->syncEntries = keptSync;
+
+	std::vector<BehaviorEndEntry> toKeep{};
+	for (const auto& entry : this->endEntries) {
+		if ((interruptCharge || interruptAttack) && entry.behavior->m_templateId == BehaviorTemplates::BEHAVIOR_START) {
```
aronwk-aaron, line 282: Does this not need to happen for block as well??

dGame/dBehaviors/BehaviorContext.cpp
```diff
@@ -260,16 +260,32 @@ bool BehaviorContext::CalculateUpdate(const float deltaTime) {
...
 }
 
-void BehaviorContext::Interrupt() {
+void BehaviorContext::Interrupt(bool interruptAttack, bool interruptBlock, bool interruptCharge) {
 	std::vector<BehaviorSyncEntry> keptSync{};
-
 	for (const auto& entry : this->syncEntries) {
-		if (!entry.ignoreInterrupts) continue;
+		if (!entry.ignoreInterrupts) {
+			if (interruptAttack && entry.behavior->m_templateId == BehaviorTemplates::BEHAVIOR_ATTACK_DELAY) {
```
aronwk-aaron, line 267: could these be combined into one if statement?

aronwk-aaron (approved):
just two questions, but other than that, looks good

dGame/dBehaviors/BehaviorContext.cpp
EmosewaMC: Not according to the client, but I will check again

dGame/dBehaviors/BehaviorContext.cpp
EmosewaMC: It could, but it would be quite large so keeping them as their own statement so its clear to a reader what is going on I feel would be more ideal

dScripts/02_server/Enemy/AM/AmDarklingDragon.cpp
```diff
@@ -62,7 +62,6 @@ void AmDarklingDragon::OnHitOrHealResult(Entity* self, Entity* attacker, int32_t
 			}
 
 			if (skillComponent != nullptr) {
-				skillComponent->Interrupt();
```
Wincent01, line 65: Why has this been removed?

dScripts/02_server/Enemy/General/BaseEnemyApe.cpp
```diff
@@ -126,7 +126,7 @@ void BaseEnemyApe::StunApe(Entity* self, bool stunState) {
 
 		auto* skillComponent = self->GetComponent<SkillComponent>();
 		if (skillComponent != nullptr) {
-			skillComponent->Interrupt();
```
Wincent01, line 129: Same here

Wincent01 (commented):
Looks good, just a bit confused why the scripts were changed.

dScripts/02_server/Enemy/AM/AmDarklingDragon.cpp
EmosewaMC: They cannot be interrupted anyways and are reset in the line below

dScripts/02_server/Enemy/General/BaseEnemyApe.cpp
EmosewaMC: This one is replaced by a reset, which actually resets the ape upon waking up (even though it shouldn't have skills active at that point) as opposed to interrupt which does not do anything because the ape is stun immune

dGame/dBehaviors/BehaviorContext.cpp
EmosewaMC: According to the client Block behaviors are not to be interrupted, but instead just ended if this behavior is able to punch through them. Will make that change.

EmosewaMC:
Converting to draft while I work out the improvements mentioned above and verify the new version is 100% correct

EmosewaMC:
old pr - binning

PR #925, opened 2022-12-26
Imminuty updates

aronwk-aaron:
This PR implements the groundwork for immunities.

The serializations have been added in the destroyable and controllable physics components
Tested that the client does not have any issues with this serialization (no errors in client logs)

Added the GM's for Status and Stun immunities
Tested that they work

Added Getters and setters for status and stun immunities in their respective components

Updated Immunity behavior to check for and properly handle all immunities 
NOTE: 3 are unused, two do not appear in the client, one does. Leaving that one in since it could be used in the future

For code simplification, the setter in the components also send the GM, this is to allow for simpler code calling and to let us keep track of client and server side immunities in-sync

This is the first of 3 (or more) PR's for immunities
This one is to get the system in place to use
Next one will be using the immunities properly server side
Third will be to update the usage of the old (or non-usage in scripts), which may be split into multiple PR's due to the scope of the many scripts that may be touched


dScripts/02_server/Enemy/AG/BossSpiderQueenEnemyServer.cpp
```diff
@@ -114,7 +106,10 @@ void BossSpiderQueenEnemyServer::WithdrawSpider(Entity* self, const bool withdra
 
 		//TODO: Set faction to -1 and set immunity
 		destroyable->SetFaction(-1);
-		destroyable->SetIsImmune(true);
+		GameMessages::SendSetStatusImmunity(
+			self->GetObjectID(), eStateChangeType::PUSH, self->GetSystemAddress(),
+			true, true, false, false, true
```
EmosewaMC, line 111: What is the motivation behind using some of these defaults?  Assuming it is because of the client script?

dScripts/EquipmentScripts/PersonalFortress.cpp
```diff
@@ -2,49 +2,56 @@
...
 	self->AddTimer("FireSkill", 1.5);
-	GameMessages::SendSetStunned(owner->GetObjectID(), PUSH, owner->GetSystemAddress(), LWOOBJID_EMPTY,
-		true, true, true, true, true, true, true, true, true
-	);
 
-	auto* destroyableComponent = owner->GetComponent<DestroyableComponent>();
+	auto* destroyableComp = owner->GetComponent<DestroyableComponent>();
+	if (destroyableComp) destroyableComp->SetStatusImmunity(
+			eStateChangeType::PUSH,
+			true, true, true, true, true, false, true, false, false
```
EmosewaMC, line 15: Im seeing a mismatch in immunity pushes here.  Can you elaborate on the change?

dScripts/EquipmentScripts/PersonalFortress.cpp
```diff
@@ -2,49 +2,56 @@
...
-	}
+	auto* controllablePhysicsComp = owner->GetComponent<ControllablePhysicsComponent>();
+	if (controllablePhysicsComp) controllablePhysicsComp->SetStunImmunity(
+			eStateChangeType::PUSH, LWOOBJID_EMPTY,
+			true, true, true, true, true, true
+		);
+
+	GameMessages::SendSetStunned(owner->GetObjectID(), eStateChangeType::PUSH, owner->GetSystemAddress(), LWOOBJID_EMPTY,
+		true, true, true, true, true, true, true, true, true
+	);
```
EmosewaMC, line 26: What is the motivation for this change?

dScripts/ai/MINIGAME/SG_GF/SERVER/SGCannon.cpp
```diff
@@ -76,7 +76,7 @@ void SGCannon::OnActivityStateChangeRequest(Entity* self, LWOOBJID senderID, int
 		auto* player = EntityManager::Instance()->GetEntity(self->GetVar<LWOOBJID>(PlayerIDVariable));
 		if (player != nullptr) {
 			Game::logger->Log("SGCannon", "Player is ready");
-			/*GameMessages::SendSetStunned(player->GetObjectID(), PUSH, player->GetSystemAddress(), LWOOBJID_EMPTY,
+			/*GameMessages::SendSetStunned(player->GetObjectID(), eStateChangeType::PUSH, player->GetSystemAddress(), LWOOBJID_EMPTY,
```
EmosewaMC, line 79: Why are we keeping this code?  Is this in the live script?

tests/dGameTests/dComponentsTests/DestroyableComponentTests.cpp
```diff
@@ -41,10 +41,19 @@ class DestroyableTest : public GameDependenciesTest {
...
+		bool optionStatusImmunityInfo{};
+		uint32_t ImmuneToBasicAttackCount{};
+		uint32_t ImmuneToDamageOverTimeCount{};
+		uint32_t ImmuneToKnockbackCount{};
+		uint32_t ImmuneToInterruptCount{};
+		uint32_t ImmuneToSpeedCount{};
+		uint32_t ImmuneToImaginationGainCount{};
+		uint32_t ImmuneToImaginationLossCount{};
+		uint32_t ImmuneToQuickbuildInterruptCount{};
+		uint32_t ImmuneToPullToPointCount{};
```
EmosewaMC, line 56: it may be worth considering pushing some immune stacks and checking that the right variables are set in the right spots i.e. basicAttackCount set to 1, DamageOverTImeCount set to 2 etc. so if the order is swapped on accident, this test would catch that

dGame/dComponents/InventoryComponent.cpp
```diff
@@ -994,7 +994,7 @@ void InventoryComponent::HandlePossession(Item* item) {
 	auto* destroyableComponent = mount->GetComponent<DestroyableComponent>();
 	if (destroyableComponent) {
 		destroyableComponent->SetIsSmashable(false);
-		destroyableComponent->SetIsImmune(true);
+		// TODO: Get proper immunities
```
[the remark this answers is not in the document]
aronwk-aaron: left over from a refactor, and it's part of the mount system which wasn't in live so my thought was "I'll lookup what the packets say later", will remove since the previous code works

dScripts/02_server/Enemy/AG/BossSpiderQueenEnemyServer.cpp
```diff
@@ -77,28 +77,20 @@ void BossSpiderQueenEnemyServer::OnDie(Entity* self, Entity* killer) {
...
-
-		controllable->SetDirtyPosition(true);
-
-		rot = controllable->GetRotation();
+		// Move spider away from battle zone
+		// Disabled because we cant option the reset collision group right now
+		GameMessages::SendNotifyClientObject(self->GetObjectID(), u"SetColGroup", 10, 0, 0, "", UNASSIGNED_SYSTEM_ADDRESS);
 
+		controllable->SetRotation({0, -0.005077, 0, 0.999});
+		controllable->SetIsTeleporting(true);
```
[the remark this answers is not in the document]
aronwk-aaron: I got into scripts, which I realized I should not have, will revert

dScripts/02_server/Enemy/AG/BossSpiderQueenEnemyServer.cpp
aronwk-aaron: yes, scripts largely don't use stun or use the jank immunities, this is the equivalent to the lua script here. I will revert this and will do a script pass in another PR

dScripts/02_server/Enemy/AG/BossSpiderQueenEnemyServer.cpp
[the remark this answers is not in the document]
aronwk-aaron: his script uses a lot of stun and immunities, and I was like "oooo I can fis this now" then back off, missed reverting this before I pushed, even thought this is more correct

dScripts/ai/MINIGAME/SG_GF/SERVER/SGCannon.cpp
aronwk-aaron: I found and replaced that variable so that's why it changed, and yes it's in the live script.

tests/dGameTests/dComponentsTests/DestroyableComponentTests.cpp
aronwk-aaron: Can do sir

dScripts/EquipmentScripts/PersonalFortress.cpp
aronwk-aaron: Okay so personal fortress is the one script I will leave in this PR since I completely broke it.
I removed the two function calls it was using which was basically all the script did beside a cast skill, so I properly implemented it as a test to the immunity code

dScripts/EquipmentScripts/PersonalFortress.cpp
aronwk-aaron: this is in line with the script from live, since it was a small script and made use of the immunities, and since I broke it, it was good to test with.
I just double checked that this lines up with the lua script, and it does

dGame/dGameMessages/GameMessages.cpp
```diff
@@ -2947,6 +2947,69 @@ void GameMessages::SendSetStunned(LWOOBJID objectId, eStunState stateChangeType,
...
+
+	bitStream.Write(bImmuneToStunAttack);
+	bitStream.Write(bImmuneToStunEquip);
+	bitStream.Write(bImmuneToStunInteract);
+	bitStream.Write(bImmuneToStunJump);
+	bitStream.Write(bImmuneToStunMove);
+	bitStream.Write(bImmuneToStunTurn);
+	bitStream.Write(bImmuneToStunUseItem);
+
+	if (sysAddr == UNASSIGNED_SYSTEM_ADDRESS) SEND_PACKET_BROADCAST;
```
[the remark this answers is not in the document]
aronwk-aaron: Yes, this logic lines up with other GM's in packet captures

Jettford (approved):

EmosewaMC (approved):

PR #926, opened 2022-12-27
Basic Attack Behavior Live Accuracy Improvements

EmosewaMC:
The motivation behind this change is to ensure we are sending exactly what we need to when serializing behaviors to the client.  The Basic Attack Behavior was correct before in what bits went where, however the client would be expecting less information than what we were sending which caused disagreements between the client and the server with certain variables.  Handles and Calculations were updated to have a wrapper class that handles the reading / writing of the size offset variable that is serialized in the bitStream.  An enum was added to clarify what success states are what, as well as the switch case statement being changed to be how the client handles the behavior.  The unknown in the bitStream was proven to be armor damage taken, both through deserialization and also through Reverse Engineering.

Some easy to find notable changes:
The stromling mech now properly shows the block animation when the player is blocking their attack.
The player while in the Personal Fortress no longer moves around in most contexts when the ability is active.  Some cases use Verify Behavior to confirm this information, so those cases are still non-functional.
Client logs show significant reductions in number of bits unread in bitStreams
Pre update: `SyncSkill sbitstream has unread bits:104 skillhandle:12742 skillid:949 caster:288300748454049596`
Post update: `SyncSkill sbitstream has unread bits:24 skillhandle:12735 skillid:949 caster:288300748454049596`
Likely more skill related bugs fixed, however there is only so much I can test.  If there are more ideas of what to test, please let me know or suggest some things to test down below.

### DIFF
--- a/dGame/dBehaviors/BasicAttackBehavior.cpp
+++ b/dGame/dBehaviors/BasicAttackBehavior.cpp
@@ -33,11 +33,23 @@ void BasicAttackBehavior::Handle(BehaviorContext* context, RakNet::BitStream* bi
 	const auto baseAddress = bitStream->GetReadOffset();
 
 	DoHandleBehavior(context, bitStream, branch);
-	
+
 	bitStream->SetReadOffset(baseAddress + allocatedBits);
 }
 
 void BasicAttackBehavior::DoHandleBehavior(BehaviorContext* context, RakNet::BitStream* bitStream, BehaviorBranchContext branch) {
+	auto* targetEntity = EntityManager::Instance()->GetEntity(branch.target);
+	if (!targetEntity) {
+		Game::logger->Log("BasicAttackBehavior", "Target targetEntity %i not found.", branch.target);
+		return;
+	}
+
+	auto* destroyableComponent = targetEntity->GetComponent<DestroyableComponent>();
+	if (!destroyableComponent) {
+		Game::logger->Log("BasicAttackBehavior", "No destroyable found on the obj/lot %llu/%i", branch.target, targetEntity->GetLOT());
+		return;
+	}
+
 	bool isBlocked{};
 	bool isImmune{};
 	bool isSuccess{};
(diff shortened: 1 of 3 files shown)


dGame/dBehaviors/BasicAttackBehavior.cpp
```diff
@@ -89,79 +112,119 @@ void BasicAttackBehavior::Handle(BehaviorContext* context, RakNet::BitStream* bi
...
+		return;
+	}
 
-			auto* destroyableComponent = entity->GetComponent<DestroyableComponent>();
-			if (damage != 0 && destroyableComponent != nullptr) {
-				PlayFx(u"onhit", entity->GetObjectID(), 1);
-				destroyableComponent->Damage(damage, context->originator, context->skillID, false);
-				context->ScheduleUpdate(branch.target);
-			}
+	const bool isBlocking = destroyableComponent->GetAttacksToBlock() > 0;
```
Wincent01, line 164: Should we decrease this value when evaluating this behavior?

EmosewaMC: Good catch.  My thought process was this is handled in the Damage method in DestroyableComponent, so I completely missed this.

dGame/dBehaviors/BasicAttackBehavior.cpp
```diff
@@ -40,47 +47,63 @@ void BasicAttackBehavior::Handle(BehaviorContext* context, RakNet::BitStream* bi
 		return;
 	}
 
-	if (isBlocked) return;
+	if (isBlocked) {
+		this->m_OnFailBlocked->Handle(context, bitStream, branch);
```
EmosewaMC, line 51: Decrement block stacks here by the proper amount

dGame/dBehaviors/BasicAttackBehavior.cpp
```diff
@@ -31,137 +31,208 @@ void BasicAttackBehavior::Handle(BehaviorContext* context, RakNet::BitStream* bi
...
 
 	if (!bitStream->Read(isBlocked)) {
-		Game::logger->LogDebug("BasicAttackBehavior", "Unable to read isBlocked");
+		Game::logger->Log("BasicAttackBehavior", "Unable to read isBlocked");
 		return;
 	}
 
-	if (isBlocked) return;
+	if (isBlocked) {
+		destroyableComponent->SetAttacksToBlock(std::min(destroyableComponent->GetAttacksToBlock() - 1, static_cast<uint32_t>(0)));
```
Wincent01, line 63: Instead of static cast you can just do ‘0u‘

aronwk-aaron (approved):

Wincent01 (approved):

PR #930, opened 2022-12-29
Fix Hash Collisions in CDBehaviorParameter table

EmosewaMC:
Fixes and issue where the fire sword would deal the incorrect amount of damage.
Reduces hashing time by instead hashing 2 numbers instead of a string and a number.

new limit for the hashes is the uint32 max for both behaviorID and parameterID


Wincent01 (approved):
I thought I was very cleaver when introducing hashing here, didn't check for collisions

aronwk-aaron (approved):

PR #931, opened 2022-12-31
Fix serratorizer chargeup time

EmosewaMC:
The Serratorizers charge up time is limited at 20 as shown [here](https://explorer.squareville.org/skills/behaviors/20933/20999) however its following switch multiples' value 2 is limited at 2, which is not in line with any other valiant weapon which have value 2 at 20.  This migration addresses this issue.

Tested that the serratorizer can be used after charging for longer than 2 seconds

### DIFF
--- a/migrations/cdserver/5_serratorizer_chargeup_fix.sql
+++ b/migrations/cdserver/5_serratorizer_chargeup_fix.sql
@@ -1,0 +1,1 @@
+UPDATE behaviorParameter SET value = 20 WHERE behaviorID = 21001 AND parameterID = "value 2";


aronwk-aaron (approved):

Wincent01 (commented):
Does this change only have to be applied server-side?

EmosewaMC:
> Does this change only have to be applied server-side?

From what I can tell yes.  The client handles the behavior without this update just fine, possibly due to it handling the switch multiple differently than we do

Wincent01 (approved):

PR #934, opened 2022-12-31
Fix Complete Overhaul

EmosewaMC:
I don't like how this is done, but there is literally 1 singular mission that uses this feature, so i implemented this the best I could using the information of that one mission

Tested that you need to swap precisely the 6 possible unique parts in order for the mission to complete and that you have to have put a car into the build first in order to allow the mission to complete.

### DIFF
--- a/dGame/dGameMessages/GameMessages.cpp
+++ b/dGame/dGameMessages/GameMessages.cpp
@@ -28,6 +28,7 @@
 #include "GameConfig.h"
 #include "RocketLaunchLupComponent.h"
 #include "eUnequippableActiveType.h"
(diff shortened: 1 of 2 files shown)


aronwk-aaron (approved):

Wincent01 (approved):
Typical stuff. If you wanted to you could define away the magic number or something, but otherwise this is likely as good as it's going to get.

EmosewaMC:
> Typical stuff. If you wanted to you could define away the magic number or something, but otherwise this is likely as good as it's going to get.

I'll tack that onto xiphos tracking issue of our list of magic numbers to eliminate

PR #935, opened 2023-01-01
Eliminate WorldConfig Based Magic Numbers

EmosewaMC:
Loads the entire WorldConfig into memory and uses a struct to hold all of the world config information.  This eliminates several uses of magic numbers across the codebase and simplifies some ZoneManager variables all into one object.  

Tested that values used by DLU match what they were before, but instead get this value from this new method call.

### DIFF
--- a/dGame/dComponents/DestroyableComponent.cpp
+++ b/dGame/dComponents/DestroyableComponent.cpp
@@ -30,6 +30,7 @@
 #include "PossessorComponent.h"
 #include "InventoryComponent.h"
 #include "dZoneManager.h"
+#include "WorldConfig.h"
 
 DestroyableComponent::DestroyableComponent(Entity* parent) : Component(parent) {
 	m_iArmor = 0;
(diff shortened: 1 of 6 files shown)


EmosewaMC:
Even though many fields are pretty self explanatory, I am choosing to not comment what some fields are due to not being absolutely sure what they are used for, even if the name implies what it may be used for.

aronwk-aaron (dismissed):

dGame/dComponents/DestroyableComponent.cpp
```diff
@@ -765,20 +766,16 @@ void DestroyableComponent::Smash(const LWOOBJID source, const eKillType killType
...
-				}
+				uint64_t coinsToLose = std::max(static_cast<uint64_t>(coinsTotal * coinPercentageToLose), minCoinsToLose);
+				coinsToLose = std::min(maxCoinsToLose, coinsToLose);
 
-				coinsTotal -= coinsToLoose;
+				coinsTotal -= coinsToLose;
 
-				LootGenerator::Instance().DropLoot(m_Parent, m_Parent, -1, coinsToLoose, coinsToLoose);
+				LootGenerator::Instance().DropLoot(m_Parent, m_Parent, -1, coinsToLose, coinsToLose);
 				character->SetCoins(coinsTotal, eLootSourceType::LOOT_SOURCE_PICKUP);
```
Wincent01, line 779: This seems like a bit of a change in logic. Should the player lose a linear amount of coins immediately? Not sure where that initial 200 number came from, but this will act differently from how it currently does. Also as a redundancy check, should the above if-statement be conditioned on `coinsTotal  >= coinsLostOnDeathMin` as to prevent accidental integer overflows if the value is changed.

Wincent01 (dismissed):

dGame/dComponents/DestroyableComponent.cpp
EmosewaMC: That's the problem with a magic number like 200.  We have no idea where that value came from. There should be a floor that you have at least 1 coin to lose yes and ill implement that, but this should ensure that you lose at least 1 coin.  Unless there's a missing worldconfig option, this logic is correct with the floor of coins to lose

EmosewaMC: I've pushed changes to change the if conditional.  Unsure what you mean about the first question, however the existence of the minimum coins to lose to me implies that if you have money > min coins to lose, then you lose at least that amount of currency. Tested that dying with 2 1 and 0 coins nets a loss of 1 1 and 0 coins respectively.

aronwk-aaron (approved):

dGame/dComponents/DestroyableComponent.cpp
EmosewaMC: the 200 check is completely arbitrary.  The minimum dropped coins is based on `coins_lost_on_death_min` as changing this value to 15 in the cdclient.fdb resulted in 15 coins being dropped and having less than 15 coins resulted in the client dropping all the coins the player had.  This can be tested by turning the flag `bSpawnLoot` to true in the SendDie GameMessage and not handling any server coin drop logic at all.

DarwinAnim8or (approved):
Looks like a logical step up from hardcoded values overall

PR #936, opened 2023-01-02
Further implement Property Behavior parsing

EmosewaMC:
The goal here is to implement a robust, modular base class to be used for the implementation of behaviors such that clutter is reduced and very few modifications would be needed to maintain the system in the future.

- [x] Change the name of BEHAVIORSTATE and STRIPID to be BehaviorState and StripId to follow our current PascalCase naming conventions
- [x] Parse the BlocksDef file if it exists and if it doesn't disable behavior editing since cheat detection would be disabled.
- [x] Create classes for all the ControlBehavior messages to allow them to be passed easily in the future
- [x] Implement tests for the split out classes to ensure values are accessed correctly and get the correct values.
- [x] Implement a way to prove that the blocks def parsing can be used and what the use case is.


dGame/dPropertyBehaviors/BehaviorAction.h
```diff
@@ -0,0 +1,83 @@
...
+	  * @param value The new action name
+	  */
+	void SetActionName(std::string& value) { this->actionName = value; };
+
+	/**
+	 * @brief Gets the actions name
+	 *
+	 * @return This actions name
+	 */
+	std::string GetActionName() const { return actionName; };
```
Wincent01, line 36: Could return `const std::string&`

dGame/dPropertyBehaviors/BehaviorAction.h
```diff
@@ -0,0 +1,83 @@
...
+	 * @param value The actions' new name.
+	 */
+	void SetParameterName(std::string& value) { this->parameterName = value; };
+
+	/**
+	 * @brief Gets the parameter name for this action.
+	 *
+	 * @return The actions' name.
+	 */
+	std::string GetParameterName() const { return parameterName; };
```
Wincent01, line 50: Same here

dGame/dPropertyBehaviors/BehaviorAction.h
```diff
@@ -0,0 +1,83 @@
...
+
+	/**
+	 * Unused parameter.  Always serialized as an empty string. __callbackID__
+	 */
+	void SetCallbackID(std::string& value) { this->callbackID = value; };
+
+	/**
+	 * Unused parameter.  Always serialized as an empty string. __callbackID__
+	 */
+	std::string GetCallbackID() const { return callbackID; };
```
Wincent01, line 74: Same here

Wincent01 (approved):
Looks good

EmosewaMC:
holy this became large.  Cutting off features here.

tests/dGameTests/dGameMessagesTests/GameMessageTests.cpp
```diff
@@ -1,15 +1,50 @@
...
+		TearDownDependencies();
+	}
+
+	std::string ReadFromFile(std::string filename) {
+		std::ifstream file(filename, std::ios::binary);
+		std::string readFile;
+		while (file.good()) {
+			char readCharacter = file.get();
+			readFile.push_back(readCharacter);
 		}
```
Xiphoseer, line 37: There are much better ways to turn an ifstream into a string, I'm sure we have at least 5 of them in DLU already

dGame/dPropertyBehaviors/ControlBehaviors.cpp
```diff
@@ -107,292 +92,68 @@ void ModelTypeChanged(AMFArrayValue* arguments, ModelComponent* ModelComponent)
...
-		// modelComponent->AddStrip(stateID, stripID, type, behaviorID, valueParameterName, valueParameterString, valueParameterDouble, "", xPosition, yPosition);
-		type.clear();
-		valueParameterName.clear();
-		valueParameterString.clear();
-		valueParameterDouble = 0.0;
-	}
-	// RequestUpdatedID(behaviorID);
+void ControlBehaviors::AddStrip(AMFArrayValue* arguments) {
+	AddStripMessage addStripMessage;
+	addStripMessage.Parse(arguments);
```
Xiphoseer, line 101: If you call parse immediately on a new instance and don't use it anywhere else, just make it into a constructor or a static factory function

Xiphoseer (approved):

dGame/dPropertyBehaviors/ControlBehaviors.cpp
Xiphoseer: (otherwise you may have some code down the line that calls parse on an already initialized instance and cause heisenbugs if there some old data not properly reset)

EmosewaMC: Would the static factory function take in an AMFArrayValue* and then return a AddStripMessage in this case?

tests/dGameTests/dGameMessagesTests/GameMessageTests.cpp
EmosewaMC: there probably are!  I just don't know where to look since they arent in general utils :P

aronwk-aaron (dismissed):
you're insane, great work!

dGame/dPropertyBehaviors/ControlBehaviors.cpp
Xiphoseer: yes, exactly

Wincent01 (approved):

Wincent01:
I believe all of the getters for strings here can be `const std::string&`

Wincent01 (approved):

aronwk-aaron (approved):

PR #940, opened 2023-01-04
Fix debug logging newline

EmosewaMC:
debug logs properly put a new line at the end of a line


Wincent01 (commented):
It's usually up to the caller to have a new line in their contents. Believe adding this would create unintentional double linebreak. It's not great to be inconsistent with the normal log function either.

aronwk-aaron:
Normal logs were changed to not need a new line at the end several months ago because it was more of a pain without.

EmosewaMC:
No current LogDebug has a default newline in it

aronwk-aaron (approved):

Wincent01 (approved):

PR #941, opened 2023-01-04
Add BubbleStatue Script

EmosewaMC:
Need to find a way to properly ghost this object however the script works as is functionally.


dGame/EntityManager.cpp
```diff
@@ -37,7 +37,10 @@ std::vector<LWOMAPID> EntityManager::m_GhostingExcludedZones = {
 	1250,
 	1251,
 	1350,
-	1450
+	1450,
+
+	// Zorillo Plaza
+	3200
```
Wincent01, line 43: This should not be on DLU main

dGame/dGameMessages/GameMessages.cpp
```diff
@@ -6006,3 +6006,19 @@ void GameMessages::HandleUpdatePlayerStatistic(RakNet::BitStream* inStream, Enti
 		characterComponent->UpdatePlayerStatistic((StatisticID)updateID, (uint64_t)std::max(updateValue, int64_t(0)));
 	}
 }
+
+void GameMessages::SendActivateBubbleBuffFromServer(LWOOBJID objectId, bool specialAnims, std::u16string wszType, const SystemAddress& sysAddr) {
+	CBITSTREAM;
+	CMSGHEADER;
+
+	bitStream.Write(objectId);
+	bitStream.Write<uint16_t>(734);
```
Wincent01, line 6015: Add an entry to the game message enum

Wincent01 (commented):
Wondering if this should be on DLU main. Is the script from live?

aronwk-aaron:
> Wondering if this should be on DLU main. Is the script from live?

Yes, this was shipped with the live game

dGame/EntityManager.cpp
EmosewaMC: Didn't even mean to push that as it was a bodge change.  Fixes

EmosewaMC:
> Wondering if this should be on DLU main. Is the script from live?

As aron said, this is something in the final client version 1.10.64 and like frostburgh, is unusable in its current state simply due to its map being inaccessible, but its object could be spawned serverside, used for objects or used in another map should someone want to do so.

aronwk-aaron (approved):

PR #942, opened 2023-01-04
Implement bubble seriliaztion in controllable physics

aronwk-aaron:
This PR implements the bubbleBuff serialization in controllable physics as well as the activate and deactivate of said buff.
The timer is there to deactivate the bubble buff since the bubble pops itself after ~10 seconds

Added some doc strings to to functions that I didn't document
Added a slash command to help test bubble stuff
Fill out the unknown and correctly name dirty equipeditemInfo


aronwk-aaron:
picking this up to continue work on it

dGame/dComponents/ControllablePhysicsComponent.cpp
```diff
@@ -84,14 +86,23 @@ void ControllablePhysicsComponent::Serialize(RakNet::BitStream* outBitStream, bo
...
+		outBitStream->Write(m_InJetpackMode);
+		m_DirtyEquippedItemInfo = false;
 	}
 
-	outBitStream->Write0();
+	outBitStream->Write(m_BubbleDirty);
+	if (m_BubbleDirty) {
+		outBitStream->Write1(); // another dirty flag?
+		if (m_BubbleDirty) {
+			outBitStream->Write<uint32_t>(1); // 1 not full bubble, 2 full bubble
```
EmosewaMC, line 100: should probably have an enum here?

aronwk-aaron: Yep, agreed. Haven't actually done work in this since jett, just washed to get the branch updated and merge conflicts resolved so I could start easier lol

Wincent01 (approved):

EmosewaMC:
Could you verify if any game message is networked to the server to deactivate the bubble that we perhaps aren't handling?  If no message is networked then a timer makes the most sense.  Would be more ideal to find the true cause behind that timer length, not going to block past the initial question however.

dGame/dComponents/ControllablePhysicsComponent.cpp
```diff
@@ -84,14 +91,22 @@ void ControllablePhysicsComponent::Serialize(RakNet::BitStream* outBitStream, bo
...
 	}
 
-	outBitStream->Write(m_DirtyPickupRadiusScale);
-	if (m_DirtyPickupRadiusScale) {
+	outBitStream->Write(m_DirtyEquippedItemInfo);
+	if (m_DirtyEquippedItemInfo) {
 		outBitStream->Write(m_PickupRadius);
-		outBitStream->Write0(); //No clue what this is so im leaving it false.
-		m_DirtyPickupRadiusScale = false;
+		outBitStream->Write(m_InJetpackMode);
```
EmosewaMC, line 97: Is there any observed changes to jetpack with this change?

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -1300,6 +1301,35 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
...
+			controllablePhysicsComponent->DeactivateBubbleBuff();
+		} else{
+			uint32_t type_intermed;
+			if (!GeneralUtils::TryParse(args[0], type_intermed)) {
+				ChatPackets::SendSystemMessage(sysAddr, u"Invalid bubbleType.");
+				return;
+			}
+			bool floating = true;
+			if (args.size() == 2) {
+				uint32_t type_intermed2;
```
EmosewaMC, line 1320: Do booleans need an intermediate value?  
Following with our style guidelines these should be named typeIntermed as well.

EmosewaMC (approved):
Just a couple questions and a style comment

aronwk-aaron:
Timer was ooga booga smooth brain guessed cause the "skunk" type is referenced in the files and it's about 10 secodnds

dGame/dUtilities/SlashCommandHandler.cpp
aronwk-aaron: yeah, I'll fix that, wasn't going to include the command, but decided to last minute

dGame/dComponents/ControllablePhysicsComponent.cpp
aronwk-aaron: well that value is never changed, soooooo no

aronwk-aaron: Will add functionality in future PR for this variable to be used by the jetpack behavior

dGame/dUtilities/SlashCommandHandler.cpp
aronwk-aaron: To clarify why I used an intermed for floating, it didn't work the way I expected so I just made it more verbose about it to be explict. 
Which I think is fine for a GM command to test this functionality

aronwk-aaron:
> Could you verify if any game message is networked to the server to deactivate the bubble that we perhaps aren't handling? If no message is networked then a timer makes the most sense. Would be more ideal to find the true cause behind that timer length, not going to block past the initial question however.

Good call, didn't even think to see if the client want sending messages, it is infact spamming the server with deactivate bubble buff, will impl

Jettford (dismissed):

EmosewaMC (approved):

EmosewaMC (approved):

Jettford (approved):

PR #945, opened 2023-01-07
Use field names instead of numbers in CDTable reading

EmosewaMC:
Used a horrendous script to do most of this work for me, aside from the tables that were unlike the others, which were done by hand.

Server compiles and boots which means no table fields failed to be read.


aronwk-aaron (approved):

aronwk-aaron:
I watched him commit python war crimes

Xiphoseer (approved):
LGTM

PR #948, opened 2023-01-09
Add note for compiling with multiple jobs

EmosewaMC:
Tested that appending -j4 compiles with 4 jobs at once.


aronwk-aaron:
want to add to the packages to install `libssl-dev zlib1g zlib1g-dev` for linux as well?

EmosewaMC:
> want to add to the packages to install `libssl-dev zlib1g zlib1g-dev` for linux as well?

Those are already listed?

EmosewaMC:
Screen Shot 2023-01-17 at 4 22 17 AM
`zlib1g-dev` is just a more updated version of `zlib` so im more inclined to support that over the standard package unless of course other systems just don't have it, then we can probably add it

aronwk-aaron:
Ah, so not only can users not read, I cannot read!

aronwk-aaron (approved):

Wincent01 (approved):

PR #949, opened 2023-01-10
Take flower if it is in the vault

EmosewaMC:
Tested that the behavior of the script mimics live

### DIFF
--- a/dScripts/02_server/Map/AM/AmTeapotServer.cpp
+++ b/dScripts/02_server/Map/AM/AmTeapotServer.cpp
@@ -1,14 +1,21 @@
 #include "AmTeapotServer.h"
 #include "InventoryComponent.h"
 #include "GameMessages.h"
-
+#include "Item.h"
 
 void AmTeapotServer::OnUse(Entity* self, Entity* user) {
 	auto* inventoryComponent = user->GetComponent<InventoryComponent>();
 	if (!inventoryComponent) return;
 
-	if (inventoryComponent->GetLotCount(BLUE_FLOWER_LEAVES) >= 10) {
-		inventoryComponent->RemoveItem(BLUE_FLOWER_LEAVES, 10);
+	auto* blueFlowerItem = inventoryComponent->FindItemByLot(BLUE_FLOWER_LEAVES, eInventoryType::ITEMS);
+	if (!blueFlowerItem) {
+		blueFlowerItem = inventoryComponent->FindItemByLot(BLUE_FLOWER_LEAVES, eInventoryType::VAULT_ITEMS);
+		if (!blueFlowerItem) return;
+	}
+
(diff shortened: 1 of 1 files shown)


aronwk-aaron (approved):

Wincent01 (commented):
Would the behavior here be that you need a stack of 10 somewhere, and multiple stacks in the same or different inventories are not allowed?

dScripts/02_server/Map/AM/AmTeapotServer.cpp
```diff
@@ -1,14 +1,21 @@
...
-
+#include "Item.h"
 
 void AmTeapotServer::OnUse(Entity* self, Entity* user) {
 	auto* inventoryComponent = user->GetComponent<InventoryComponent>();
 	if (!inventoryComponent) return;
 
-	if (inventoryComponent->GetLotCount(BLUE_FLOWER_LEAVES) >= 10) {
-		inventoryComponent->RemoveItem(BLUE_FLOWER_LEAVES, 10);
+	auto* blueFlowerItem = inventoryComponent->FindItemByLot(BLUE_FLOWER_LEAVES, eInventoryType::ITEMS);
```
Wincent01, line 10: I believe this will always return the smallest stack.

Wincent01: I believe best practice is to not interact with the items themselves and let the inventory component / inventory class handle stuff.

EmosewaMC: The problem is its not looking in the correct spot.  Remove item only looks in the expected inventory of the lot and not in other places said lot may be.  What is the proposed solution for this?

EmosewaMC:
> Would the behavior here be that you need a stack of 10 somewhere, and multiple stacks in the same or different inventories are not allowed?
From my testing if you had 1 in your main inventory, and 10 in your vault, the client would not let you interact with the teapot so this case needs to be accounted for

dScripts/02_server/Map/AM/AmTeapotServer.cpp
EmosewaMC: Should there be a check where if there is some count X left, check the mirrored vault inventory for the remaining items?

EmosewaMC:
This behavior also only seems to happen with the teapot - the Maelstrom brick consoles do not replicate this behavior

dScripts/02_server/Map/AM/AmTeapotServer.cpp
Wincent01: Not sure there's currently a neat way of doing this. One easy solution would be to add an option or another method to the inventory system to look for the biggest stack, alternatively adding even more functionality to the inventory system to be able to handle such a case.

Wincent01 (approved):
Approving this. It is an improvement upon what is presently here.

PR #959, opened 2023-01-17
Don't enable HC on minigame worlds

aronwk-aaron:
Fix max's ouchi

### DIFF
--- a/dGame/EntityManager.cpp
+++ b/dGame/EntityManager.cpp
@@ -70,9 +70,9 @@ void EntityManager::Initialize() {
 	auto hcDropInv = Game::config->GetValue("hardcore_dropinventory_on_death");
 	m_HardcoreDropinventoryOnDeath = hcDropInv.empty() ? false : (hcDropInv == "1");
 
-	// If instance is not zero, then hardcore mode is disabled
+	// If cloneID is not zero, then hardcore mode is disabled
 	// aka minigames and props
-	if (dZoneManager::Instance()->GetZoneID().GetInstanceID() != 0) m_HardcoreMode = false;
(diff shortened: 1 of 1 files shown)


EmosewaMC (dismissed):
Couldn't we just check if cloneid is non zero?  If clone id is zero you should be able to respawn in that world

aronwk-aaron:
> Couldn't we just check if cloneid is non zero?  If clone id is zero you should be able to respawn in that world

Yep, that's a better idea

EmosewaMC (approved):

dGame/EntityManager.cpp
```diff
@@ -69,6 +69,10 @@ void EntityManager::Initialize() {
 	m_HardcoreUscoreEnemiesMultiplier = hcUscoreMult.empty() ? 2 : std::stoi(hcUscoreMult);
 	auto hcDropInv = Game::config->GetValue("hardcore_dropinventory_on_death");
 	m_HardcoreDropinventoryOnDeath = hcDropInv.empty() ? false : (hcDropInv == "1");
+
+	// If instance is not zero, then hardcore mode is disabled
+	// aka minigames and props
+	if (dZoneManager::Instance()->GetZoneID().GetInstanceID() != 0) m_HardcoreMode = false;
```
Wincent01, line 75: Believe instance-ID is different from clone-ID

EmosewaMC: ah yes it is.  Glanced right over that 😅

EmosewaMC (approved):

Wincent01 (approved):
HC could still be enabled in instances like FV Dragons and Frakjaw, but this catches most cases where is would be a problem.

PR #989, opened 2023-02-11
Implement some more trigger event calls and command handlers

aronwk-aaron:
This PR implements some more of the easier to do events triggering and command handling.

Mainly cutting this off where it is since any more would involve more complex changes

Tested things still work as they should, some of these are never used in the live game, but are implemented for potential map makers


dGame/dComponents/TriggerComponent.cpp
```diff
@@ -75,8 +100,13 @@ void TriggerComponent::HandleTriggerCommand(LUTriggers::Command* command, Entity
...
 			case eTriggerCommandType::LOCK_OR_UNLOCK_CONTROLS: break;
-			case eTriggerCommandType::PLAY_EFFECT: break;
-			case eTriggerCommandType::STOP_EFFECT: break;
+			case eTriggerCommandType::PLAY_EFFECT:
+				HandlePlayEffect(targetEntity, argArray);
+				break;
+			case eTriggerCommandType::STOP_EFFECT:
+				GameMessages::SendStopFXEffect(targetEntity, true, command->args);
+				break;
+			// DEPRECATED BLOCK START
```
EmosewaMC, line 109: Thoughts on moving this closer to the default so all the unhandled triggers are in one block?

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -113,22 +160,169 @@ std::vector<Entity*> TriggerComponent::GatherTargets(LUTriggers::Command* comman
...
+		for (const auto memberId : team->members) {
+			auto* member = EntityManager::Instance()->GetEntity(memberId);
+			if (member) entities.push_back(member);
+		}
+	} else if (command->target == "objGroup") entities = EntityManager::Instance()->GetEntitiesInGroup(command->targetName);
+	else if (command->target == "allPlayers") {
+		for (auto* player : Player::GetAllPlayers()) {
+			entities.push_back(player);
+		}
+	} else if (command->target == "allNPCs") { /*UNUSED*/ }
```
EmosewaMC, line 175: Unused in the live client or unused in general?

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -113,22 +160,169 @@ std::vector<Entity*> TriggerComponent::GatherTargets(LUTriggers::Command* comman
...
+
+void TriggerComponent::HandleToggleTrigger(Entity* targetEntity, std::string args){
+	auto* triggerComponent = targetEntity->GetComponent<TriggerComponent>();
+	if (!triggerComponent) return;
+	triggerComponent->SetTriggerEnabled(args == "1");
+}
+
+void TriggerComponent::HandleResetRebuild(Entity* targetEntity, std::string args){
+	auto* rebuildComponent = targetEntity->GetComponent<RebuildComponent>();
+	if (!rebuildComponent) return;
```
EmosewaMC, line 200: These should likely log if they do not find their associated components since we expect them to exist and them not existing would be an issue with initialization

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -113,22 +160,169 @@ std::vector<Entity*> TriggerComponent::GatherTargets(LUTriggers::Command* comman
...
+	auto position = targetEntity->GetPosition();
+
+	NiPoint3 offset = NiPoint3::ZERO;
+	GeneralUtils::TryParse<float>(argArray.at(0), offset.x);
+	GeneralUtils::TryParse<float>(argArray.at(1), offset.y);
+	GeneralUtils::TryParse<float>(argArray.at(2), offset.z);
+
+	position.x += offset.x;
+	position.y += offset.y;
+	position.z += offset.z;
```
EmosewaMC, line 214: `position + offset`
NiPoints have their operator+= overloaded iirc but if they don't it would make more sense to implement that than to have to type these three lines every time

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -113,22 +160,169 @@ std::vector<Entity*> TriggerComponent::GatherTargets(LUTriggers::Command* comman
...
+}
+
+
+void TriggerComponent::HandleRepelObject(Entity* targetEntity, std::string args){
+	auto* phantomPhysicsComponent = m_Parent->GetComponent<PhantomPhysicsComponent>();
+	if (!phantomPhysicsComponent) return;
+	float forceMultiplier;
+	GeneralUtils::TryParse<float>(args, forceMultiplier);
+	phantomPhysicsComponent->SetPhysicsEffectActive(true);
+	phantomPhysicsComponent->SetEffectType(1); // repel/repulse
```
EmosewaMC, line 249: possible enum here?

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -113,22 +160,169 @@ std::vector<Entity*> TriggerComponent::GatherTargets(LUTriggers::Command* comman
...
+	phantomPhysicsComponent->SetDirectionalMultiplier(forceMultiplier);
+
+	auto triggerPos = m_Parent->GetPosition();
+	auto targetPos = targetEntity->GetPosition();
+
+	// normalize the vectors to get the direction
+	auto delta = targetPos - triggerPos;
+	auto abs = sqrtf(delta.x * delta.x + delta.y * delta.y + delta.z * delta.z);
```
EmosewaMC, line 257: `NiPoint::Length();`

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -113,22 +160,169 @@ std::vector<Entity*> TriggerComponent::GatherTargets(LUTriggers::Command* comman
...
+	NiPoint3 direction = NiPoint3::ZERO;
+	direction.x = delta.x / abs;
+	direction.y = delta.y / abs;
+	direction.z = delta.z / abs;
```
EmosewaMC, line 261: `NiPoint3 direction = delta / abs;`

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -113,22 +160,169 @@ std::vector<Entity*> TriggerComponent::GatherTargets(LUTriggers::Command* comman
...
+	float leadIn = -1.0;
+	if (argArray.size() >= 2) GeneralUtils::TryParse<float>(argArray.at(1), leadIn);
+	int wait = 0;
+	if (argArray.size() >= 3 && argArray.at(2) == "wait") wait = 1;
+	bool unlock = true;
+	if (argArray.size() >= 4 && argArray.at(3) == "unlock") unlock = false;
+	bool leaveLocked = false;
+	if (argArray.size() >= 5 && argArray.at(4) == "leavelocked") leaveLocked = true;
+	bool hidePlayer = false;
+	if (argArray.size() >= 6 && argArray.at(5) == "hideplayer") hidePlayer = true;
```
EmosewaMC, line 279: Is this supposed to be an else if block instead so we dont check for elements which are obviously out of range?

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -113,22 +160,169 @@ std::vector<Entity*> TriggerComponent::GatherTargets(LUTriggers::Command* comman
...
+	character->SetBuildMode(buildMode);
+}
+
+void TriggerComponent::HandleUpdateMission(Entity* targetEntity, std::vector<std::string> argArray) {
+	// there are only explore tasks used
+	// If others need to be implemented for modding
+	// then we need a good way to convert this from a string to that enum
+	if (argArray.at(0) != "exploretask") return;
+	MissionComponent* missionComponent = targetEntity->GetComponent<MissionComponent>();
+	if (!missionComponent) return;
```
EmosewaMC, line 299: We should verify that these do not erronously progress missions with the current POI system.

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -113,22 +160,169 @@ std::vector<Entity*> TriggerComponent::GatherTargets(LUTriggers::Command* comman
...
+	missionComponent->Progress(eMissionTaskType::EXPLORE, 0, 0, argArray.at(4));
+}
+
+void TriggerComponent::HandlePlayEffect(Entity* targetEntity, std::vector<std::string> argArray) {
+	// [nameID],[effectID],[effectType],[priority(optional)]
+	int32_t effectID = 0;
```
EmosewaMC, line 305: A bounds check would be good for this vector at the start of the function since we use quite a few indexes in it

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -137,51 +331,72 @@ void TriggerComponent::HandleSetPhysicsVolume(Entity* targetEntity, std::vector<
...
+		spawner->Activate();
+	}
+}
 
-	for (const CDMissionTasks& task : missionTasks) {
-		MissionComponent* missionComponent = targetEntity->GetComponent<MissionComponent>();
-		if (!missionComponent) continue;
+void TriggerComponent::HandleDeactivateSpawnerNetwork(std::string args){
+	for (auto* spawner : dZoneManager::Instance()->GetSpawnersByName(args)) {
+		spawner->Deactivate();
```
EmosewaMC, line 371: should probably check nullptr here

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -137,51 +331,72 @@ void TriggerComponent::HandleSetPhysicsVolume(Entity* targetEntity, std::vector<
...
+	}
+}
 
-		missionComponent->ForceProgress(task.id, task.uid, std::stoi(argArray[2]));
+void TriggerComponent::HandleResetSpawnerNetwork(std::string args){
+	for (auto* spawner : dZoneManager::Instance()->GetSpawnersByName(args)) {
+		spawner->Reset();
```
EmosewaMC, line 377: and here

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -137,51 +331,72 @@ void TriggerComponent::HandleSetPhysicsVolume(Entity* targetEntity, std::vector<
...
+void TriggerComponent::HandleDestroySpawnerNetworkObjects(std::string args){
+	for (auto* spawner : dZoneManager::Instance()->GetSpawnersByName(args)) {
+		for (auto* node : spawner->m_Info.nodes) {
+			for (const auto& element : node->entities) {
+				auto* entity = EntityManager::Instance()->GetEntity(element);
+				if (entity == nullptr) continue;
+				entity->Kill();
+			}
+			node->entities.clear();
+		}
```
EmosewaMC, line 390: Could we not just call Spawner::Reset here?

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -113,22 +160,169 @@ std::vector<Entity*> TriggerComponent::GatherTargets(LUTriggers::Command* comman
...
+	if (argArray.size() >= 5 && argArray.at(4) == "leavelocked") leaveLocked = true;
+	bool hidePlayer = false;
+	if (argArray.size() >= 6 && argArray.at(5) == "hideplayer") hidePlayer = true;
+	GameMessages::SendPlayCinematic(targetEntity->GetObjectID(), GeneralUtils::UTF8ToUTF16(argArray.at(0)), targetEntity->GetSystemAddress(), true, true, false, false, 0, hidePlayer, leadIn, leaveLocked, unlock);
+}
+
+void TriggerComponent::HandleToggleBBB(Entity* targetEntity, std::string args) {
+	auto* character = targetEntity->GetCharacter();
+	if (!character) return;
+	bool buildMode = false;
```
EmosewaMC, line 286: Should this default to false or the current build status?

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -113,22 +160,169 @@ std::vector<Entity*> TriggerComponent::GatherTargets(LUTriggers::Command* comman
...
+void TriggerComponent::HandlePushObject(Entity* targetEntity, std::vector<std::string> argArray){
+	auto* phantomPhysicsComponent = m_Parent->GetComponent<PhantomPhysicsComponent>();
+	if (!phantomPhysicsComponent) return;
+	phantomPhysicsComponent->SetPhysicsEffectActive(true);
+	phantomPhysicsComponent->SetEffectType(0); // push
+	phantomPhysicsComponent->SetDirectionalMultiplier(1);
+	NiPoint3 direction = NiPoint3::ZERO;
+	GeneralUtils::TryParse<float>(argArray.at(0), direction.x);
+	GeneralUtils::TryParse<float>(argArray.at(1), direction.y);
+	GeneralUtils::TryParse<float>(argArray.at(2), direction.z);
```
EmosewaMC, line 236: A great addition to TryParse for the sake of not repeating yourself would be something like this implemented into GeneralUtils as a template specialization for TryParse
```
    template <>
	bool TryParse(const std::string& value, NiPoint3& dst) {
		return TryParse<float>(value.c_str(), dst.x) &&
				TryParse<float>(value.c_str(), dst.y) &&
				TryParse<float>(value.c_str(), dst.z);
	}
```
I will also note, the purpose of TryParse is to fail and return false if the value failed to be parsed.  It may be wise to check that you actually got the A-OK that the value was parsed correctly

dGame/dComponents/TriggerComponent.cpp
aronwk-aaron: In general, through all the files that we have, I have not seen any use of it

dGame/dComponents/TriggerComponent.cpp
EmosewaMC: update to this - the function needs to be inlined instead and ive updated the code above to reflect that

dGame/dComponents/TriggerComponent.cpp
aronwk-aaron: It's still referenced in he client itself, but for as if it was ever used, looks like no

dGame/dComponents/TriggerComponent.cpp
EmosewaMC: another update - ive realized you need to be able to pass more than one value to the function in this specialization which makes this a bit more complicated than I initially thought.  Not sure if this is exactly doable with simple changes as above since the value is a single string and not a vector.

EmosewaMC: Looks like the best way to do this is to just overload the function so it takes 3 parameters and is not a template<>.  Makes more sense than what i suggested in the previous comment 😅

dGame/dComponents/TriggerComponent.cpp
aronwk-aaron: can be. will change

dGame/dComponents/TriggerComponent.cpp
aronwk-aaron: I did check these specifically, nothing changed

dGame/dComponents/TriggerComponent.cpp
aronwk-aaron: yes, should be an enum

dGame/dComponents/TriggerComponent.cpp
aronwk-aaron: either, i have the logic for all cases, but could start with being !CurrentBuildStaus, since that's what I'm doing last

dGame/dComponents/TriggerComponent.cpp
aronwk-aaron: I would say no, since reset does more than that, and there's specifically a reset, vs kill all network objects

EmosewaMC: It may make more sense to make this functionality into a function then so we aren't repeating ourself with the same loop in two places

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -242,51 +249,53 @@ void TriggerComponent::HandlePushObject(Entity* targetEntity, std::vector<std::s
...
 
 	// normalize the vectors to get the direction
 	auto delta = targetPos - triggerPos;
-	auto abs = sqrtf(delta.x * delta.x + delta.y * delta.y + delta.z * delta.z);
-	NiPoint3 direction = NiPoint3::ZERO;
-	direction.x = delta.x / abs;
-	direction.y = delta.y / abs;
-	direction.z = delta.z / abs;
-
+	auto abs = delta.Length();
```
EmosewaMC, line 267: Didn't catch this initially, but this variable name overloads a function with the same name <https://cplusplus.com/reference/cstdlib/abs/>
We could be change this so that if abs is used here in the future, there isn't a collision

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -242,51 +249,53 @@ void TriggerComponent::HandlePushObject(Entity* targetEntity, std::vector<std::s
...
-	if (argArray.size() >= 5 && argArray.at(4) == "leavelocked") leaveLocked = true;
 	bool hidePlayer = false;
-	if (argArray.size() >= 6 && argArray.at(5) == "hideplayer") hidePlayer = true;
-	GameMessages::SendPlayCinematic(targetEntity->GetObjectID(), GeneralUtils::UTF8ToUTF16(argArray.at(0)), targetEntity->GetSystemAddress(), true, true, false, false, 0, hidePlayer, leadIn, leaveLocked, unlock);
+
+	if (argArray.size() >= 2) GeneralUtils::TryParse<float>(argArray.at(1), leadIn);
+	else if (argArray.size() >= 3 && argArray.at(2) == "wait") wait = eEndBehavior::WAIT;
+	else if (argArray.size() >= 4 && argArray.at(3) == "unlock") unlock = false;
+	else if (argArray.size() >= 5 && argArray.at(4) == "leavelocked") leaveLocked = true;
+	else if (argArray.size() >= 6 && argArray.at(5) == "hideplayer") hidePlayer = true;
```
EmosewaMC, line 285: was poor phrasing on my end on what change I was thinking about here - 
I was more thinking of something akin to what we see in the client when reading packets
```
if () {
    if () {
        if () {
        }
    }
}
```
Sorry about that!  Right now this ignores all elements past the first one

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -296,48 +305,54 @@ void TriggerComponent::HandleUpdateMission(Entity* targetEntity, std::vector<std
...
 
 void TriggerComponent::HandlePlayEffect(Entity* targetEntity, std::vector<std::string> argArray) {
-	// [nameID],[effectID],[effectType],[priority(optional)]
+	if (argArray.size() <= 2) return;
 	int32_t effectID = 0;
 	if (!GeneralUtils::TryParse<int32_t>(argArray.at(1), effectID)) return;
 	std::u16string effectType = GeneralUtils::UTF8ToUTF16(argArray.at(2));
 	float priority = 1;
-	GeneralUtils::TryParse<float>(argArray.at(3), priority);
+	if (argArray.size() == 4) GeneralUtils::TryParse<float>(argArray.at(3), priority);
```
EmosewaMC, line 321: == 4 or >= 4?  Would there be a case here for more arguments being passed?
Also the few lines above this, are they intending to access argArray.at(0) and 1?  They check for a size of 2 but then access elements 2 and 3

EmosewaMC (commented):
couple notes

dGame/dComponents/TriggerComponent.cpp
aronwk-aaron: oh yeah, that's my bad too, this code just doesn't work at all

dGame/dComponents/TriggerComponent.cpp
aronwk-aaron: `[nameID],[effectID],[effectType],[priority(optional)]` is the arg format
So the last one is optional, and if `size <= 2` it returns, because there are 3 required params. I can do `size < 3` to make that a bit clearer, just me being weird lol

aronwk-aaron: index 0 is already in the right type so I just put it straight in the GM call, but gotta convert indexes 1 and 2 to the right types, and index 3 as well, if we had 4 args, so the 4th arg would be at the index of 3

EmosewaMC (approved):
lgtm

EmosewaMC (approved):
Looks good to me!

Jettford (dismissed):

EmosewaMC (dismissed):

EmosewaMC (approved):
No, the recent commit was not a typo.

Jettford (approved):

PR #990, opened 2023-02-11
Break out the model component

aronwk-aaron:
into the model, item, and mutable model behaviors


aronwk-aaron:
Moving this into the ECS rewrite #1110

PR #996, opened 2023-02-19
breakout gmlevel into a scoped enum

aronwk-aaron:
tested that things still work
slash command,
chat restrictions,
packets and serializations

PR #997, opened 2023-02-19
breakout object bits into scoped enum

aronwk-aaron:
tested that things still work as expected
Waiting on #996 since there's likely to be merge conflicts due to includes, thanks git


dCommon/GeneralUtils.cpp
```diff
@@ -202,18 +202,18 @@ bool GeneralUtils::CaseInsensitiveStringCompare(const std::string& a, const std:
...
+int64_t GeneralUtils::ClearBit(int64_t value, eObjectBits index) {
+	return value &= ~(1ULL << static_cast<uint32_t>(index));
 }
 
 //! Checks a specific bit in a signed 64-bit integer
-bool GeneralUtils::CheckBit(int64_t value, uint32_t index) {
-	return value & (1ULL << index);
+bool GeneralUtils::CheckBit(int64_t value, eObjectBits index) {
+	return value & (1ULL << static_cast<uint32_t>(index));
 }
```
aronwk-aaron, line 217: Real question: could I avoid this static cast with an operator overload

EmosewaMC: I may have an idea to alleviate these

EmosewaMC (approved):
lgtm

EmosewaMC (dismissed):
lgtm maybe inline in the future since the function is small

EmosewaMC (approved):

Jettford (dismissed):

EmosewaMC (approved):
lol

DarwinAnim8or (commented):
While I'm not going to consider this as *blocking*, note that as per the CONTRIBUTING.md file, you should use "#pragma once" and not C-style header guards

DarwinAnim8or (approved):

EmosewaMC:
> While I'm not going to consider this as _blocking_, note that as per the CONTRIBUTING.md file, you should use "#pragma once" and not C-style header guards

Pre compiled headers cannot use pragma once.  There are issues with pragma once which require standard include guards when pre compiled headers are used.

DarwinAnim8or:
> 

Looks like it's a GCC thing? "that they have no plan to add at the present" 
🙃

EmosewaMC (approved):
4 merge conflicts later

PR #1003, opened 2023-02-25
Fix stuns with mast teleport

### LINKED ISSUE #706: BUG: Taking hits forces you out of Pirate-only swing quickbuild animation

### DarkflameServer Version

d2b05a1

### Issue Description

When hit after building a pirate only quickbuild, it forces the character out of the animation of swinging, however you are still sent to the top of the cliff.

https://user-images.githubusercontent.com/48499798/182682367-5e38e816-8050-4397-a96e-20a29ef1ab48.mp4

### Reproduction Steps

go to the pirate only quickbuild in gnarled forest near the elephant escarpment where the flags are on the cliff, and with the stromling pirates aggroed, build the quickbuild.

### Expected Behavior

There is footage here https://youtu.be/fII2nP02_P0?t=958 but I am unsure if they didn't aggro over there in live or they were just distracted by the other player. Even if they did go there, my assumption is that you would get fully knocked out of the swing, or at least make the swing without being knocked out of the animation.

### Environment

ubuntu server

EmosewaMC:
Fixes #706 

Tested that the animation is not interrupted while the player is in the animation.  Tested that the player can correctly take damage afterwards.
Also correct the animation priority as the script from live used 4.0f and not 1.0f

### DIFF
--- a/dScripts/02_server/Map/GF/MastTeleport.cpp
+++ b/dScripts/02_server/Map/GF/MastTeleport.cpp
@@ -2,6 +2,7 @@
 #include "EntityManager.h"
 #include "GameMessages.h"
 #include "Preconditions.h"
+#include "DestroyableComponent.h"
 
 #ifdef _WIN32
 #define _USE_MATH_DEFINES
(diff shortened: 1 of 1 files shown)


EmosewaMC:
I will add this also now matches the live script

aronwk-aaron (approved):

Wincent01 (approved):

PR #1008, opened 2023-03-02
Fix non-parallel timers in CombatAIComponent

### LINKED ISSUE #603: BUG: Incorrect Assembly Engineer Turret Behavior

### DarkflameServer Version

Server version commit: 94e161df55ed3bf28e4c71c0758299540247ba84

### Issue Description

At present, the Assembly Engineer Rank 2 turret outperforms the Rank 3 turret in DPS. The Rank 3 turret has a rate of fire faster than that of the Rank 1 turret but slower than that of the Rank 2 turret, meaning the Rank 2 and Rank 3 behaviors are switched. The Rank 2 turret currently fires at around one shot per half-second, which, while it cannot be verified against any live footage I know of, aligns with information about the Rank 3 turret on the LU wiki circa 2012.

image

https://user-images.githubusercontent.com/108308609/176054492-f66cd349-09ae-45be-9a4e-e18bbf4decd8.mp4

https://user-images.githubusercontent.com/108308609/176054347-90d7d1ff-b6d0-4529-b7ef-ede5bd56bed6.mp4

https://user-images.githubusercontent.com/108308609/176054322-d206cf03-4c5b-4c5b-b7fc-7456777dba12.mp4

### Reproduction Steps

Spawn an Engineer Rank 2 and/or Rank 3 turret.

### Expected Behavior

Rank 3 turrets should exhibit the behavior currently shown by Rank 2 turrets, and vice versa.

### Environment

Client: Windows 10 + unpacked client
Server: Google Cloud, Debian 11

EmosewaMC:
Fixes #603 

Still needs to be tested for functionality.

Make a few changes to the CombatAIComponent:
- Simplify the skill timer decrement loop and use a range-based for loop with references.
- Now a skill is only attempted to be cast if the entity actually has a skill to cast.  This removes the need for the downtime completely since now we will, again, only attempt to cast a skill if we arent stunned or currently in a skill.
- Perf is identical to before, with no change, however now instead of entities having the hardcoded 0.5 second cooldown between skills at a minimum, the skills are able to be cast as soon as they are available

The Engineer Rank 2 and 3 turrets now properly shoot at their correct cooldowns.
AI Across the board now use their skills at the correct cooldown

### DIFF
--- a/dCommon/Metrics.cpp
+++ b/dCommon/Metrics.cpp
@@ -14,6 +14,7 @@ std::vector<MetricVariable> Metrics::m_Variables = {
 	MetricVariable::CPUTime,
 	MetricVariable::Sleep,
 	MetricVariable::Frame,
+	MetricVariable::CombatAiComponentUpdateLoop,
 };
 
 void Metrics::AddMeasurement(MetricVariable variable, int64_t value) {
--- a/dCommon/Metrics.cpp
+++ b/dCommon/Metrics.cpp
@@ -14,7 +14,6 @@ std::vector<MetricVariable> Metrics::m_Variables = {
 	MetricVariable::CPUTime,
 	MetricVariable::Sleep,
 	MetricVariable::Frame,
-	MetricVariable::CombatAiComponentUpdateLoop,
 };
 
 void Metrics::AddMeasurement(MetricVariable variable, int64_t value) {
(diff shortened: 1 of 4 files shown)


EmosewaMC:
leaving my commit about the metrics in there so i can reference it in the future.

EmosewaMC:
Played around Crux for about 30 minutes, fighting stuff with the Serratorizer and The Wormholer and AI seemed to overall be a bit faster as far as casting skills goes since they now don't have a 0.5 second delay added on top of their already existing cooldown.  The Engineer turrets feel amazing now and no longer feel like a waste to use.

aronwk-aaron (approved):

Wincent01 (approved):

PR #1010, opened 2023-03-03
Fix Trigger Missions

EmosewaMC:
Fix trigger missions as previously they were all compared as lowercase, but the code was change to be uppercase.  To keep consistent with the value in the mission and the value in the lutriggers file

Tested that Discover the Orange Path, Cobwebs!, and Insect Music all progressed correctly (they are all trigger progression)


aronwk-aaron (approved):

MustacheMan2:
Are there currently known issues with these trigger missions? Having issues with the orange path and others on the latest commit.

EmosewaMC:
Yes, this addresses those issues with trigger missions.  The monument is weird where 1 of the paths uses triggers and the other 2 use POI. @MustacheMan2

Wincent01 (approved):

PR #1011, opened 2023-03-04
fix compatibility with nexus dash

### LINKED ISSUE #1009: BUG: Adding an admin user with MasterServer -a fails

### DarkflameServer Version

b6fc959

### Issue Description

The app gives a crash dump, but as far as I can tell this issue is actually SQL related. I enabled the mysql query log and when I try running the command manually, this happens:

```
MariaDB [darkflame]> INSERT INTO accounts (name, password, gm_level) VALUES ('admin', '<REDACTED>', 9);
ERROR 1452 (23000): Cannot add or update a child row: a foreign key constraint fails (`darkflame`.`accounts`, CONSTRAINT `accounts_ibfk_1` FOREIGN KEY (`play_key_id`) REFERENCES `play_keys` (`id`) ON DELETE CASCADE)
```

### Reproduction Steps

Run `MasterServer -a` and try to add a user.

### Expected Behavior

A new admin user is created.

### Environment

Ubuntu 22.04 (on a VM, not WSL)

aronwk-aaron:
add some more utility to the masterserver -a command
fixes #1009

### DIFF
--- a/dMasterServer/MasterServer.cpp
+++ b/dMasterServer/MasterServer.cpp
@@ -217,7 +217,7 @@ int main(int argc, char** argv) {
 		std::unique_ptr<sql::ResultSet> res(userLookupStatement->executeQuery());
 		if (res->rowsCount() > 0) {
 			Game::logger->Log("MasterServer", "Account with name \"%s\" already exists", username.c_str());
-			std::cout << "Do you want to change the password to the one that you input? [y/n]?";
+			std::cout << "Do you want to change the password of that account? [y/n]?";
 			std::string prompt = "";
 			std::cin >> prompt;
 			if (prompt == "y" || prompt == "yes"){
--- a/dMasterServer/MasterServer.cpp
+++ b/dMasterServer/MasterServer.cpp
@@ -283,6 +283,7 @@ int main(int argc, char** argv) {
 			statement->execute();
 		} catch(sql::SQLException& e) {
 			Game::logger->Log("MasterServer", "A SQL error occurred!:\n %s",  e.what());
+			return EXIT_FAILURE;
 		}
 
 		Game::logger->Log("MasterServer", "Account created successfully!");
--- a/dMasterServer/MasterServer.cpp
+++ b/dMasterServer/MasterServer.cpp
@@ -212,21 +212,12 @@ int main(int argc, char** argv) {
 		std::cout << "Enter a username: ";
 		std::cin >> username;
 
-		//Read the password from the console without echoing it.
-#ifdef __linux__
-		//This function is obsolete, but it only meant to be used by the
-		//sysadmin to create their first account.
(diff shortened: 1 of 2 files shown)


Wincent01 (changes requested):
Would it be more intuitive to check for an existing user before taking the password?

Maybe this password change functionality should be limited to the operator user (GM=9). This would be to impose good practice, not have an admin change the passwords of the users for them. I've also thought about limiting the use of this command line option to only work for the first operator. There at least were some tutorials out there just after release which told folks to use -a to create accounts, which isn't what it should be used for.

aronwk-aaron:
> Would it be more intuitive to check for an existing user before taking the password?
> 
> Maybe this password change functionality should be limited to the operator user (GM=9). This would be to impose good practice, not have an admin change the passwords of the users for them. I've also thought about limiting the use of this command line option to only work for the first operator. There at least were some tutorials out there just after release which told folks to use -a to create accounts, which isn't what it should be used for.

Some good points. Not having to use a whole admin dashboard for a 1-2 player local server is kinda the biggest advocate for this being able to operate on multiple accounts, maybe only making a gm9 account if no other accounts exist, else make a GM 0 account.
But being able to reset any account's passwords is useful since I've seen people forget their passwords frequently. So having minimal account management tools in the server would be great for a single player or LAN server.

Taking in the password once at the beginning is just me being lazy and not duplicating more code, will fix

aronwk-aaron:
going to scope this down to just fixing compatibility with nexus dash, and improvements or locking down the -a command can be done separately

Wincent01 (approved):

Wincent01:
The command doesn't have to be locked down. I can see it being useful for LAN setups as you say.

dMasterServer/MasterServer.cpp
```diff
@@ -212,43 +212,80 @@ int main(int argc, char** argv) {
...
 		//Create account
+		try {
+			std::unique_ptr<sql::PreparedStatement> statement(Database::CreatePreppedStmt("INSERT INTO accounts (name, password, gm_level) VALUES (?, ?, ?);"));
+			statement->setString(1, username.c_str());
+			statement->setString(2, std::string(hash, BCRYPT_HASHSIZE).c_str());
+			statement->setInt(3, 9);
+			statement->execute();
+		} catch(sql::SQLException& e) {
+			Game::logger->Log("MasterServer", "A SQL error occurred!:\n %s",  e.what());
+		}
```
EmosewaMC, line 286: This catch will output
```
error
account created successfully
```
Should probably move the success statement to below execute so it only gets output on a true success.

EmosewaMC (dismissed):

EmosewaMC (approved):

Wincent01 (approved):

PR #1014, opened 2023-03-06
Fix zombie bug

EmosewaMC:
Fixes an issue where the player could get stuck in a zombie state with zero health.  This could be caused by dying while the player was respawning, which would allow the player to roam the world without a care in the world.

Tested that the players who die while respawning remain dead (0 hp, armor and imagination) until the respawn animation has finished.  This is when the client is ready to let the player die.


dScripts/ai/ACT/ActPlayerDeathTrigger.cpp
```diff
@@ -1,5 +1,7 @@
 #include "ActPlayerDeathTrigger.h"
 
+#include "Entity.h"
```
aronwk-aaron, line 3: Is this needed?

aronwk-aaron (approved):

dScripts/ai/ACT/ActPlayerDeathTrigger.cpp
EmosewaMC: Prevents some IDE errors when viewing the file

Wincent01 (approved):

PR #1019, opened 2023-03-13
Correct incorrect inventory types

EmosewaMC:
Reverse Engineering in the client has revealed that this is the precise correct inventory type mapping for items by item type.

Will test that LUP_MODELS appear in the models inventory as opposed to items


EmosewaMC:
There are no lup models and currently pets are hard coded to go to models, so this is effectively a no-op

aronwk-aaron (approved):

DarwinAnim8or (approved):

PR #1023, opened 2023-03-16
CDClient cleanup and optimization

EmosewaMC:
- Use static function to get table name
- Remove unused GetName function
- Replace above function with a static GetTableName function
- Remove verbose comments
- Remove verbose initializers
- Remove need to specify table name when getting a table by name
- Remove unused typedef for mac and linux
- Pre-allocate space in the buckets for the tables
- Speedups to come next

Tested that the server compiles and runs.  The table code was all parsed manually whereas code outside of dDatabase was replaced with find and replaced so that should be proofread to make sure no erroneous pointers were created


Wincent01 (approved):
Looks good. An additional optimization, although a small one, would be to use an enum and eliminate table names in the table-map, or eliminate that map in favor of singletons.

EmosewaMC:
> Looks good. An additional optimization, although a small one, would be to use an enum and eliminate table names in the table-map, or eliminate that map in favor of singletons.

Now thinking about it, singletons for each table makes much more sense.  After all, you are always looking for a specific table.

DarwinAnim8or (commented):
From what I can tell here, this is effectively using std::map for entries like some tables already did, and updating the tables to all be Singletons inside of the CDManager?

DarwinAnim8or (approved):
Forgot to approve-- Curious to see how you'll improve it further

EmosewaMC:
> Forgot to approve-- Curious to see how you'll improve it further

yes, right now this just removes the main map and all sub tables remain unchanged functionality wise so far.
There are many more changes planned, but I went for getting the tables to a smaller foot print first.  Functionality of the tables themselves is unchanged, but will be improved in the future.  i have some ideas to reduce foot print overall and possibly reduce memory usage by implementing a Least Recently Used data structure to unload data that is not used after some time, as well as some ideas on how to cut lookup times to be constant in some cases rather than log(n) or nlog(n)

PR #1025, opened 2023-03-21
Implement animation table 

EmosewaMC:
Edit: all of my goals were squashed and I ended up implementing a completely different feature.

Implemented the Animations table and allow for 100% live accurate animation look up and operation.
I have also implemented some helper template classes to assist with future CDClient optimizations as well as a debug assert so we can leave assertions in the code that are optimized out without being in debug mode.  All animations now call a static member in RenderComponent to play an animation which does all the heavy lifting of actually playing said animation and looking up which one to play and for how long.  

Regression tests:
Tested that the Gorilla, Dragon and pirate swing animations play correctly and as expected.
Tested that the Ninjago Lanterns played their animations correctly.
Tested that LUP worlds and teleports show the proper animation alongside crux-> nexus tower and vice versa.
The anchor now logs a warning on the server alongside the client as it did in live due to the anchor not having an existing `smash` animation.  This is live accurate.  As per keeping with live standards, even if an animation is known not to be present on the server, a message is networked to play the animation to keep with live standards

New tests:
Tested that the correct animation time is chosen for teleporting to and from the paradox areas.


EmosewaMC:
I am not sure if this logic makes more sense in RenderComponent (where it is in the client) or if it should be a static method on an Entity that passes it to the RenderComponent.  Open to thoughts on where this should go.

aronwk-aaron:
> I am not sure if this logic makes more sense in RenderComponent (where it is in the client) or if it should be a static method on an Entity that passes it to the RenderComponent. Open to thoughts on where this should go.

I'm down for either. 
I've thought about adding static methods to the entity to handle things that would simplify things like getting the component and checking if it exist so that we don't have to do that in every spot of code where we use a component

aronwk-aaron (approved):
LGTM, always happy to get run of the need for magic numbers

DarwinAnim8or (approved):

DarwinAnim8or:
don't forget to resolve your conflicts

aronwk-aaron (approved):

EmosewaMC (commented):
t

aronwk-aaron (approved):

DarwinAnim8or (approved):

PR #1026, opened 2023-03-23
Allow name billboards to be toggled

EmosewaMC:
Allows Developer GMs to turn their nameplates above their head off to others, or civilians to do the same if a config option is enabled.

Tested that using the command properly turns the nameplate off and on, and that when turning it back on it is only visible to other players and not the owning player.
Tested that another player who travelled worlds still sees the billboard if they were
present when the original toggle was run, travelled worlds, travelled back and then the toggle was run again.

https://www.youtube.com/watch?v=6lCTYuvRN_4
Attached is a video of the feature working.


dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -171,6 +171,28 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
...
+			ChatPackets::SendSystemMessage(sysAddr, u"Your nameplate is now on and visible to all players.");
+			character->SetBillboardVisible(true);
+			// The GameMessage we send for turning the nameplate off just deletes the BillboardSubcomponent from the parent component.
+			// Because that same message does not allow for custom parameters, we need to create the BillboardSubcomponent a different way
+			// This workaround involves sending an unrelated GameMessage that does not apply to player entites, 
+			// but forces the client to create the necessary SubComponent that controls the billboard.
+			GameMessages::SendShowBillboardInteractIcon(UNASSIGNED_SYSTEM_ADDRESS, entity->GetObjectID());
+
+			// Now turn off the billboard for the owner.
+			GameMessages::SendSetNamebillboardState(sysAddr, entity->GetObjectID());
```
Wincent01, line 191: Would it make more sense to move the logic behind sending the GMs to the setter?

Wincent01 (dismissed):
Looks good. One small style thing which you are free to ignore.

Maybe now lcdr can finally have a normal player as their vanity NPC like they've been wanting :P

dGame/dUtilities/SlashCommandHandler.cpp
EmosewaMC: Yes it would.  I'll go ahead and move this!

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -171,6 +171,19 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 	}
 #endif
 
+	if (chatCommand == "togglenameplate" && (Game::config->GetValue("allownameplateoff") == "1" || entity->GetGMLevel() > GAME_MASTER_LEVEL_DEVELOPER)) {
```
aronwk-aaron, line 174: :smiling_face_with_tear: I'll need to fix this in #996 Note to self also to Emo whow will have to re-review that RM

aronwk-aaron (approved):

PR #1031, opened 2023-03-27
Implement Timer handlers in triggers and ontimerdone trigger event

aronwk-aaron:
This PR implements the timer command handlers (settimer and canceltimer) as well as the ontimerdone event
Additionally, the AGJetEffect Server has been cleaned up and made 100% live accurate, which is the main reason for adding the timer trigger stuff since the beacons use them
Tested that it works properly and doesn't crash.

After delving into the triggers more, the Jeteffect is the only place it was used in Live, the rest are remnants from older versions of worlds

### DIFF
--- a/dGame/dComponents/SkillComponent.cpp
+++ b/dGame/dComponents/SkillComponent.cpp
@@ -36,8 +36,6 @@ bool SkillComponent::CastPlayerSkill(const uint32_t behaviorId, const uint32_t s
 
 	context->caster = m_Parent->GetObjectID();
 
-	context->skillID = skillID;
-
 	this->m_managedBehaviors.insert_or_assign(skillUid, context);
 
 	auto* behavior = Behavior::CreateBehavior(behaviorId);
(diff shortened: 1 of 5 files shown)


aronwk-aaron:
Note: this does NOT fix the cast skill of jet beacons. that is a separate issue with AOE that will be addressed in a separate PR

dGame/dComponents/SkillComponent.cpp
```diff
@@ -36,8 +36,6 @@ bool SkillComponent::CastPlayerSkill(const uint32_t behaviorId, const uint32_t s
 
 	context->caster = m_Parent->GetObjectID();
 
-	context->skillID = skillID;
```
EmosewaMC, line 39: This needs to be kept.  It will break missions if this is removed here

dGame/dComponents/SkillComponent.cpp
```diff
@@ -248,6 +246,8 @@ SkillExecutionResult SkillComponent::CalculateBehavior(const uint32_t skillId, c
 
 	context->caster = m_Parent->GetObjectID();
 
+	context->skillID = skillId;
```
EmosewaMC, line 249: Keep this in though.  Since most of the time skillID was only needed for client casts I ignored setting it here.

EmosewaMC (changes requested):

DarwinAnim8or (approved):

EmosewaMC (approved):

PR #1032, opened 2023-03-27
Fix Wingreaper birds not moving

### LINKED ISSUE #1030: BUG: Wingreaper owls do not move

### DarkflameServer Version

72ca0f1

### Issue Description

Wingreaper owls no longer move.

### Reproduction Steps

Use the Assembly Summoner's Wingreaper owl ability. The owls can attack but no longer fly around.

### Expected Behavior

they should fly up to nearby enemies (though they teleported previously).

### Environment

ubuntu server

EmosewaMC:
Fixes #1030 

The client seems to do the following for getting the speed from the ControllablePhysics table (of which there is only ever 1 fetch of this field and it is in a ControllablePhysicsComponent):
Default speed set to 10.0f
Check the PhysicsComponent table for the column speed and if it exists set speed to that value and if the value was null set it to the default again.

Tested that the birds of the Wingreaper now properly move towards their enemy.  Tested that the dragons in Forbidden Valley dragon battle still do not move (since they have null speed too).  Tested that enemies with specifically zero speed remain at zero speed as set in their PhysicsComponent entry.

### DIFF
--- a/dGame/dComponents/MovementAIComponent.cpp
+++ b/dGame/dComponents/MovementAIComponent.cpp
@@ -307,13 +307,12 @@ float MovementAIComponent::GetBaseSpeed(LOT lot) {
 
 foundComponent:
 
-	float speed;
+	// Client defaults speed to 10 and if the speed is also null in the table, it defaults to 10.
+	float speed = 10.0f;
 
-	if (physicsComponent == nullptr) {
-		speed = 8;
-	} else {
-		speed = physicsComponent->speed;
-	}
+	if (physicsComponent) speed = physicsComponent->speed;
+
+	if (speed == -1.0f) speed = 10.0f;
 
 	m_PhysicsSpeedCache[lot] = speed;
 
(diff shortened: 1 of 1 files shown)


dGame/dComponents/MovementAIComponent.cpp
```diff
@@ -307,13 +307,12 @@ float MovementAIComponent::GetBaseSpeed(LOT lot) {
...
+	float speed = 10.0f;
 
-	if (physicsComponent == nullptr) {
-		speed = 8;
-	} else {
-		speed = physicsComponent->speed;
-	}
+	if (physicsComponent) speed = physicsComponent->speed;
+
+	if (speed == -1.0f) speed = 10.0f;
```
Wincent01, line 315: Not done regularly throughout the codebase, but an abs check would be best.

Wincent01 (approved):
Good! One non-blocking note.

aronwk-aaron (dismissed):

aronwk-aaron (approved):

aronwk-aaron (approved):

DarwinAnim8or (approved):

PR #1035, opened 2023-03-29
refactor:  re-write AOE, add FilterTargets, Update TacArc Reading

aronwk-aaron:
Re-write AOE behavior for new filter targets
Update Tacarc to use new filter targets
Added dev commands for skill and attack debugging

Tested that changes to tac arc only affect filtered enemies.
Removed hardcoded logic to add faction to the includeFactionsList for lot 14466
Tested that the skill from 14466 will hit players (Rain of Fire attack in the spider boss battle).

Tested that AOE filter and sorting targets works as expected given the filter parameters
Targets are hit from closest to furthest up to the max target count.


dGame/dBehaviors/BehaviorContext.cpp
```diff
@@ -289,46 +290,74 @@ void BehaviorContext::Reset() {
...
-	if (!ignoreFaction && !includeFaction) {
-		for (auto entry : entity->GetTargetsInPhantom()) {
-			auto* instance = EntityManager::Instance()->GetEntity(entry);
-
-			if (instance == nullptr) {
-				continue;
+	auto* casterDestroyableComponent = caster->GetComponent<DestroyableComponent>();
+	if (!casterDestroyableComponent) return targets;
+
+	auto candidates = EntityManager::Instance()->GetEntitiesByComponent(eReplicaComponentType::DESTROYABLE);
```
aronwk-aaron, line 310: This should actually be done by getting all entities in the radius in the AOE component (capping out at a radius of 1000.0f) and then filter on those targets, but this is how DLU was doing it before and it wasn't visible affecting performance.
Will probably fix this later, but it's fine as it for now

aronwk-aaron:
Converting to draft to re-write using the logic the client does with passing one vector as a reference to filter and the list being populated outside of that filter.
will save on memory, potentially instead of having multiple vectors of entity pointers just sitting in memory

aronwk-aaron:
re-tested everything listed int the original comment

dGame/dBehaviors/AreaOfEffectBehavior.cpp
```diff
@@ -20,134 +20,123 @@ void AreaOfEffectBehavior::Handle(BehaviorContext* context, RakNet::BitStream* b
...
+			bitStream->Write(target->GetObjectID());
 		}
 
-		const auto distance = Vector3::DistanceSquared(reference, entity->GetPosition());
-
-		if (this->m_radius * this->m_radius >= distance && (this->m_maxTargets == 0 || targets.size() < this->m_maxTargets)) {
-			targets.push_back(entity);
+		// then cast all the actions
+		for (auto* target : targets) {
+			bitStream->Write(target->GetObjectID());
```
EmosewaMC: Extra BitStream write here.  This will not work

dGame/dBehaviors/AreaOfEffectBehavior.cpp
```diff
@@ -20,134 +20,123 @@ void AreaOfEffectBehavior::Handle(BehaviorContext* context, RakNet::BitStream* b
...
 
+	if (this->m_useTargetPosition){
+		Game::logger->Log("AreaOfEffectBehavior", "target is %llu", branch.target);
+		if (branch.target == LWOOBJID_EMPTY)
+		return;
+	}
+
+	if (targetCount == 0){
+		Game::logger->Log("AreaOfEffectBehavior", "no targets");
+		PlayFx(u"miss", context->originator);
```
EmosewaMC, line 26: Now thinking about it, the client handles what FX to play in a handle.  Possibly in calculate too.  Should double check this to make sure we arent doubling up on FX

dGame/dBehaviors/BehaviorContext.cpp
```diff
@@ -289,46 +290,102 @@ void BehaviorContext::Reset() {
...
-		}
+// some basic checks as well as the check that matters for this: if the quickbuild is complete
+bool BehaviorContext::CheckTargetingRequirements(const Entity* target) const {
+	// if the target is a nullptr, then it's not valid
+	if (!target) return false;
 
-		auto entities = EntityManager::Instance()->GetEntitiesByComponent(eReplicaComponentType::CONTROLLABLE_PHYSICS);
-		for (auto* candidate : entities) {
-			const auto id = candidate->GetObjectID();
+	// only target quickbuilds in the are completed
```
EmosewaMC, line 371: Comment needs to be updated

dGame/dBehaviors/BehaviorContext.cpp
```diff
@@ -289,46 +290,102 @@ void BehaviorContext::Reset() {
...
+						}
+						index++;
+					}
+				}
 			}
-
-			targets.push_back(entry);
+			continue;
 		}
 	}
```
EmosewaMC, line 362: Is there any way we can reduce the nesting here?  This goes quite deep and I think we could probably make this look and read much easier

docs/Commands.md
```diff
@@ -106,6 +106,11 @@ These commands are primarily for development and testing. The usage of many of t
 |Set Level|`/setlevel <requested_level> (username)`|Sets the using entities level to the requested level.  Takes an optional parameter of an in-game players username to set the level of.|8|
 |crash|`/crash`|Crashes the server.|9|
 |rollloot|`/rollloot <loot matrix index> <item id> <amount>`|Rolls loot matrix.|9|
+|castskill|`/castskill <skill id>`|Casts the skill as the player|9|
+|setskillslot|`/setskillslot <slot> <skill id>`||9|
+|setfaction|`/setfaction <faction id>`|Clears the users current factions and sets it|9|
+|addfaction|`/addfaction <faction id>`|Add the faction to the users list of factions|9|
+|getfactions|`/getfactions`|Shows the player's factions|9|
```
EmosewaMC, line 113: 9 is operator, 8 is developer

docs/Commands.md
```diff
@@ -120,6 +125,7 @@ Finds the closest entity with the given component or LDF variable (ignoring play
 * `-s`: Prints the entity's settings and spawner ID.
 * `-p`: Prints the entity's position
 * `-f`: If the entity has a destroyable component, prints whether the entity is smashable and its friendly and enemy faction IDs; if `faction` is specified, adds that faction to the entity.
+* `-cf`: check if the faction of this entity is your enenmy or friend
```
EmosewaMC, line 128: enemny
also this could be clarified to entity as this would refer to self when inspect only ever refers something else

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -1875,6 +1875,88 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
...
+			if (!GeneralUtils::TryParse(args[0], slot)) {
+				ChatPackets::SendSystemMessage(sysAddr, u"Error getting slot.");
+				return;
+			} else {
+				Game::logger->Log("SlashCommandHandler", "Got slot %i", slot);
+				if (!GeneralUtils::TryParse(args[1], skillId)) {
+					ChatPackets::SendSystemMessage(sysAddr, u"Error getting skill.");
+					return;
+				} else {
+					Game::logger->Log("SlashCommandHandler", "Got skill %i", skillId);
```
EmosewaMC, line 1907: Unneeded logs

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -1875,6 +1875,88 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 		ChatPackets::SendSystemMessage(sysAddr, u"Deleted inventory " + GeneralUtils::UTF8ToUTF16(args[0]));
 	}
 
+	if (chatCommand == "castskill" && entity->GetGMLevel() >= eGameMasterLevel::DEVELOPER && args.size() >= 1) {
+		auto* skillComponent = entity->GetComponent<SkillComponent>();
+		if (skillComponent){
+			uint32_t skillId;
+
+			if (!GeneralUtils::TryParse(args[0], skillId)) {
```
EmosewaMC, line 1883: .at(0)

dGame/dBehaviors/BehaviorContext.h
```diff
@@ -104,7 +105,11 @@ struct BehaviorContext
 
 	void Reset();
 
-	std::vector<LWOOBJID> GetValidTargets(int32_t ignoreFaction = 0, int32_t includeFaction = 0, const bool targetSelf = false, const bool targetEnemy = true, const bool targetFriend = false) const;
+	void FilterTargets(std::vector<Entity*>& targetsReference, std::forward_list<int32_t> ignoreFaction = {}, std::forward_list<int32_t> includeFaction = {}, const bool targetSelf = false, const bool targetEnemy = true, const bool targetFriend = false, const bool targetTeam = false) const;
```
EmosewaMC, line 108: can we not pass those forward_lists by reference?

dGame/dBehaviors/BehaviorContext.h
```diff
@@ -104,7 +105,11 @@ struct BehaviorContext
...
+
+	bool CheckTargetingRequirements(const Entity* target) const;
+
+	bool CheckFactionList(std::forward_list<int32_t> factionList, std::vector<int32_t> objectsFactions) const;
```
EmosewaMC, line 112: Same here

dGame/dBehaviors/TacArcBehavior.h
```diff
@@ -2,56 +2,43 @@
 #include "Behavior.h"
 #include "dCommonVars.h"
 #include "NiPoint3.h"
+#include <forward_list>
 
-class TacArcBehavior final : public Behavior
-{
+class TacArcBehavior final : public Behavior {
 public:
+	float m_maxRange;
```
EmosewaMC, line 9: These all should be private

dGame/dBehaviors/AreaOfEffectBehavior.h
```diff
@@ -1,30 +1,25 @@
...
-	int32_t m_targetEnemy;
-
-	int32_t m_targetFriend;
-
-	/*
-	 * Inherited
-	 */
-	explicit AreaOfEffectBehavior(const uint32_t behaviorId) : Behavior(behaviorId) {
-	}
+	bool m_useTargetPosition;
```
EmosewaMC, line 11: all these should be private

dGame/dBehaviors/BehaviorContext.cpp
aronwk-aaron: So with this, I knew that was going to be a comment.
I personally like nesting like that. but yeah, it does make it not as easy to read.
The reason I did it like this was because this is the flow in the client and it helped me make sure I had the checks correct. I can try to re-write it more flat, but that will take a bit to do and verify

dGame/dBehaviors/AreaOfEffectBehavior.cpp
```diff
@@ -20,134 +20,101 @@ void AreaOfEffectBehavior::Handle(BehaviorContext* context, RakNet::BitStream* b
...
 	}
 }
 
 void AreaOfEffectBehavior::Calculate(BehaviorContext* context, RakNet::BitStream* bitStream, BehaviorBranchContext branch) {
-	auto* self = EntityManager::Instance()->GetEntity(context->caster);
-	if (self == nullptr) {
-		Game::logger->Log("AreaOfEffectBehavior", "Invalid self for (%llu)!", context->originator);
+	auto* caster = EntityManager::Instance()->GetEntity(context->caster);
+	if (!caster) {
+		Game::logger->Log("AreaOfEffectBehavior", "Invalid caster for (%llu)!", context->originator);
```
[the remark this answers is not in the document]
aronwk-aaron: Would the %i in this log be the behavior ID or the handle ID?

dGame/dBehaviors/AreaOfEffectBehavior.cpp
aronwk-aaron: I did check this and the FX were not happening before these changes

dGame/dBehaviors/AreaOfEffectBehavior.cpp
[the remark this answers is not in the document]
EmosewaMC: behavior ID since handles are generated but behaviors can be cast

dGame/dBehaviors/BehaviorContext.cpp
EmosewaMC: Another reason to try to reduce nesting is to reduce the number of control paths there are.  There are quite a few spots where you do the same logic so reducing the number of control flow paths would help speed and also make it more clear whats going on

EmosewaMC (approved):

Jettford (approved):

Jettford:
Looks all good, a few notes about programming practices passed along but nothing worth blocking over.

PR #1041, opened 2023-04-03
Fix models not disassembling into bricks

EmosewaMC:
Tested that the LOTs 14709 and 9629 can be trashed into bricks now whereas 14709 could not previously.


aronwk-aaron (approved):

aronwk-aaron:
Just to check, models with null or empty string in that column still function correctly, right?

EmosewaMC:
yes

Wincent01 (approved):
Didn't know this column existed, neat

PR #1043, opened 2023-04-03
Fix and fully implement ChangeOrientation behavior

aronwk-aaron:
Tested that this works as intended.
to_angle in unused in behaviors in live but has handling in the client.
to_point in unused in both, so not implementing it, even in where it is used in the cdclient, it doesn't make sense sine there i no point associated with it to orient to.

Tested that this works as intended now.
See Epsilon or the Sentinel Guards in AG. they will look at what they are shooting at now


EmosewaMC (dismissed):
Client seems to use NiPoint::UNIT_Z but other than that if it looks correct then it is correct

PR #1044, opened 2023-04-05
Resolve all issues with invisible enemies and End Behavior nodes not firing

### LINKED ISSUE #1001: BUG: Enemies sometimes don't fully smash

### DarkflameServer Version

All

### Issue Description

Enemies sometimes don't smash when they should take lethal damage.  This issue is really several other issues that we're lumping into 1 since they are all related to race conditions.  

This issue will never be fully solved, just mitigated as the server becomes more robust.

### Reproduction Steps

Deal what would be lethal damage to an enemy at the same frame you take damage from them on the server
Enemy respawns after 4 seconds

### Expected Behavior

Enemy dies as expected

### Environment

All

EmosewaMC:
Fixes #1001 

Removes the interruptions of synced player skills from interruptions and implements a garbage collection system to delete old skills that have not been synced since their timeout time + 10 seconds to account for super laggy frames.

The motivation for this change is the realization that it is not physically possible to do any server verification for synced skills because of latency, lag frames and delayed sends via RakNet.  Because we could not verify when a message arrived, we cannot implement a timing mechanism to expire synced skills right as they would expire on the client.  For the same reason, we cannot simply calculate ahead of time the result of the skill and just echo it down to the client.  There could be extra latency tossed on top of these casts between the start and sync or a delay so large before the start that start and sync are sent back to back.  For this reason we have to trust clients that their synced times are correct and the bitstreams are sent in good faith.

During a full round of Battle of Nimbus Station I encountered zero issues with enemies going invisible and zero issues with the slow down for the wormholer remaining after I was interrupted out of the skill.  
Tested with a log line in the destructor that sync behavior entries were properly expiring and that the context was being correctly cleaned up due to the syncs expiring.  

Acknowledgements: If someone lags for 10 or more seconds then yes the skills will desync as they did previously, but if someone is at 10000ms delay between packets being processed then I believe there are other issues at play.
Bad actors can abuse this however the number of actors that can do this is 4 people and creating a well tuned playing experience for all players should be targeted first and foremost with cheater protection being taken into account with earlier validity checks and properly functioning interrupts ending the sync when it is finished on the client.

Overall the net improvement is Battle of Nimbus Station is much more enjoyable to play and I was not dying due to invisible enemies at all.


aronwk-aaron (dismissed):
👏👏👏

EmosewaMC:
marking as draft for some new bugs that are much more reproducable

EmosewaMC:
Tests reveal the issue is somewhere else and is caused by an under-read BitStream

dGame/dComponents/SkillComponent.cpp
```diff
@@ -192,7 +196,15 @@ void SkillComponent::Interrupt() {
 	auto* combat = m_Parent->GetComponent<BaseCombatAIComponent>();
 	if (combat != nullptr && combat->GetStunImmune()) return;
 
-	for (const auto& behavior : this->m_managedBehaviors) behavior.second->Interrupt();
+	for (const auto& behavior : this->m_managedBehaviors) {
+		for (const auto& behaviorEndEntry : behavior.second->endEntries) {
+			behaviorEndEntry.behavior->End(behavior.second, behaviorEndEntry.branchContext, behaviorEndEntry.second);
+		}
+		behavior.second->endEntries.clear();
+		if (m_Parent->GetLOT() == 1) continue;
```
Wincent01, line 204: IsPlayer() could be used, believe it's a bit cleaner

dGame/dComponents/SkillComponent.cpp
```diff
@@ -134,6 +134,10 @@ void SkillComponent::Update(const float deltaTime) {
 		CalculateUpdate(deltaTime);
 	}
 
+	if (m_Parent->GetLOT() == 1) {
```
Wincent01, line 137: IsPlayer() could be used here

dGame/dBehaviors/BehaviorContext.cpp
```diff
@@ -183,6 +186,21 @@ void BehaviorContext::SyncCalculation(const uint32_t syncId, const float time, B
 	this->syncEntries.push_back(entry);
 }
 
+void BehaviorContext::UpdatePlayerSyncs(float deltaTime) {
```
Wincent01, line 189: This does account for two entries timing out on the same frame, right? I could see it, just an odd way to iterate.

dGame/dComponents/SkillComponent.cpp
```diff
@@ -192,7 +196,15 @@ void SkillComponent::Interrupt() {
 	auto* combat = m_Parent->GetComponent<BaseCombatAIComponent>();
 	if (combat != nullptr && combat->GetStunImmune()) return;
 
-	for (const auto& behavior : this->m_managedBehaviors) behavior.second->Interrupt();
+	for (const auto& behavior : this->m_managedBehaviors) {
```
Wincent01, line 199: Should the managed behaviors be cleared after this?

Wincent01 (commented):
Some thoughts on the implementation. Good job.

dGame/dBehaviors/BehaviorContext.cpp
EmosewaMC: Yes, which is why it iterates so oddly.  If you use range based you are modifying the vector while in a range based vector which invalidates all iterators past the one you erased, hence decrement time, check if we should remove it and if we should, just remove it otherwise go to the next element

dGame/dComponents/SkillComponent.cpp
EmosewaMC: The managed behaviors are need to be maintained and not cleared as clearing them would cause the desyncs players currently experience.  To match the client, end nodes fire off as to keep start nodes in sync with what the client would expect them to be in, however we dont want to interrupt anything that could be synced from a player because if we cleared those during the lag frames, players would experience the issues they currently see with skills not having a valid uid entry in the managed behaviors.

dGame/dComponents/SkillComponent.cpp
EmosewaMC: Yes, much better than lot 1

aronwk-aaron (approved):

Wincent01 (approved):

PR #1047, opened 2023-04-07
Allow case insensitive commands

EmosewaMC:
Tested that /lookup sword of fire and /Lookup sword of fire both logged a chat message.

### DIFF
--- a/dGame/dUtilities/SlashCommandHandler.cpp
+++ b/dGame/dUtilities/SlashCommandHandler.cpp
@@ -82,44 +82,24 @@
 #include "CDZoneTableTable.h"
 
 void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entity* entity, const SystemAddress& sysAddr) {
+	auto commandCopy = command;
+	// Sanity check that a command was given
+	if (command.empty() || command.front() != u'/') return;
+	commandCopy.erase(commandCopy.begin());
+
+	// Split the command by spaces
 	std::string chatCommand;
 	std::vector<std::string> args;
+	auto wideCommand = GeneralUtils::SplitString(commandCopy, u' ');
+	if (wideCommand.empty()) return;
 
-	uint32_t breakIndex = 0;
-	for (uint32_t i = 1; i < command.size(); ++i) {
-		if (command[i] == L' ') {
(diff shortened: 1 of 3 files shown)


dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -82,44 +82,24 @@
...
-		}
-
-		if (arg != "") {
-			args.push_back(arg);
-		}
-
-		break;
-	}
+	// Convert the command to lowercase
+	chatCommand = GeneralUtils::UTF16ToWTF8(wideCommand.front());
```
Wincent01, line 97: If I recall correctly this method can throw if the input ain't possible to convert. If that's the case this shouldn't be used to handle user input.

Xiphoseer: Technically not throwing is the point of wtf8: unpaired utf16 surrogates are still converted to their reserved codepoints in utf8. Which is invalid utf8 but only matters for strings broken to begin with

Xiphoseer (approved):

dGame/dUtilities/SlashCommandHandler.cpp
Xiphoseer: AsciiToUtf16 was the one that throws because it has not enough info on how to deal with bytes above 127 and was meant for easy conversion of stuff like identifiers from CDClient

EmosewaMC: yeah thats what I recall the tests have however as a safety net I can throw a try catch if that is ideal?

Xiphoseer: If anything it's this `assert` that will abort, which I guess try/catch doesn't solve because it aborts the program. However, to reach this, you'd need to pass a parameter above 0x10FFFF, which you can't get with UTF-16 input because the values are either already `<= 0xFFFF` by being a `uint16_t` or a 20 bit surrogate pair which are `<= FFFFF` and offest by `0x10000` which also remains within `0x10000` to `0x10FFFF`

aronwk-aaron (approved):

PR #1050, opened 2023-04-09
Implements all the pickup scripts

aronwk-aaron:
Tested to make sure they work
Tested to make sure existing script works still
killing it immediately is live accurate, the timer was not accurate


DarwinAnim8or (commented):
Why seperate the various coin spawners into a different class? All that seems different here is just the value of coins, which could just be an argument to the constructor of your base SpecialCoinSpawner class

aronwk-aaron:
> Why seperate the various coin spawners into a different class? All that seems different here is just the value of coins, which could just be an argument to the constructor of your base SpecialCoinSpawner class

@EmosewaMC brought this up, but I had already done it this way lol

aronwk-aaron:
note, armor and speed buff don't work due to something in the skill. effects play but repair and speedboost are not given. out of scope for this PR. but noting for me to look at later

EmosewaMC:
Has thos broken the imagination pickups on the ag monument? Those were working fine before

EmosewaMC:
Ah armor and speedboost specifically nvm

dScripts/ai/SPEC/SpecialPowerupSpawner.cpp
```diff
@@ -0,0 +1,26 @@
...
+}
+
+void SpecialPowerupSpawner::OnProximityUpdate(Entity* self, Entity* entering, const std::string name, const std::string status) {
+	if (name != "powerupEnter" && status != "ENTER") return;
+	if (!entering->IsPlayer()) return;
+	if (self->GetVar<bool>(u"bIsDead")) return;
+
+	GameMessages::SendPlayFXEffect(self, -1, u"pickup", "", LWOOBJID_EMPTY, 1, 1, true);
+
+	SkillComponent* skillComponent;
```
EmosewaMC, line 20: We should use GetComponent instead.  This is an outdated method

dScripts/CppScripts.cpp
```diff
@@ -860,6 +861,36 @@ CppScripts::Script* CppScripts::GetScript(Entity* parent, const std::string& scr
...
+		script = new SpecialCoinSpawner(250000);
+	else if (scriptName == "scripts\\ai\\SPEC\\L_SPECIAL_25_SILVER-COIN-SPAWNER.lua")
+		script = new SpecialCoinSpawner(2500);
+	else if (scriptName == "scripts\\ai\\SPEC\\L_SPECIAL_IMAGINE-POWERUP-SPAWNER.lua")
+		script = new SpecialPowerupSpawner(13);
+	else if (scriptName == "scripts\\ai\\SPEC\\L_SPECIAL_IMAGINE-POWERUP-SPAWNER-2PT.lua")
+		script = new SpecialPowerupSpawner(129);
+	else if (scriptName == "scripts\\ai\\SPEC\\L_SPECIAL_LIFE-POWERUP-SPAWNER.lua")
+		script = new SpecialPowerupSpawner(5);
+	else if (scriptName == "scripts\\ai\\SPEC\\L_SPECIAL_ARMOR-POWERUP-SPAWNER.lua")
```
EmosewaMC, line 889: I dont understand why this doesn't work.  All it does is target caster and repair their armor.  Are you sure that issue is skill related?

aronwk-aaron: Yes, see previous commit where I had a log to verify that the skill was being cast. And the effects for the speed boost pay as well, just don't get the boost

dScripts/ai/SPEC/SpecialPowerupSpawner.cpp
aronwk-aaron: Ah yep, I just used the existing script that was there, will update later

dScripts/CppScripts.cpp
EmosewaMC: Hm I'm still quite skeptical of that claim considering how simple the skill is.  Can you verify that the repair behavior is just not repairing or that it's repairing the wrong entity?

aronwk-aaron: speed needed to be split out and use target as caster since the pickup itself dies and stops the skill
armor was using the wrong skill, they updated the object but not the script at some point

dScripts/ai/SPEC/SpecialPowerupSpawner.cpp
```diff
@@ -17,8 +17,9 @@ void SpecialPowerupSpawner::OnProximityUpdate(Entity* self, Entity* entering, co
 
 	GameMessages::SendPlayFXEffect(self, -1, u"pickup", "", LWOOBJID_EMPTY, 1, 1, true);
 
-	SkillComponent* skillComponent;
-	if (!self->TryGetComponent(eReplicaComponentType::SKILL, skillComponent)) return;
+	auto skillComponent = self->GetComponent<SkillComponent>();
+	if (!skillComponent) return;
+	Game::logger->Log("SpecialPowerupSpawner", "cast skill %i on %llu", this->m_SkillId, entering->GetObjectID());
```
EmosewaMC, line 22: DebugLog here

EmosewaMC (approved):

PR #1053, opened 2023-04-11
Resolve all but one compiler warning

EmosewaMC:
Server compiles still.  Most fixes are extra switch statements that were missing or saving return values

### DIFF
--- a/dCommon/Diagnostics.cpp
+++ b/dCommon/Diagnostics.cpp
@@ -111,7 +111,7 @@ static void ErrorCallback(void* data, const char* msg, int errnum) {
 
 void GenerateDump() {
 	std::string cmd = "sudo gcore " + std::to_string(getpid());
-	system(cmd.c_str());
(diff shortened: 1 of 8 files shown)


dCommon/Diagnostics.cpp
```diff
@@ -111,7 +111,7 @@ static void ErrorCallback(void* data, const char* msg, int errnum) {
 
 void GenerateDump() {
 	std::string cmd = "sudo gcore " + std::to_string(getpid());
-	system(cmd.c_str());
+	int ret = system(cmd.c_str()); // Saving a return just to prevent warning
```
Wincent01, line 114: I think you could do `(void) system(...)`

Wincent01 (commented):
I think you could use void casting to get rid of some of these warnings

Wincent01 (approved):

aronwk-aaron (approved):

PR #1056, opened 2023-04-12
Add random end of race behavior

aronwk-aaron:
for the different animation types

### DIFF
--- a/dGame/dComponents/VehiclePhysicsComponent.cpp
+++ b/dGame/dComponents/VehiclePhysicsComponent.cpp
@@ -93,7 +93,7 @@ void VehiclePhysicsComponent::Serialize(RakNet::BitStream* outBitStream, bool bI
 	}
 
 	if (bIsInitialUpdate) {
-		outBitStream->Write<uint8_t>(5);
+		outBitStream->Write<uint8_t>(GeneralUtils::GenerateRandomNumber<uint32_t>(0, 7));
 		outBitStream->Write1();
 	}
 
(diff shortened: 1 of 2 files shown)


EmosewaMC (approved):
just make sure this does in fact fix what we believe it does

dGame/dComponents/VehiclePhysicsComponent.cpp
```diff
@@ -93,7 +93,7 @@ void VehiclePhysicsComponent::Serialize(RakNet::BitStream* outBitStream, bool bI
 	}
 
 	if (bIsInitialUpdate) {
-		outBitStream->Write<uint8_t>(5);
+		outBitStream->Write<uint8_t>(GeneralUtils::GenerateRandomNumber<uint32_t>(0, 7));
```
Wincent01, line 96: It seems a bit weird to have this kind of logic in the serialization. Would prefer if this was in the constructor, or better in the logic of the racing manager. This would also make sure that clients see the same player doing different animations because they loaded in at separate times.

Wincent01 (changes requested):

dGame/dComponents/VehiclePhysicsComponent.cpp
EmosewaMC: Ah yeah good catch.  Didn't think about that bit

EmosewaMC: I think this would be better in the vehiclePhysicsComponent since the behavior is only attached to this component and adding more to the racing manager wouldn't be ideal since it is quite large as it is.  Just my 2 cents though

aronwk-aaron:
tested that the behavior set on initialization is the one that gets played at the end of the race for that player

EmosewaMC (approved):

Wincent01 (approved):

PR #1062, opened 2023-05-01
Add tests for LDF parsing and serialization.  Cleanup LDF header

EmosewaMC:
- Implement a speedup by using overloaded operators to put non-strings directly to a stringstream as opposed to doing to_string first.
- Replace header guard with a traditional C style one
- Fix ReadFromString so it can handle a string of colons (rare case, but an edge case regardless)\
- Add further edge case testing for LDF keys to ensure bad keys are not parsed and to ensure correct behavior

All new tests are attached and were created _before_ making any changes and will be run through regression when the PR is finished


EmosewaMC:
All tests pass after the fixes and most tests pass pre-fixes (except those which threw originally or had legitimate bugs.  Will now do a final sweep for optimization to prevent copying of so many strings

dMasterServer/MasterServer.cpp
```diff
@@ -334,6 +334,15 @@ int main(int argc, char** argv) {
...
 		Game::im->GetInstance(1000, false, 0);
+		Game::im->GetInstance(1100, false, 0);
+		Game::im->GetInstance(1200, false, 0);
+		Game::im->GetInstance(1300, false, 0);
+		Game::im->GetInstance(1400, false, 0);
+		Game::im->GetInstance(1600, false, 0);
+		Game::im->GetInstance(1700, false, 0);
+		Game::im->GetInstance(1800, false, 0);
+		Game::im->GetInstance(1900, false, 0);
+		Game::im->GetInstance(2000, false, 0);
```
Wincent01, line 345: This shouldn't be here.

Wincent01 (changes requested):

dMasterServer/MasterServer.cpp
EmosewaMC: Lol I forgot about this

aronwk-aaron (approved):

DarwinAnim8or (approved):

PR #1075, opened 2023-05-07
Removed some hardcoded logic for racing

aronwk-aaron:
Including return world and what activity ID to use for rewards

Tested that single player racing still works and that proper racing rewards are given


dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -279,15 +279,15 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 	if (chatCommand == "leave-zone") {
 		const auto currentZone = dZoneManager::Instance()->GetZone()->GetZoneID().GetMapID();
 
-		auto newZone = 0;
```
DarwinAnim8or, line 282: why is Auto not preferred here?

DarwinAnim8or (approved):
I like the moving of accessible zone checking to dZoneManager / checking map files locally :)

EmosewaMC (approved):
Only nit is this could just be static since it doesn't need the ZoneManager class to work.

PR #1079, opened 2023-05-09
refactor: removed hardcoded ag laser logic

### LINKED ISSUE #598: BUG: Monument lasers don't function 100% properly

### DarkflameServer Version

yadda yadda

### Issue Description

already discussed extensively on discord, this is largely just a reminder

- the animation/particle effect of the player being burned plays but the knockback isn't applied
- the particle effect/anim only plays the first time you walk into a laser, all times following that you just silently take 1 damage
- special laser death anim does not play upon dying, under any circumstances i can find

### Reproduction Steps

walk into monument laser trigger w/ sensor script

### Expected Behavior

https://youtu.be/dUE7YilB_wc?t=3
https://youtu.be/dUE7YilB_wc?t=179
https://youtu.be/dUE7YilB_wc?t=207

it should be casting this skill on the player
https://explorer.lu-dev.net/skills/163
https://explorer.lu-dev.net/skills/behaviors/15714/15713
(does 1 damage, applies knockback, and includes special laser death anim)
at least that's what the old pre-freetrial version of AG did on its triggers, and available footage of live post-freetrial seems to have the exact same effect, just accomplished via lua instead of lutriggers

lcdr — Yesterday at 4:22 PM
fwiw, these are the script implementations on lcdru:
https://github.com/lcdr/luserver/blob/master/luserver/scripts/avant_gardens/laser.py
https://github.com/lcdr/luserver/blob/master/luserver/scripts/avant_gardens/laser_sensor.py
iirc i implemented them pretty much the same way they worked in live

### Environment

punchcard machine

aronwk-aaron:
Fixes #598 
Tested that lasers work when beam is red hot
Tested that no burning happens when there is no glowy doom ray
Tested that the hurt machine starts working again when the laser respawns

### DIFF
--- a/dScripts/02_server/Map/AG/AgLaserSensorServer.cpp
+++ b/dScripts/02_server/Map/AG/AgLaserSensorServer.cpp
@@ -21,7 +21,7 @@ void AgLaserSensorServer::OnCollisionPhantom(Entity* self, Entity* target) {
 	auto active = self->GetVar<bool>(u"active");
 	if (!active) return;
 	auto skillCastID = self->GetVarAs<float>(u"skillCastID");
-	if (!skillCastID) skillCastID = m_SkillCastID;
+	if (skillCastID == 0) skillCastID = m_SkillCastID;
 	auto* skillComponent = self->GetComponent<SkillComponent>();
 	if (!skillComponent) return;
 	skillComponent->CastSkill(m_SkillCastID, target->GetObjectID());
--- a/dScripts/02_server/Map/AG/AgLaserSensorServer.cpp
+++ b/dScripts/02_server/Map/AG/AgLaserSensorServer.cpp
@@ -2,56 +2,27 @@
 
 #include "PhantomPhysicsComponent.h"
 #include "SkillComponent.h"
-#include "EntityManager.h"
-#include "AgMonumentLaserServer.h"
-#include "EntityManager.h"
 #include "ePhysicsEffectType.h"
-#include "eReplicaComponentType.h"
 
 void AgLaserSensorServer::OnStartup(Entity* self) {
-
-	PhantomPhysicsComponent* physComp = static_cast<PhantomPhysicsComponent*>(self->GetComponent(eReplicaComponentType::PHANTOM_PHYSICS));
-	physComp->SetPhysicsEffectActive(true);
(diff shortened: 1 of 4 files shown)


dScripts/02_server/Map/AG/AgLaserSensorServer.cpp
```diff
@@ -2,56 +2,27 @@
...
-		}
-	}
-
-	if (laser != nullptr) {
-		m_Skill->CalculateBehavior(m_SkillCastID, 15714, target->GetObjectID());
-	}
+	auto active = self->GetVar<bool>(u"active");
+	if (!active) return;
+	auto skillCastID = self->GetVarAs<float>(u"skillCastID");
+	if (!skillCastID) skillCastID = m_SkillCastID;
```
EmosewaMC, line 24: This is an integer right?  Not sure what this check is checking for unless you mean if != 0

Also why are we storing this member variable on the script?  imo the variable should either be attached only to the entity, or only to the script.  not both.  If its attached to the script it should be hard coded from the get go and if its attached to the entity, it should only be fetched through GetVar.  If going for the latter, the member variable should be removed but if going for the former, this should be done once on entity initialization.
This comment also applies for the repel force

dScripts/02_server/Map/AG/AgMonumentLaserServer.h
```diff
@@ -4,7 +4,5 @@
 class AgMonumentLaserServer : public CppScripts::Script {
 public:
 	void OnStartup(Entity* self);
-	void OnProximityUpdate(Entity* self, Entity* entering, std::string name, std::string status);
-private:
-	float m_Radius = 25.0f;
+	void OnDie(Entity* self, Entity* killer) override;
```
EmosewaMC, line 7: `override`

dScripts/02_server/Map/AG/AgLaserSensorServer.cpp
EmosewaMC: I read the setter backwards - but we should still compare with 0 and not check for t/f with an int.

DarwinAnim8or (approved):

EmosewaMC (approved):

PR #1081, opened 2023-05-10
fix: add check for arg nums on handlepushobject to prevent crashes

aronwk-aaron:


EmosewaMC (approved):

dGame/dComponents/TriggerComponent.cpp
```diff
@@ -235,6 +235,8 @@ void TriggerComponent::HandleRotateObject(Entity* targetEntity, std::vector<std:
 }
 
 void TriggerComponent::HandlePushObject(Entity* targetEntity, std::vector<std::string> argArray){
+	if (argArray.size() < 3) return;
```
Wincent01, line 238: When does this evaluate to false? Is there something missing in the implementation to handle that case?

aronwk-aaron: There's a case in old AG maps where a command should be Castskill and not push opbject. I thought about adding an edge case to handle that one instance, but didn't

Wincent01 (approved):

PR #1083, opened 2023-05-10
fix: prevent negative imagination

### LINKED ISSUE #1027: BUG: Negative Imagination

### DarkflameServer Version

Don't know, I'm afraid - BLU's current version.

### Issue Description

If a boulder from an ape hits the player between them activating the imagination regeneration ability on the pharaoh's hat, and them receiving the first 10 imagination, the client and server will disagree on how much imagination the player has - the server will reset the player's imagination to 0, whilst the client will increase it. Anything which only uses clientside imagination checks will then cause the player to end up with negative imagination on the server (i.e., using an ability). This can persist across sessions, and can be tested by using the pharaoh's hat again, and noting that it will give less than 10 imagination on first use.
This can also be used to complete quickbuilds, such as those on Crux, with insufficient imagination.

### Reproduction Steps

Use all of your imagination.
Equip pharaoh's hat, and an item which will let you use imagination.
Walk up to ape.
Great it cordially, but from a distance, to incite it to throw a boulder.
Activate the pharaoh's hat just before the boulder is thrown, so that it hits the player before imagination has been received.

Then either:
1)
Wait until you have enough imagination to activate your ability.
Activate the ability.
2)
Build a quickbuild.

Then attempt to regen imagination by any means again, soon after, and you'll see you receive less than you should.

### Expected Behavior

The ability from the pharaoh's hat should probably be cancelled, the server could probably check imagination requirements on casting skills.

However, the only really broken thing about this is infinite imagination for quickbuilds, so the server should probably just check the player's imagination is greater than zero before allowing them to begin a quickbuild.

### Environment

BLU Skill Issue.

aronwk-aaron:
And fail switch if we don't have enough imagination
Fixes #1027

### DIFF
--- a/dGame/dBehaviors/SwitchBehavior.cpp
+++ b/dGame/dBehaviors/SwitchBehavior.cpp
@@ -30,7 +30,7 @@ void SwitchBehavior::Handle(BehaviorContext* context, RakNet::BitStream* bitStre
 
 	Game::logger->LogDebug("SwitchBehavior", "[%i] State: (%d), imagination: (%i) / (%f)", entity->GetLOT(), state, destroyableComponent->GetImagination(), destroyableComponent->GetMaxImagination());
 
-	if (state && destroyableComponent->GetImagination() >= m_imagination) {
+	if (state) {
 		this->m_actionTrue->Handle(context, bitStream, branch);
 	} else {
 		this->m_actionFalse->Handle(context, bitStream, branch);
--- a/dGame/dBehaviors/SwitchBehavior.cpp
+++ b/dGame/dBehaviors/SwitchBehavior.cpp
@@ -30,7 +30,7 @@ void SwitchBehavior::Handle(BehaviorContext* context, RakNet::BitStream* bitStre
 
 	Game::logger->LogDebug("SwitchBehavior", "[%i] State: (%d), imagination: (%i) / (%f)", entity->GetLOT(), state, destroyableComponent->GetImagination(), destroyableComponent->GetMaxImagination());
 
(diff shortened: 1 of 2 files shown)


aronwk-aaron:
Tested that racing imagination drain works as expected still and that the reproduction does not work anymore

dGame/dBehaviors/SwitchBehavior.cpp
```diff
@@ -30,7 +30,7 @@ void SwitchBehavior::Handle(BehaviorContext* context, RakNet::BitStream* bitStre
 
 	Game::logger->LogDebug("SwitchBehavior", "[%i] State: (%d), imagination: (%i) / (%f)", entity->GetLOT(), state, destroyableComponent->GetImagination(), destroyableComponent->GetMaxImagination());
 
-	if (state || (entity->GetLOT() == 8092 && destroyableComponent->GetImagination() >= m_imagination)) {
```
Wincent01, line 33: What was this edge case for?

EmosewaMC: Likely server side verification of the skill, but as we learned with some earlier PRs we shouldn't be trying to do server side calculation of skill handles as it may cause desyncs.  At least thats my thought process when it comes to skill Handles now with what we know

aronwk-aaron: I did test that racing still works as expected

Wincent01 (approved):

EmosewaMC (approved):

PR #1084, opened 2023-05-10
feature: Implement FallSpeedBehavior

aronwk-aaron:
This is actually how it should work, based on RE in the client


aronwk-aaron:
Tested that it works

dGame/dBehaviors/FallSpeedBehavior.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "ControllablePhysicsComponent.h"
+#include "BehaviorContext.h"
+#include "BehaviorBranchContext.h"
+
+
+void FallSpeedBehavior::Handle(BehaviorContext* context, RakNet::BitStream* bitStream, BehaviorBranchContext branch) {
+	// make sure required parameter has non-default value
+	if (m_PercentSlowed == 0.0f) return;
+
+	branch.target = context->caster;
```
Wincent01, line 12: This seems odd, is this necessary?

Wincent01 (commented):
What items use this behavior?

aronwk-aaron:
> What items use this behavior?

There's some parachutes left over from a game challenge

dGame/dComponents/ControllablePhysicsComponent.h
```diff
@@ -276,7 +276,7 @@ class ControllablePhysicsComponent : public Component {
 	 * The speed boosts of this component.
 	 * @return All active Speed boosts for this component.
 	 */
-	std::vector<float> GetActiveSpeedboosts() { return m_ActivePickupRadiusScales; };
+	std::vector<float> GetActiveSpeedboosts() { return m_ActiveSpeedBoosts; };
```
EmosewaMC, line 279: lol

dGame/dComponents/ControllablePhysicsComponent.cpp
```diff
@@ -388,3 +388,29 @@ void ControllablePhysicsComponent::SetStunImmunity(
...
+	if (pos != m_ActiveFallSpeeds.end()) {
+		m_ActiveFallSpeeds.erase(pos);
+	} else {
+		Game::logger->LogDebug("ControllablePhysicsComponent", "Warning: Could not find FallSpeed %f in list of active FallSpeeds.  List has %i active FallSpeeds.", value, m_ActiveFallSpeeds.size());
+		return;
+	}
+
+	// Recalculate FallSpeed since we removed one
+	m_FallSpeed = 0.0f;
+	if (m_ActiveFallSpeeds.size() == 0) { // no active fall speeds left, so return to base gravity
```
EmosewaMC, line 409: `m_ActiveFallSpeeds.empty()`

dGame/dComponents/ControllablePhysicsComponent.cpp
```diff
@@ -388,3 +388,29 @@ void ControllablePhysicsComponent::SetStunImmunity(
...
+		m_FallSpeed = 1;
+	} else { // Used the last applied FallSpeed
+		m_FallSpeed = m_ActiveFallSpeeds.back();
+	}
+	SetSpeedMultiplier(m_FallSpeed); // 500 being the base speed
```
EmosewaMC, line 414: `SetGravityScale(value); // 1 being the base fall speed`

dGame/dComponents/ControllablePhysicsComponent.cpp
```diff
@@ -388,3 +388,29 @@ void ControllablePhysicsComponent::SetStunImmunity(
 		bImmuneToStunUseItem
 	);
 }
+
+void ControllablePhysicsComponent::AddFallSpeed(float value) {
+	m_ActiveFallSpeeds.push_back(value);
+	m_FallSpeed = value;
+	SetGravityScale(value);
```
EmosewaMC, line 395: If possible please implement proper dirtying for Gravity scale.  Right now it is not properly dirtied

EmosewaMC (dismissed):
This implementation is confusing at the moment.
There are references to m_FallSpeed variable being set, but then also references to GravityScale being set instead.  m_FallSpeed is only implemented in the ControllablePhysicsComponent as a member but is never used outside these two functions.  If it is the same as m_GravityScale I think it'd be wise to absolve them together.

aronwk-aaron:
> This implementation is confusing at the moment. There are references to m_FallSpeed variable being set, but then also references to GravityScale being set instead. m_FallSpeed is only implemented in the ControllablePhysicsComponent as a member but is never used outside these two functions. If it is the same as m_GravityScale I think it'd be wise to absolve them together.

So this is done since gravity scale can be set externally, the fall speed is done via behaviors only, same as speed it done.

After looking at the client again, this is the wrong implementation.
it's just basic and sets to the value on cast, and then set to 1 on uncast

DarwinAnim8or (approved):
Code itself looks good, didn't test for live accuracy

Wincent01 (approved):

PR #1085, opened 2023-05-11
fix: not exiting shooting gallery when clicking activity close button

### LINKED ISSUE #436: BUG: Gnarled Forest Cannon Cove Exploit

### Issue Description

Clicking the "X" exit instance button in the top right corner and selecting the X/No option causes the instance to soft reset itself, resulting in additional targets from the previous instance spawning. This allows a player to easily obtain significantly higher scores than normally possible.

Under some circumstances, this can crash the instance.

### Reproduction steps

Enter the instance.

Click the Exit button in the top right corner, but click the no option to remain in the instance.

Wait several seconds.

Click the Start button to begin the instance.

Additionally you can perform this exploit while the instance is in progress. The issue appears to be caused specifically by clicking the no option for exiting the instance.

### Expected Behavior

Not this.

### Environment

Save the environment.
Plant some trees.
Touch some grass.

aronwk-aaron:
Fixes #436
Fixes crash when replaying as well
Fixed replaying

### DIFF
--- a/dGame/dGameMessages/GameMessages.cpp
+++ b/dGame/dGameMessages/GameMessages.cpp
@@ -91,6 +91,7 @@
 #include "eReplicaComponentType.h"
 #include "eClientMessageType.h"
 #include "eGameMessageType.h"
+#include "ActivityManager.h"
 
 #include "CDComponentsRegistryTable.h"
 #include "CDObjectsTable.h"
@@ -6203,4 +6204,19 @@ void GameMessages::SendShowBillboardInteractIcon(const SystemAddress& sysAddr, L
 
 	if (sysAddr == UNASSIGNED_SYSTEM_ADDRESS) SEND_PACKET_BROADCAST
 	else SEND_PACKET
+}
+
+void GameMessages::HandleRequestActivityExit(RakNet::BitStream* inStream, Entity* entity) {
+	bool canceled = false;
+	inStream->Read(canceled);
+	if (!canceled) return;
+
+	LWOOBJID player_id = LWOOBJID_EMPTY;
+	inStream->Read(player_id);
+	auto player = EntityManager::Instance()->GetEntity(player_id);
+	if (!entity || !player) return;
+
+	for (auto* shootingGallery : EntityManager::Instance()->GetEntitiesByComponent(eReplicaComponentType::SHOOTING_GALLERY)) {
+		shootingGallery->RequestActivityExit(entity, player_id, true);
+	}
 }
--- a/dGame/dGameMessages/GameMessages.cpp
+++ b/dGame/dGameMessages/GameMessages.cpp
@@ -6215,8 +6215,5 @@ void GameMessages::HandleRequestActivityExit(RakNet::BitStream* inStream, Entity
 	inStream->Read(player_id);
 	auto player = EntityManager::Instance()->GetEntity(player_id);
 	if (!entity || !player) return;
-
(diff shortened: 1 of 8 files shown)


aronwk-aaron:
Tested that the game can't be started multiple times
Tested that exiting now leaves the instance
Tested that replaying works

Wincent01 (dismissed):

dGame/dGameMessages/GameMessages.cpp
```diff
@@ -6204,3 +6205,18 @@ void GameMessages::SendShowBillboardInteractIcon(const SystemAddress& sysAddr, L
...
+	bool canceled = false;
+	inStream->Read(canceled);
+	if (!canceled) return;
+
+	LWOOBJID player_id = LWOOBJID_EMPTY;
+	inStream->Read(player_id);
+	auto player = EntityManager::Instance()->GetEntity(player_id);
+	if (!entity || !player) return;
+
+	for (auto* shootingGallery : EntityManager::Instance()->GetEntitiesByComponent(eReplicaComponentType::SHOOTING_GALLERY)) {
```
EmosewaMC, line 6219: This GM is not sent by _any_ other activity?  Could you confirm this?

dScripts/ai/MINIGAME/SG_GF/SERVER/SGCannon.h
```diff
@@ -63,12 +63,11 @@ class SGCannon : public ActivityManager {
...
 	void OnFireEventServerSide(Entity* self, Entity* sender, std::string args, int32_t param1, int32_t param2, int32_t param3) override;
-	void OnActivityStateChangeRequest(Entity* self, LWOOBJID senderID, int32_t value1,
-		int32_t value2, const std::u16string& stringValue) override;
-	void OnMessageBoxResponse(Entity* self, Entity* sender, int32_t button, const std::u16string& identifier,
-		const std::u16string& userData) override;
+	void OnActivityStateChangeRequest(Entity* self, LWOOBJID senderID, int32_t value1, int32_t value2, const std::u16string& stringValue) override;
+	void OnMessageBoxResponse(Entity* self, Entity* sender, int32_t button, const std::u16string& identifier, const std::u16string& userData) override;
 	void OnActivityTimerDone(Entity* self, const std::string& name) override;
 	void OnActivityTimerUpdate(Entity* self, const std::string& name, float_t timeRemaining, float_t elapsedTime) override;
+	void OnRequestActivityExit(Entity* self, LWOOBJID player, bool canceled);
```
EmosewaMC, line 70: `override`

dGame/dGameMessages/GameMessages.cpp
aronwk-aaron: The only other place I saw it being sent was in roller component, which we don't have implemented, but I'll double check

EmosewaMC: I'd double check integration testing just in case imo.  Not all GMs are sent from the client exe, some are sent straight from lua, straight from flash etc and the GM call would not be visible in the client (assuming you just looked for references to the function in the client and not in flash/lua)

aronwk-aaron: Lua scripts only used it in sg and old racing

EmosewaMC (approved):

DarwinAnim8or (approved):

Issue #691, opened 2022-07-31
Tracking Issue: Hardcoded Content


Xiphoseer:
This issues is meant to track source code (except `dScripts`) that includes hardcoded content data. Such code should generally not exist, but there may be reasons it does.

- [ ] [dGame/dInventory/ItemSetPassiveAbility.cpp](https://github.com/DarkflameUniverse/DarkflameServer/blob/main/dGame/dInventory/ItemSetPassiveAbility.cpp)

EmosewaMC:
GameMessages HandleModularBuildComplete lot 8129

PR #1087, opened 2023-05-12
fix: orient player correctly when using pirate mast in FV

### LINKED ISSUE #1066: BUG: Misaligned player animation for pirate-only quickbuild in Forbidden Valley

### DarkflameServer Version

2e284eb

### Issue Description

Old bug. When building the pirate-only quickbuild in Forbidden Valley at Mantis Rock, the animation of the player swinging is greatly misaligned but the rope's movement is accurate. When the player finishes the animation in the air, their position is corrected visually.

### Reproduction Steps

Build the pirate-only quickbuild from the crate smashable in Mantis Rock in Forbidden Valley.

### Expected Behavior

The player's animation should be aligned with the quickbuild's animation.

### Environment

ubuntu server

aronwk-aaron:
Fixes #1066 
Tried to make a simpler fix overall to have have these magic numbers, but could not find an overall solution to eliminate the root issue and need for the hacky fixes.

While this "works" we should probably try to investigate the real issue and fix it.

### DIFF
--- a/dScripts/02_server/Map/GF/MastTeleport.cpp
+++ b/dScripts/02_server/Map/GF/MastTeleport.cpp
@@ -43,10 +43,12 @@ void MastTeleport::OnTimerDone(Entity* self, std::string timerName) {
 		GameMessages::SendTeleport(playerId, position, rotation, player->GetSystemAddress(), true);
 
 		// Hacky fix for odd rotations
-		if (self->GetVar<std::u16string>(u"MastName") != u"Jail") {
+		if (self->GetVar<std::u16string>(u"MastName") == u"Elephant") {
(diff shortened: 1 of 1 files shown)


EmosewaMC (commented):
Only get the var once.  Not a fan of this change.  Would rather figure out why its not working.

Wincent01 (dismissed):
The game can be very stubborn sometimes

aronwk-aaron:
> If you fix the triple getting of the same var I can approve this, but an issue should be created in its place to remind us to find the actual cause. My main concern is fixing and forgetting

added it to #691

EmosewaMC (approved):

Wincent01 (approved):

PR #1096, opened 2023-05-17
Correction: Remove intermediate read from bitstream in AMFDeserialize

EmosewaMC:
Per ISO C++ standard 9.7.1 clause 8,
"For an enumeration whose underlying type is fixed, the values of the enumeration are the values of the
underlying type" it is not undefined behavior to set a scoped enum to a value outside of its constant range because all values of the underlying type can represent the scoped enum


Wincent01 (approved):
Per the old laws, this pullio requsto is hiderto approved

aronwk-aaron (approved):

PR #1098, opened 2023-05-18
feat: implement consume item behavior

aronwk-aaron:
Implement consume Item behavior


EmosewaMC:
How has this been tested?

aronwk-aaron:
tested with skill 456, works as expected now.
Needed to add the silent due to the client handling visually removing the item and all we need to do in the server is mirror that, but not echo the remove to the client

DarwinAnim8or (approved):

EmosewaMC (approved):

PR #1099, opened 2023-05-22
Logging Improvements

EmosewaMC:
main points to hit
- Macro expansion allows for easy logging and also provides the precise location of the log as opposed to just a general class location.  This removes any issues with repeated or duplicated logs in the same class as each filename is generally unique.  
- Logger itself is managed by a unique_ptr so it is automatically deleted on application close
- Simplified some terminology (SetLogDebugStatements -> SetLogDebug)
- Added color logging for high priority alerts to the console
- Log not the class name, but rather the file name and line number.  This string is calculated at compile time.

TODO:
- [x] Implement color console logging for Windows
- [ ] - Play with Parameter Packing to remove need to C style arrays and string management


EmosewaMC:
Going to bin color logging for now.  Going to be a slog to implement.

EmosewaMC:
This is bad.

PR #1105, opened 2023-05-30
Fix being able to buy an item from a vendor even if they dont sell it

EmosewaMC:
Tested that I can still buy items from the commendation vendor in Nexus Tower and from a sentinel vendor in Nimbus Station.

### DIFF
--- a/dGame/dGameMessages/GameMessages.cpp
+++ b/dGame/dGameMessages/GameMessages.cpp
@@ -4733,11 +4733,16 @@ void GameMessages::HandleBuyFromVendor(RakNet::BitStream* inStream, Entity* enti
 
 	const auto isCommendationVendor = entity->GetLOT() == 13806;
 
-	VendorComponent* vend = static_cast<VendorComponent*>(entity->GetComponent(eReplicaComponentType::VENDOR));
+	auto* vend = entity->GetComponent<VendorComponent>();
 	if (!vend && !isCommendationVendor) return;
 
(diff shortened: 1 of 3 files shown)


dGame/dGameMessages/GameMessages.cpp
```diff
@@ -4733,12 +4733,17 @@ void GameMessages::HandleBuyFromVendor(RakNet::BitStream* inStream, Entity* enti
...
 
-	VendorComponent* vend = static_cast<VendorComponent*>(entity->GetComponent(eReplicaComponentType::VENDOR));
+	auto* vend = entity->GetComponent<VendorComponent>();
 	if (!vend && !isCommendationVendor) return;
 
-	InventoryComponent* inv = static_cast<InventoryComponent*>(player->GetComponent(eReplicaComponentType::INVENTORY));
+	auto* inv = player->GetComponent<InventoryComponent>();
 	if (!inv) return;
 
+	if (!isCommendationVendor && !vend->SellsItem(item)) {
```
aronwk-aaron, line 4742: Non-blocking
Should there be a check for the commendation vendor as well? shouldn't be toooooo hard :P
But should probably be it's own PR since it will be a bit more involved

aronwk-aaron (approved):

dGame/dGameMessages/GameMessages.cpp
EmosewaMC: He has his own check below.  It checks that the item is actually sold for tokens hence dont check here

Wincent01 (approved):

PR #1140, opened 2023-07-15
Move EntityManager to Game namespace

EmosewaMC:
Makes the code more testable.

There are the same number of `Game::entityManager` as there are `EntityManager::Instance()`

Server still runs.

### DIFF
--- a/dCommon/Game.h
+++ b/dCommon/Game.h
@@ -10,6 +10,7 @@ class dConfig;
 class RakPeerInterface;
 class AssetManager;
 struct SystemAddress;
+class EntityManager;
 
 namespace Game {
 	extern dLogger* logger;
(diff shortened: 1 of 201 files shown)


aronwk-aaron (dismissed):

EmosewaMC:
Need to move ZoneControl Entity initialization to later.  recursive dependency detected.

Wincent01 (approved):
Good way to handle singletons. If a hypothetical future where multiple zones run on the same instance, these variables could easily be marked as thread_local.

EmosewaMC:
> Good way to handle singletons. If a hypothetical future where multiple zones run on the same instance, these variables could easily be marked as thread_local.

This also makes the code so much more maintainable and testable!  Many benefits here and an overall win :)

aronwk-aaron (approved):

PR #1147, opened 2023-07-19
fix: Nexus Tower Combat Challenge can be started twice if two players start it at the same time

### LINKED ISSUE #1135: BUG: Nexus Tower Combat Challenge can be started twice if two players start it at the same time

### DarkflameServer Version

All

### Issue Description

If two players start the combat challenge at the same time, the game will be started twice with two dummies spawning.  This causes the initial starter to not be saved and makes it so the extra spawned dummy does nothing except spawn the next dummy in the list, terminating this spawning when the player who started the bugged instance leaves the zone and the dummy dies.

### Reproduction Steps

1) have two players with the window to start the combat challenge open at the same time
2) have both players start at the same time
3) Observe that there are two sets of dummies.

### Expected Behavior

Only initial player who started the game has its dummies spawn

### Environment

All

TAHuntling:
DESC: Fix for dual instance starting when two players hit start on the challenge at the same time.

MOTIVATION: Want to fix issue for any future occurances

TESTING: Will test tomorrow/later

Fixes #1135

### DIFF
--- a/dScripts/02_server/Map/NT/NtCombatChallengeServer.cpp
+++ b/dScripts/02_server/Map/NT/NtCombatChallengeServer.cpp
@@ -43,7 +43,7 @@ void NtCombatChallengeServer::OnFireEventServerSide(Entity* self, Entity* sender
 
 
 void NtCombatChallengeServer::OnMessageBoxResponse(Entity* self, Entity* sender, int32_t button, const std::u16string& identifier, const std::u16string& userData) {
-	if (identifier == u"PlayButton" && button == 1 && !self->GetVar<bool>(u"bInUse")) {
+	if (identifier == u"PlayButton" && button == 1 && !self->GetNetworkVar<bool>(u"bInUse")) {
 		self->SetNetworkVar(u"bInUse", true);
 
 		self->SetVar(u"playerID", sender->GetObjectID());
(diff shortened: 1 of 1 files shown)


dScripts/02_server/Map/NT/NtCombatChallengeServer.cpp
```diff
@@ -43,7 +43,7 @@ void NtCombatChallengeServer::OnFireEventServerSide(Entity* self, Entity* sender
 
 
 void NtCombatChallengeServer::OnMessageBoxResponse(Entity* self, Entity* sender, int32_t button, const std::u16string& identifier, const std::u16string& userData) {
-	if (identifier == u"PlayButton" && button == 1) {
+	if (identifier == u"PlayButton" && button == 1 && !self->GetVar<bool>(u"bInUse")) {
```
Wincent01, line 46: The variable in this case is a network variable, which is separate from the normal once. This will have to use the GetNetworkVar variant on the method.

EmosewaMC: Oh interesting, I must have mis-read the API.  Should be a simple fix :)

aronwk-aaron (approved):

EmosewaMC (approved):

PR #1149, opened 2023-07-20
fix: Players are able to join a race without having a race car

### LINKED ISSUE #780: BUG: Players are able to join a race without having a race car

### DarkflameServer Version

All

### Issue Description

Not a high priority issue, but one to be noted.  If you are doing this you know what you are doing and are doing it with intention.

### Reproduction Steps

1) have a race car and go to dismantle it.
2) dont close the dismantle pop-up
3) go to start a race (solo or not doesnt matter)
4) ready up for the race and dismantle the car
5) you will now be in a race without a car and are stuck until you re-log or quit the game.

### Expected Behavior

race does not launch or the player without the car gets removed.

### Environment

All

TAHuntling:
DESC: Changed RacingControlComponent to boot players back to the hub world when trying to race after dismantling a vehicle.

MOTIVATION: Avoid players being stuck indefinitely when trying to load into a race after dismantling a vehicle and starting the race

CHANGES: Code added to the block for identifying the player's vehicle. If the vehicle pointer is null, then rather than simply returning, it will now put the player back into the hub world where the race was accessed, and update the race instance to remove the player from the loading queue so the race can continue for other players who do have their vehicles properly equipped.

TESTING: Tested solo, with two players, and with three players. Test results were successful.

Fixes #780


dGame/dComponents/RacingControlComponent.cpp
```diff
@@ -99,7 +98,9 @@ void RacingControlComponent::LoadPlayerVehicle(Entity* player,
 
 	if (item == nullptr) {
 		Game::logger->Log("RacingControlComponent", "Failed to find item");
-
+		auto* playerInstance = dynamic_cast<Player*>(player);
```
EmosewaMC, line 101: I know its not done above, but when doing a `dynamic_cast` from one pointer to another, you generally want to check that the resulting pointer is not null.  Should the dynamic_cast fail, this would be a nullptr and crash.  It wouldn't happen in practice, but always good to check your bounds.
c.f. https://en.cppreference.com/w/cpp/language/dynamic_cast section 5.c

dGame/dComponents/RacingControlComponent.cpp
```diff
@@ -99,7 +98,9 @@ void RacingControlComponent::LoadPlayerVehicle(Entity* player,
...
+		m_LoadedPlayers--;
```
EmosewaMC, line 102: You'll also want to remove the player from the m_LobbyPlayers vector, that or just do the SendToZone and let the update method take care of churning the correct variables.

dGame/dComponents/RacingControlComponent.cpp
```diff
@@ -99,8 +103,13 @@ void RacingControlComponent::LoadPlayerVehicle(Entity* player,
 
 	if (item == nullptr) {
 		Game::logger->Log("RacingControlComponent", "Failed to find item");
-
+		auto* playerInstance = dynamic_cast<Player*>(player);
```
EmosewaMC, line 106: dont forget to remove the player from the list of lobby players here

EmosewaMC (approved):

aronwk-aaron (approved):

PR #1151, opened 2023-07-22
Pet Dig Distance Slightly Raised to Help Combat Skeleton Dragon Dig issue on Crux

TAHuntling:
Description: Slightly increased the distance for pet digs which should help combat the issue with skeleton dragons being unable to dig up bones on Crux.

MOTIVATION: Currently there is an issue with bone digs on crux due to the navmesh. This is a small fix to help combat that issue

DESCRIPTION: Increases dig distance slightly

TESTING: Tested using the dragon to dig up bones over time

Thanks to flamethrower for reporting the issue

### DIFF
--- a/dGame/dComponents/PetComponent.cpp
+++ b/dGame/dComponents/PetComponent.cpp
@@ -449,7 +449,7 @@ void PetComponent::Update(float deltaTime) {
 
 		NiPoint3 tresurePosition = closestTresure->GetPosition();
 		float distance = Vector3::DistanceSquared(position, tresurePosition);
-		if (distance < 3 * 3) {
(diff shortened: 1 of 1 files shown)


EmosewaMC (approved):

Wincent01 (approved):
Probably difficult to determine an accurate number here. Pet behavior is one of the biggest deviations in DLU from Live.

EmosewaMC:
Yeah this was a change I suggested to just fix it because of the navmesh issues.  Doesn't help that yeah, we dont do the live accurate pet thing of having them bounce on the dig spot and pressing shift to dig.  I figured this was the best solution so players get a less poor experience, even if it is super bodge.

PR #1152, opened 2023-07-22
fix: Remove usage of rand() and seed randomness for chatserver

EmosewaMC:
Replace it with the random engine mt19937.
convert time(0) to time(NULL)

Will test later.


EmosewaMC:
tested that ag survival still spawns enemies

Jettford (approved):

PR #1153, opened 2023-07-22
feat: Remove NET_VERSION define

EmosewaMC:
I get to test this on 4 different platforms.  yay...

Will test later, mostly because I have to do docker stuff to test.


EmosewaMC:
tested that config option gets updated in docker sharedconfig.ini as expected.

aronwk-aaron (approved):

DarwinAnim8or (dismissed):
I would explicitedly state however, that despite this being a variable, this is very much not going to guarantee that any version you set it to will actually work.

EmosewaMC:
> I would explicitedly state however, that despite this being a variable, this is very much not going to guarantee that any version you set it to will actually work.

added this as a comment.  Good catch

aronwk-aaron (approved):

Jettford (approved):

PR #1160, opened 2023-07-25
fix: Implement proper Sound trigger component serialization 

aronwk-aaron:
This implements full serialization of the sound trigger component:
tested that all the concert instruments still work (that is the only place this is used)

Additionally, adding handling for racing sound trigger use (although it's unused currently)
since it's the exact same as sound trigger for serialization and function


dGame/Entity.cpp
```diff
@@ -317,6 +317,9 @@ void Entity::Initialize() {
 	if (compRegistryTable->GetByIDAndType(m_TemplateID, eReplicaComponentType::SOUND_TRIGGER, -1) != -1) {
 		auto* comp = new SoundTriggerComponent(this);
 		m_Components.insert(std::make_pair(eReplicaComponentType::SOUND_TRIGGER, comp));
+	} else if (compRegistryTable->GetByIDAndType(m_TemplateID, eReplicaComponentType::RACING_SOUND_TRIGGER, -1) != -1) {
+		auto* comp = new SoundTriggerComponent(this);
+		m_Components.insert(std::make_pair(eReplicaComponentType::SOUND_TRIGGER, comp));
```
EmosewaMC, line 322: Ive thought about this a little bit more and this would make it impossible to get the racing sound trigger component as a component since it does not have ComponentType defined inline in a class.  Even if it simply extends the base SoundTriggerComponent class, we should have the RacingSoundTriggerComponent extend the SoundTriggerComponent simply so we can properly get the component if you request it.

Jettford (approved):

EmosewaMC (approved):

PR #1163, opened 2023-07-26
refactor: Vendor inventory loading

aronwk-aaron:
Implement proper delta compression
dynamically determine multicostitems and standard cost items 
Quarantine max's custom code

Tested that buying and selling from vendors work
Tested that crafting vendor works

### DIFF
--- a/dGame/dComponents/VendorComponent.cpp
+++ b/dGame/dComponents/VendorComponent.cpp
@@ -1,7 +1,8 @@
 #include "VendorComponent.h"
 #include "BitStream.h"
-#include "Game.h"
 #include "dServer.h"
+#include "dZoneManager.h"
+#include "WorldConfig.h"
 #include "CDComponentsRegistryTable.h"
 #include "CDVendorComponentTable.h"
 #include "CDLootMatrixTable.h"
@@ -32,12 +33,14 @@ void VendorComponent::OnUse(Entity* originator) {
 void VendorComponent::RefreshInventory(bool isCreation) {
 	SetHasStandardCostItems(false);
 	SetHasMultiCostItems(false);
+	m_Inventory.clear();
 
 	// Custom code for Max vanity NPC and Mr.Ree cameras
-	MaxCustomVendorRefreshInventory(isCreation);
-	if(m_Parent->GetLOT() == 9749 && Game::server->GetZoneID() == 1201) return;
+	if(isCreation && m_Parent->GetLOT() == 9749 && Game::server->GetZoneID() == 1201) {
+		SetupMaxCustomVendor();
+		return;
+	}
 
-	m_Inventory.clear();
 	auto* lootMatrixTable = CDClientManager::Instance().GetTable<CDLootMatrixTable>();
 	const auto lootMatrices = lootMatrixTable->Query([=](CDLootMatrix entry) { return (entry.LootMatrixIndex == m_LootMatrixID); });
 
@@ -54,6 +57,10 @@ void VendorComponent::RefreshInventory(bool isCreation) {
 			for (const auto& item : vendorItems) {
 				if (!m_HasStandardCostItems || !m_HasMultiCostItems) {
 					auto itemComponentID = compRegistryTable->GetByIDAndType(item.itemid, eReplicaComponentType::ITEM);
+					if (itemComponentID != -1) {
+						Game::logger->Log("VendorComponent", "Attempted to add item %i with ItemComponent ID -1 to vendor %i inventory. Not adding item!", itemComponentID, m_Parent->GetLOT());
+						continue;
+					}
 					auto itemComponent = itemComponentTable->GetItemComponentByID(itemComponentID);
 					if (!m_HasStandardCostItems && itemComponent.baseValue != -1) SetHasStandardCostItems(true);
 					if (!m_HasMultiCostItems && !itemComponent.currencyCosts.empty()) SetHasMultiCostItems(true);
@@ -70,7 +77,7 @@ void VendorComponent::RefreshInventory(bool isCreation) {
 				vendorItems.erase(vendorItems.begin() + randomItemIndex);
 				if (!m_HasStandardCostItems || !m_HasMultiCostItems) {
 					auto itemComponentID = compRegistryTable->GetByIDAndType(randomItem.itemid, eReplicaComponentType::ITEM, -1);
-					if (itemComponentID == -1) {
+					if (itemComponentID != -1) {
 						Game::logger->Log("VendorComponent", "Attempted to add item %i with ItemComponent ID -1 to vendor %i inventory. Not adding item!", itemComponentID, m_Parent->GetLOT());
 						continue;
 					}
@@ -82,6 +89,7 @@ void VendorComponent::RefreshInventory(bool isCreation) {
 			}
 		}
 	}
+	HandleMrReeCameras();
 
 	// Callback timer to refresh this inventory.
 	if (m_RefreshTimeSeconds > 0.0) {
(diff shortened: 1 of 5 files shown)


EmosewaMC (dismissed):
delayed review until 2026

aronwk-aaron:
my own joke thrown back at me, how could I see this coming

dGame/dComponents/VendorComponent.cpp
```diff
@@ -1,124 +1,96 @@
...
+	SetHasStandardCostItems(false);
+	SetHasMultiCostItems(false);
 
-bool VendorComponent::HasCraftingStation() {
-	// As far as we know, only Umami has a crafting station
-	return m_Parent->GetLOT() == 13800;
-}
+	// Custom code for Max vanity NPC and Mr.Ree cameras
+	MaxCustomVendorRefreshInventory(isCreation);
+	if(m_Parent->GetLOT() == 9749 && Game::server->GetZoneID() == 1201) return;
```
EmosewaMC, line 38: we should move the max method to be in this if statement and remove the if check in the other method since it should be private.

dGame/dComponents/VendorComponent.cpp
```diff
@@ -1,124 +1,96 @@
...
 	for (const auto& lootMatrix : lootMatrices) {
 		int lootTableID = lootMatrix.LootTableIndex;
-		std::vector<CDLootTable> vendorItems = lootTableTable->Query([=](CDLootTable entry) { return (entry.LootTableIndex == lootTableID); });
+		auto vendorItems = lootTableTable->Query([=](CDLootTable entry) { return (entry.LootTableIndex == lootTableID); });
 		if (lootMatrix.maxToDrop == 0 || lootMatrix.minToDrop == 0) {
-			for (CDLootTable item : vendorItems) {
-				m_Inventory.insert({ item.itemid, item.sortPriority });
+			for (const auto& item : vendorItems) {
+				if (!m_HasStandardCostItems || !m_HasMultiCostItems) {
+					auto itemComponentID = compRegistryTable->GetByIDAndType(item.itemid, eReplicaComponentType::ITEM);
```
EmosewaMC, line 56: we should check that the returned ID is not -1 (or in this case the default 0)

dGame/dComponents/VendorComponent.cpp
```diff
@@ -129,12 +101,40 @@ void VendorComponent::SetupConstants() {
 	auto* vendorComponentTable = CDClientManager::Instance().GetTable<CDVendorComponentTable>();
 	std::vector<CDVendorComponent> vendorComps = vendorComponentTable->Query([=](CDVendorComponent entry) { return (entry.id == componentID); });
 	if (vendorComps.empty()) return;
-	m_BuyScalar = vendorComps[0].buyScalar;
-	m_SellScalar = vendorComps[0].sellScalar;
-	m_RefreshTimeSeconds = vendorComps[0].refreshTimeSeconds;
-	m_LootMatrixID = vendorComps[0].LootMatrixIndex;
+	auto vendorData = vendorComps.at(0);
+	m_BuyScalar = vendorData.buyScalar;
```
EmosewaMC, line 105: lets correct this one to use the dZoneManager::Instance()->GetWorldConfig()->sell_thingy_idk_the_name

dGame/dComponents/VendorComponent.cpp
```diff
@@ -129,12 +101,40 @@ void VendorComponent::SetupConstants() {
...
+			break;
+		case 1:
+			camera = 16254; // Hipstabrick
+			break;
+		case 2:
+			camera = 16204; // Megabrixel snapshot
+			break;
+		}
+		m_Inventory.push_back(SoldItem(camera, 0));
+	}
```
EmosewaMC, line 139: if we're going to make a special function for these items, lets put them into two separate functions since they operate on two different vendors.  Initially the name implied to me this was for the Max vendor but I didn't know until I read it that this also affects Mr. Ree.

dGame/dComponents/VendorComponent.h
```diff
@@ -2,98 +2,66 @@
...
+		if (m_HasStandardCostItems == hasStandardCostItems) return;
+		m_HasStandardCostItems = hasStandardCostItems;
+		m_DirtyVendor = true;
+	}
+
+	bool SellsItem(const LOT item) const {
+		return std::count_if(m_Inventory.begin(), m_Inventory.end(), [item](const SoldItem& lhs) {
+			return lhs.lot == item;
+			}) > 0;
+	}
```
EmosewaMC, line 53: For moving these checkers to header files, we should generally only do so if it is small like the above `SetHasStandardCostItems`.  including the entire functional header in this header just to have this in the header file is a bit heavy of a cost

dGame/dComponents/VendorComponent.h
```diff
@@ -2,98 +2,66 @@
...
-	 */
 	void SetupConstants();
-
-	bool SellsItem(const LOT item) const;
+	float GetBuyScalar() const { return m_BuyScalar; }
+	float GetSellScalar() const { return m_SellScalar; }
+	void SetBuyScalar(const float value) { m_BuyScalar = value; }
+	void SetSellScalar(const float value) { m_SellScalar = value; }
+
+	std::vector<SoldItem>& GetInventory() { return m_Inventory; }
```
EmosewaMC, line 35: Should this be returning const?

dGame/dComponents/VendorComponent.h
```diff
@@ -2,98 +2,66 @@
...
+		m_DirtyVendor = true;
+	}
+
+	bool SellsItem(const LOT item) const {
+		return std::count_if(m_Inventory.begin(), m_Inventory.end(), [item](const SoldItem& lhs) {
+			return lhs.lot == item;
+			}) > 0;
+	}
+
+	void MaxCustomVendorRefreshInventory(bool isCreation = false);
```
EmosewaMC, line 55: Does this need to be public?

dGame/dComponents/VendorComponent.cpp
```diff
@@ -1,124 +1,96 @@
...
-
 	// Callback timer to refresh this inventory.
-	m_Parent->AddCallbackTimer(m_RefreshTimeSeconds, [this]() {
-		RefreshInventory();
-		});
+	if (m_RefreshTimeSeconds > 0.0) {
+		m_Parent->AddCallbackTimer(m_RefreshTimeSeconds, [this]() {
+			RefreshInventory();
+			}
+		);
```
EmosewaMC, line 91: I dont think we generally do a new line between the last argument of a function and the parenthesis, but this is a nit.

aronwk-aaron: I'll see what the standard is, I do it cause it's easier to see scope top to bottom

dGame/dComponents/VendorComponent.cpp
aronwk-aaron: I'm not 100% sure if that's correct, it looks like they may have been attempting to use the ones on the vendor component themselves rather than in worldconfig. Maybe worldconfig ones would be a default if none was provided

EmosewaMC: The live client physically does not use the sell or buy values in the db.  It only uses the worldconfig.

aronwk-aaron: ah cool

EmosewaMC: I stand correected.  I swear I looked up buyscalar and found no results.  Your presumptions about it defaulting to worldconfig was correct.

EmosewaMC (approved):
emmo approves

Wincent01 (approved):

PR #1164, opened 2023-07-26
refactor: Move CDClient Database away from constructor queries

EmosewaMC:
In an effort to move the tables to be namespaces as opposed to static classes, I am migrating all the database loading from memory to its own function so some testing experiments can be done.  This is a net win as you can at the least initialize the database without needing to actually have a cdclient file.

Memory usage before
image
and after
before

A slight decrease because I removed all cdclient storage that used pointers and just let the container manage the memory instead.  Broken CDCLIENT_CACHE_ALL queries were all fixed up.  ComponentsRegistry has the ability to run without needing to cache all of the table, however due to hot the Loot initializer works, not caching this table causes startup times to go to about 7-8 seconds due to the sheer number of queries all at once at world start.  Not caching the table reduces memory usage by 5-7%.

Tested that I can load into worlds and queries are succeeding without errors.


dDatabase/CDClientManager.cpp
```diff
@@ -38,43 +38,53 @@
 #include "CDFeatureGatingTable.h"
 #include "CDRailActivatorComponent.h"
 
+// Uncomment this to cache the full cdclient database into memory. This will make the server load faster, but will use more memory.
+// A vanilla CDClient takes about 46MB of memory + the regular world data.
+// #define CDCLIENT_CACHE_ALL
```
aronwk-aaron, line 43: this could be a sharedconfig option

aronwk-aaron (approved):

dDatabase/CDClientManager.cpp
aronwk-aaron: well
that's a stupid comment
I meant cmake variable

EmosewaMC: you should not need to rebuild the full server to define this imho

EmosewaMC: added the variable uncommented for now

aronwk-aaron (dismissed):

Jettford (dismissed):

DarwinAnim8or:
I'd test this more if I were you-- not caching things like skills or loot, would be a net negative on anything other than localhost

EmosewaMC:
> I'd test this more if I were you-- not caching things like skills or loot, would be a net negative on anything other than localhost

We are still caching stuff?  I didnt actually remove any caching, just renamed the function that does it to not be the constructor.  The perf save is from removal of extra storage

DarwinAnim8or (approved):

aronwk-aaron (approved):

PR #1165, opened 2023-07-27
Loot memory savings

### LINKED ISSUE #274: ENH: Fix double storage in LootGenerator

EmosewaMC:
- Remove all excess memory usage
- do not cache components registry
- cache loot matrices on startup of the destroyable component
- convert LootGenerator singleton class to namespace and consolidate Loot namespace definitions
- rework loot cdclient tables to operate closer to how someone would actually use them (basically doing the previous LootGenerator::LootGenerator caching but in those tables)
- Memory usage reduced by 7.5%+ across the board

No actual Loot generation logic was re-written, I strictly focused on removing the double storage.  Same loops, same return storages, same assumptions.  

Tested that a dragon, shooting gallery, vending machine and /rollloot still shows the same metrics as before.  

**Waiting on #1164**
Fixes #274

Because most loot tables and matrices are entities in spawners, smashables or vendors, most of the useful data is cached on world start with very few lookups done post players loading into a world.

### DIFF
--- a/CMakeVariables.txt
+++ b/CMakeVariables.txt
@@ -22,3 +22,5 @@ __maria_db_connector_compile_jobs__=1
 __enable_testing__=1
 # The path to OpenSSL.  Change this if your OpenSSL install path is different than the default.
 OPENSSL_ROOT_DIR=/usr/local/opt/openssl@3/
+# Uncomment the below line to cache the entire CDClient into memory
+# CDCLIENT_CACHE_ALL=1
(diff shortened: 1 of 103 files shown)


Jettford (approved):

aronwk-aaron (approved):

Jettford:
Makes sense, good change

PR #1170, opened 2023-07-29
fix: Properly read in scene metadata

aronwk-aaron:
Tested that nothing breaks, also correct some types


EmosewaMC (commented):
perion

EmosewaMC (approved):
rip perion

DarwinAnim8or (approved):

PR #1171, opened 2023-07-29
feature: add full known GM list

aronwk-aaron:
Add almost complete list of GM's for future use.
Also, this fixes stop2dambientsound, since it was sending the GM id for play instead of stop.

the test passes


EmosewaMC (approved):
build passes therefore :+1:

Jettford (approved):

PR #1177, opened 2023-08-05
chore: Add BitstreamUtils and make better read/write overloads

aronwk-aaron:
remove all raw packet reading
remove unneeded packet reading helpers
Read in full auth packets for computer info (The orginal motivation for making bettet bitstream utils instead using raw packets)


aronwk-aaron:
going to split this PR into multiple, since I've gone down a rabbit hole with chat stuff now

PR #1178, opened 2023-08-06
perf: Add path height correction on world load

EmosewaMC:
Because many waypoints for movement place the Y at some height far below the map, we need to correct this height to match that of the navmesh. Doing this at runtime is very expensive as some maps are very tall and so the halfExtents would need to be very large. Doing this at world load is a far more ideal place to take care of this as this only really needs to be done once so that at runtime, more approximate heights can be calculated with the same halfExtents as currently.

Tested that there are no changes aside from the height being changed at world startup.  This may resolve issues with some AI being unable to find a close waypoint on the navmesh due to their waypoint height being far too low for the halfExtents to allow finding of the nearest poly.

### DIFF
--- a/dNavigation/dNavMesh.cpp
+++ b/dNavigation/dNavMesh.cpp
@@ -143,11 +143,13 @@ float dNavMesh::GetHeightAtPoint(const NiPoint3& location, const float halfExten
 	}
 #endif
 	if (toReturn == 0.0f) {
-		toReturn = location.y;
-	// If we were unable to get the poly height, but the query returned a success, just use the height of the nearest point.
-	// This is what seems to happen anyways and it is better than returning zero.
-	} else if (hasPoly == DT_SUCCESS) {
-		toReturn = nearestPoint[1];
+		// If we were unable to get the poly height, but the query returned a success, just use the height of the nearest point.
+		// This is what seems to happen anyways and it is better than returning zero.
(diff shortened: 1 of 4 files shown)


aronwk-aaron (dismissed):
lgtm

EmosewaMC:
I am seeing more successful queries when using a X and Z halfExtents of zero.  about 5% more successful queries from before.

EmosewaMC:
This is resulting in 100% successful queries now.

EmosewaMC:
ran around quite a bit in gnarled forest, crux prime, return to venture explorer and saw no issues with enemies teleporting.

EmosewaMC:
The poly height always matching the one from rhe nearest reg is further reinforced by the lookup now being directly vertical with so variance in the x or z axis.  This ensures you find the correct height or find nothing

EmosewaMC:
Re-tested without capturing the res return from getPolyHeight and actually returning the correct value and no changes except from the expected path fixes.

aronwk-aaron (approved):

DarwinAnim8or (dismissed):

Wincent01 (dismissed):

aronwk-aaron (dismissed):

aronwk-aaron (approved):

DarwinAnim8or (approved):

PR #1180, opened 2023-08-08
chore: Aggregate initializers for humongous script constants

EmosewaMC:
wow this is so much more readable


aronwk-aaron (approved):

EmosewaMC:
Because of how aggregate initializers work, you must initialize the variables in the same order as they are initialized in the constructor.  This compiles and runs fine.  I used multi line editing to copy and paste the initializer line by line and then regex for some

EmosewaMC:
my disappointment is immeasurable and my day is ruined

PR #1183, opened 2023-08-12
fix: dangling pointer in mission when saving character data

EmosewaMC:
Fixes an issue where characters would lose mission progress due to a dangling pointer in mission data.
This was proven by adding a logging message to the destructor of all CD tables which revealed that when crashing, the tables and all their data are destroyed _first_ and then character data is saved.  This causes all missions to enter a state where their pointers are dangling and character data will enter an invalid state.

This is a quick patch to get character save data stable and remove the dangling pointer.

Tested that previously, I would lose my mission progress with 3 clients in world `1251 1` every time on multiple characters and after the patch I would never lose save data.

Image of save data with fragmented memory, as you can see, the missionID is a completely invalid number
image

### DIFF
--- a/dGame/dMission/Mission.cpp
+++ b/dGame/dMission/Mission.cpp
@@ -42,9 +42,10 @@ Mission::Mission(MissionComponent* missionComponent, const uint32_t missionId) {
 
 	auto* missionsTable = CDClientManager::Instance().GetTable<CDMissionsTable>();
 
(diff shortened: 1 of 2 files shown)


EmosewaMC:
Yes there are memory usage increases.  player character safety >>>>> a bit of memory savings right now.

Wincent01 (approved):

DarwinAnim8or (approved):

PR #1187, opened 2023-08-18
fix: Limit the number of players that can join a team

aronwk-aaron:
after being invited

Tested that the server compiles and does not crash

### DIFF
--- a/dChatServer/PlayerContainer.cpp
+++ b/dChatServer/PlayerContainer.cpp
@@ -207,6 +207,11 @@ TeamData* PlayerContainer::GetTeam(LWOOBJID playerID) {
 }
 
 void PlayerContainer::AddMember(TeamData* team, LWOOBJID playerID) {
+	if (team->memberIDs.size() >= 4){
+		Game::logger->Log("PlayerContainer", "Tried to add player to team that already had 4 players");
+		return;
+	}
+
 	const auto index = std::find(team->memberIDs.begin(), team->memberIDs.end(), playerID);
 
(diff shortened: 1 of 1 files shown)


EmosewaMC (approved):

Wincent01 (commented):
Is there a way to notify the affected parties that this has happened? If not through something else, a chat message would probably be good.

aronwk-aaron:
> Is there a way to notify the affected parties that this has happened? If not through something else, a chat message would probably be good.

Done

dChatServer/PlayerContainer.cpp
```diff
@@ -209,6 +210,8 @@ TeamData* PlayerContainer::GetTeam(LWOOBJID playerID) {
 void PlayerContainer::AddMember(TeamData* team, LWOOBJID playerID) {
 	if (team->memberIDs.size() >= 4){
 		Game::logger->Log("PlayerContainer", "Tried to add player to team that already had 4 players");
+		auto* player = GetPlayerData(playerID);
+		ChatPackets::SendSystemMessage(player->sysAddr, u"The teams is full! You have not been added to a team!");
```
EmosewaMC, line 214: Check your pointers.

EmosewaMC (changes requested):

dChatServer/PlayerContainer.cpp
aronwk-aaron: tru

EmosewaMC (approved):

Jettford (approved):

PR #1202, opened 2023-09-29
fix: remove foreign key on name in cheat_detection table

aronwk-aaron:
Drop table incase it exists (which sometimes it doesn't) and then make it again without that foreign key
Tested on 10.11.5-MariaDB

Dropping table since there's little to no chance that data has gotten in it yet, on the chance it even exists


EmosewaMC (changes requested):
The old data needs to be moved over first.

EmosewaMC:
alternatively, we should instead just remove the foreign key by name.  this should be `player_cheat_detections_ibfk_1`

EmosewaMC:
```
ALTER TABLE player_cheat_detections DROP FOREIGN KEY `player_cheat_detections_ibfk_1`;
```
this should work instead.

EmosewaMC (approved):
Only suggested change is to add a comment saying why we are dropping a table in a sql query for future readers.

Wincent01 (approved):

PR #1204, opened 2023-09-30
feat: Add isolated and simplified path to add components

EmosewaMC:
Follows a structure similar to std::thread in that you say what component you want to add in the Template types, and what you want to pass to the constructor in the rest of the variadic template values.  This is a catch all solution and due to forwarding, is basically a free function call idenitcal to the old ones, but now all the newing and inserting is done in one place!  Even replacing the component with a new one is all done at once.

before: 
```
auto* newComp = new DerivedComp(this, arg1, arg2, arg3);
components.insert(pair<DerivedComp::ComponentType, newComp);
```
This had to be repeated for every component and is prone to error as well as weak to large changes (as you'll see in future changes)
after:
```
AddComponent<DerivedComp>(arg1, arg2, arg3);
```
since the component always needs to know its parent Entity, we always default the first argument to being the Entity and dont make the user specify this.  Then the other 3 arguments have their [template types deduced](https://en.cppreference.com/w/cpp/language/class_template_argument_deduction) and the component is also added to the map for you, all in 1 step.  If you try to add a component again, instead of deleting the previous one, we simply use a placement new to basically reset the current pointer to its constructed state.  This allows us to skip a delete and free and keeps the head code path at minimal branches (1 in our code and the normal amount in the map insert anyways).

A small test for the placement new is attached as well.

Tested that I could load into Venture Explorer, Forbidden Valley Race, Cannon Cove Shooting Gallery and Crux Prime and that the tests pass.


aronwk-aaron (approved):

Jettford (approved):

aronwk-aaron (dismissed):

EmosewaMC:
Need to add manual destructor call to placement newd component.

aronwk-aaron (dismissed):

aronwk-aaron (approved):

Jettford (approved):
You scare me.

PR #1206, opened 2023-10-02
chore: Use TryParse for LDF parsing

EmosewaMC:
Simplifies the work we are doing by using a common known working and tested function.

Tests still pass and server still reads all ldf data from luz files.

### DIFF
--- a/tests/dCommonTests/TestLDFFormat.cpp
+++ b/tests/dCommonTests/TestLDFFormat.cpp
@@ -17,7 +17,7 @@ protected:
 	}
 };
 
-#define LdfUniquePtr std::unique_ptr<LDFBaseData>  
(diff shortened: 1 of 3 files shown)


aronwk-aaron (approved):
LGTM

tests/dCommonTests/TestLDFFormat.cpp
```diff
@@ -17,7 +17,7 @@ class LDFTests : public dCommonDependenciesTest {
 	}
 };
 
-#define LdfUniquePtr std::unique_ptr<LDFBaseData>  
+typedef std::unique_ptr<LDFBaseData> LdfUniquePtr;
```
Jettford, line 20: Yummy

Jettford (approved):

PR #1207, opened 2023-10-02
feat: Abstract Logger and simplify code

EmosewaMC:
Need to read this PR at a later time to make sure there are no issues with the macro replacement

unix and windows writers need to be macro'd away on their respective systems and the generic ConsoleWriter should work for all consoles, regardless of OS to allow for a backup way to log to a console on a system that has specific color console logging

- Renamed class to Logger from dLogger.
- Abstract logging to base classes to isolate where the writing happens for future logging ventures where you may want to write to multiple files, or maybe you want to do color console logging which is platform dependent.  It'll be good to have this stuff isolated and in their own space.
- Macros introduce compile time calculation of the file name and line number alongside the previous info of time and date.
    - This allows the user to spam logs in a file without worry about having to put identifiers on the logs.

One downside is the binaries are going to be a touch larger now with absolute paths in them, however for WorldServer, the largest server, I only observe a 1% increase in binary size, which is a very small difference and worth the extra details and removal of ambiguity.

Tested that starting up a server and doing some property behavior messages that all logs appeared as expected.


aronwk-aaron (dismissed):

dCommon/Logger.cpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+void Logger::vLog(const char* format, va_list args) {
+	time_t t = time(NULL);
+	struct tm* time = localtime(&t);
+	char timeStr[70];
+	strftime(timeStr, sizeof(timeStr), "[%d-%m-%y %H:%M:%S ", time);
+	char message[2048];
+	vsnprintf(message, 2048, format, args);
+	std::for_each(m_Writers.begin(), m_Writers.end(), [&](const std::unique_ptr<Writer>& writer) {
```
Jettford, line 57: Why is there std::for_each being used here? Surely iterator based for loops, complete the same job?

dCommon/Logger.h
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+    return file;
+}
+
+// These have to have a constexpr variable to store the filename_and_line result in a local variable otherwise
+// they will not be valid constexpr and will be evaluated at runtime instead of compile time!
+// The full string is still stored in the binary, however the offset of the filename in the absolute paths
+// is used in the instruction instead of the start of the absolute path.
+#define LOG(message, ...) do { auto str = FILENAME_AND_LINE; Game::logger->Log(str, message, ##__VA_ARGS__); } while(0)
```
Jettford, line 32: Can you explain the `do` and `while(0)` here?

dCommon/Logger.h
```diff
@@ -0,0 +1,110 @@
...
+
+// Writer class for writing data.
+
+enum class WriterType : uint32_t {
+	Writer,
+	FileWriter,
+	ConsoleWriter
+};
+
+class Writer {
```
Jettford, line 43: I understand the breaking out of the Writer class but why the two different classes for Console and File? Surely it would be cleaner to just have on writer and both output to a FILE*, one of the FILE*'s obviously being stdout/stderr

dCommon/Logger.h
EmosewaMC: prevents a swallowing the semicolon issue possibility - compilers will optimize away the loop entirely so its a net gain for little cost (compiler optimzation)

dCommon/Logger.cpp
EmosewaMC: Yes, will swap this out

dCommon/Logger.h
EmosewaMC: Will move this and create a destructor for the file writer to close its file

EmosewaMC: The two different classes are so you can actually tell a console writer from a file writer, otherwise there is no difference aside from the destination of the file*

EmosewaMC: Reworked how this is done

aronwk-aaron (dismissed):

aronwk-aaron (approved):

Jettford (approved):
I am approving this PR with the belief that its benefit is greater than my issues with it but I do want it noted that I don't agree with some of the approach of the Logger changes.

EmosewaMC:
> I am approving this PR with the belief that its benefit is greater than my issues with it but I do want it noted that I don't agree with some of the approach of the Logger changes.

This is only iteration 1, further iterations will work closer with the writers to add color logging so we'll touch those again very soon

PR #1209, opened 2023-10-05
CheatDetection: Move player access

EmosewaMC:
in earlier blocks, we check if player is null before accessing it.  In doing so I neglected to propogate the player access for who to report into said blocks.  This will resolve crashes from null accesses.


aronwk-aaron (approved):

aronwk-aaron (approved):

Jettford (approved):

Jettford:
Looks fine, simple oversight.

PR #1211, opened 2023-10-06
fix: Address crash in Cannon Cove Shooting Gallery

EmosewaMC:
Tested that the gallery does not crash anymore when trying to access index 5 of a 5 size vector.


aronwk-aaron (approved):
LGTM
Fixes crashes, but now we gotta fix other stuff cause replays give more score due to not despawning all the targets between games

Jettford (approved):

PR #1212, opened 2023-10-06
fix: Secure Master packet handling

EmosewaMC:
Moves master packets out of the main packet handler and into its own function specifically for master packets to prevent spoofing

Tested that master packets are still routed correctly.


EmosewaMC:
So now this is functional, however I am not happy with how it looks.  I will look to abstract away this information to a packet handler for MasterPackets as opposed to making dServer handle it

PR #1215, opened 2023-10-07
Diagnostics: Fix demangling

EmosewaMC:
Rename to better names
More comments
Less branches


EmosewaMC:
Tested that crashes now demangle names when possible and print the full crash line if no demangling is doable.

aronwk-aaron (approved):
LGTM

Jettford (approved):

PR #1216, opened 2023-10-09
Scripts: Fix ambiguous base class

EmosewaMC:
Fixes an issue where we needed a C style cast to save these scripts due the the diamond problem wherein the compiler was unable to determine which CppScripts base class to inherit from with multiple inheritance of the same base class.
Moved globals to unnammed namespace
removed empty script logging as missing

Tested that scripts still compile.
Also cache the script that was being raw returned


aronwk-aaron (approved):
LGTM

Jettford (approved):

PR #1219, opened 2023-10-14
fix: Address race condition

EmosewaMC:
Fix a race condition that would overwrite the port of zone 0 with the auth server port or a chat server port which would cause a domino effect of desynced server ports

Tested that a sleep(5) in auth still allows a player to connect and login without issues


aronwk-aaron (approved):
gotta go [fast](https://www.deviantart.com/kiribbean/art/Sanic-404-gif-716079737)

DarwinAnim8or (approved):
lgtm

PR #1227, opened 2023-10-18
fix: Animated deaths not playing

### LINKED ISSUE #1222: BUG: Enemies with special death animations do not get to play the animation before being destructed

### DarkflameServer Version

All

### Issue Description

Some enemies in LU like #515 and some alpha enemies are supposed to play an animation on death, however we currently do not implement death_behavior correctly, and as such, these enemies do not play their animation before being destructed from the server.

### Reproduction Steps

1 Finish a property mission
2 turn the mission in
3 property guard does not get vacuumed away but rather gets destructed instantly

### Expected Behavior

Property guard gets vacuumed away fully before the destruction is sent

### Environment

All

EmosewaMC:
Fixes #1222

addresses an issue where the death behavior of a destructible component was not being respected and enemies with destroyable components that had special death animations were not able to play the animation on death.  This pr adds in the hardcoded constant the client uses for the same metric of 12 seconds.

Tested that claiming Nimbus Rock and completing the property guards mission allows him to vacuum away and then network the destruction packet 12 seconds later.

### DIFF
--- a/dScripts/ai/PROPERTY/AgPropguards.cpp
+++ b/dScripts/ai/PROPERTY/AgPropguards.cpp
@@ -18,7 +18,7 @@ void AgPropguards::OnMissionDialogueOK(Entity* self, Entity* target, int mission
 		&& !character->GetPlayerFlag(flag)) {
 		// If the player just started the mission, play a cinematic highlighting the target
 		GameMessages::SendPlayCinematic(target->GetObjectID(), u"MissionCam", target->GetSystemAddress());
-	} else if (missionState == eMissionState::COMPLETE_READY_TO_COMPLETE) {
+	} else if (missionState == eMissionState::READY_TO_COMPLETE) {
 		// Makes the guard disappear once the mission has been completed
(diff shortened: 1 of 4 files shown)


dScripts/ai/PROPERTY/AgPropguards.cpp
```diff
@@ -18,7 +18,7 @@ void AgPropguards::OnMissionDialogueOK(Entity* self, Entity* target, int mission
 		&& !character->GetPlayerFlag(flag)) {
 		// If the player just started the mission, play a cinematic highlighting the target
 		GameMessages::SendPlayCinematic(target->GetObjectID(), u"MissionCam", target->GetSystemAddress());
-	} else if (missionState == eMissionState::COMPLETE_READY_TO_COMPLETE) {
+	} else if (missionState == eMissionState::READY_TO_COMPLETE) {
```
EmosewaMC, line 21: Yes this is correct now.  Live script uses 4, not 12 here.

Wincent01 (approved):

PR #1228, opened 2023-10-19
Entity: Add missing destroyable case

EmosewaMC:
Adds a missing edge case on the client where you may have the is_smashable flag set, but dont actually have a destroyable component.  In this case you should create only a destroyable component and add it to the component and serialize it last after all other data.  We also needed to add the set_faction var reading and setting as this is how Oliver Sudden actually gets their faction since they have no actual information in the database, it is all stored in ldf keys.

Tested that Oliver Sudden no longer logs Unable to unserialize logs when serialized or constructed.


aronwk-aaron (approved):

Wincent01 (approved):
The procedure for adding components ain't trivial

PR #1234, opened 2023-10-23
fix: iterator invalidation in EntityManager

EmosewaMC:
While we were no longer modifying these std containers in these loops directly, this did not stop external code from accessing and modifying the containers, invalidating the contained iterators.

Tested that servers still start up, and that zones like bons no longer hard crash when all players have left the world.


EmosewaMC:
Crash logs varied from a call stack leading directly to this function, all the way to a completely empty call stack aside from main.

aronwk-aaron (approved):

PR #1246, opened 2023-10-30
refactor: Property Entrance Component 

Jettford:
Tested and working on Windows.


dDatabase/Tables/CDLootTableTable.cpp
```diff
@@ -24,6 +24,7 @@ void SortTable(LootTableEntries& table) {
 				lootToInsert = oldItrInner;
 			}
 		}
+
```
EmosewaMC, line 27: can get rid of this probably

dGame/dGameMessages/PropertyData.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
+#include "Entity.h"
+
+
+struct PropertyPersonalData final {
+	bool IsBestFriend;
```
EmosewaMC, line 7: Generally our members of structs and classes follow `m_FieldName`for member fields, is there a particular reason this isnt done here?

dGame/dGameMessages/GameMessages.cpp
```diff
@@ -3099,10 +3099,41 @@ void GameMessages::SendPropertySelectQuery(LWOOBJID objectId, int32_t navOffset,
...
-	for (auto& entry : entries) {
-		entry.Serialize(bitStream);
-	}
+	for (const auto& item : entries) {
+		bitStream.Write(item.CloneID);
+
+		const auto& owner = GeneralUtils::UTF8ToUTF16(item.PrimaryData.OwnerName);
+		bitStream.Write(uint32_t(owner.size()));
+		for (uint32_t i = 0; i < owner.size(); ++i) {
+			bitStream.Write(static_cast<uint16_t>(owner[i]));
```
EmosewaMC, line 3108: I don't believe we need the static casts since owner is a u16string which has a CharT of char16_t and can likely be moved to the T type of the Write method instead since no conversion needs to happen. What do you think?

dGame/dComponents/PropertyEntranceComponent.cpp
```diff
@@ -74,261 +67,222 @@ void PropertyEntranceComponent::OnEnterProperty(Entity* entity, uint32_t index,
...
+	propertyLookup->setUInt(1, this->m_MapID);
+	propertyLookup->setString(2, searchString.c_str());
+	propertyLookup->setString(3, searchString.c_str());
+	propertyLookup->setString(4, searchString.c_str());
+	propertyLookup->setInt(5, sortMethod == SORT_TYPE_FEATURED || sortMethod == SORT_TYPE_FRIENDS ? (uint32_t)PropertyPrivacyOption::Friends : (uint32_t)PropertyPrivacyOption::Public);
+
+	auto propertyIdRes = propertyLookup->executeQuery();
 
-		auto friendsListQueryResult = friendsListQuery->executeQuery();
+	std::vector<uint32_t> propertyIds{};
```
EmosewaMC, line 118: nit: You dont need an initializer list constructor here, std::vector on the stack will already be default initialized

dGame/dComponents/PropertyEntranceComponent.cpp
```diff
@@ -74,261 +67,222 @@ void PropertyEntranceComponent::OnEnterProperty(Entity* entity, uint32_t index,
...
-		friendsListQuery->setUInt(1, character->GetID());
-		friendsListQuery->setUInt(2, character->GetID());
+	const auto searchString = "%" + searchText + "%";
+	propertyLookup->setUInt(1, this->m_MapID);
+	propertyLookup->setString(2, searchString.c_str());
+	propertyLookup->setString(3, searchString.c_str());
+	propertyLookup->setString(4, searchString.c_str());
+	propertyLookup->setInt(5, sortMethod == SORT_TYPE_FEATURED || sortMethod == SORT_TYPE_FRIENDS ? (uint32_t)PropertyPrivacyOption::Friends : (uint32_t)PropertyPrivacyOption::Public);
+
+	auto propertyIdRes = propertyLookup->executeQuery();
```
EmosewaMC, line 116: nit: generally, we try to place a * in front of pointer autos

dGame/dComponents/PropertyEntranceComponent.cpp
```diff
@@ -74,261 +67,222 @@ void PropertyEntranceComponent::OnEnterProperty(Entity* entity, uint32_t index,
...
 
-		auto friendsListQueryResult = friendsListQuery->executeQuery();
+	std::vector<uint32_t> propertyIds{};
 
-		while (friendsListQueryResult->next()) {
-			auto playerIDToConvert = friendsListQueryResult->getInt(1);
-			friendsList = friendsList + std::to_string(playerIDToConvert) + ",";
+	while (propertyIdRes->next()) {
+		if (sortMethod == SORT_TYPE_FRIENDS || sortMethod == SORT_TYPE_FEATURED) {
+			if (m_UserFriendMap[(uint32_t)requestor->GetObjectID()].find(propertyIdRes->getUInt(2)) == m_UserFriendMap[(uint32_t)requestor->GetObjectID()].end()) {
```
EmosewaMC, line 122: This statement is a bit long, we could reduce it to `m_UserFriendMap[static_cast<uint32_t>(requestor->GetObjectID())].count(propertyIdRes->getUInt(2)) == 0` so that we dont need to do the same work twice here.  I would also recommend replacing the cast to uint32_t from an object id (which clearly means to be a characterId, to getting the characterid from the character as that is the intention of the code at a glance.

dGame/dComponents/PropertyEntranceComponent.h
```diff
@@ -59,15 +61,17 @@ class PropertyEntranceComponent : public Component {
 	 */
 	[[nodiscard]] LWOMAPID GetMapID() const { return m_MapID; };
 
-	PropertySelectQueryProperty SetPropertyValues(PropertySelectQueryProperty property, LWOCLONEID cloneId = LWOCLONEID_INVALID, std::string ownerName = "", std::string propertyName = "", std::string propertyDescription = "", float reputation = 0, bool isBFF = false, bool isFriend = false, bool isModeratorApproved = false, bool isAlt = false, bool isOwned = false, uint32_t privacyOption = 0, uint32_t timeLastUpdated = 0, float performanceCost = 0.0f);
-
-	std::string BuildQuery(Entity* entity, int32_t sortMethod, Character* character, std::string customQuery = "", bool wantLimits = true);
+	void PopulateUserFriendMap(uint32_t user);
 
+	std::vector<uint32_t> GetPropertyIDsBasedOnParams(const std::string& searchText, uint32_t sortMethod, Entity* requestor);
```
EmosewaMC, line 66: BasedOnParams seems a bit too explicit here and I dont believe it helps tell a reader what the function does.  We could probably just remove the BasedOnParams and describe in a comment block how the function arguments affect the lookup

dGame/dComponents/PropertyEntranceComponent.cpp
```diff
@@ -74,261 +67,222 @@ void PropertyEntranceComponent::OnEnterProperty(Entity* entity, uint32_t index,
...
+	return propertyIds;
 }
 
+
 void PropertyEntranceComponent::OnPropertyEntranceSync(Entity* entity, bool includeNullAddress, bool includeNullDescription, bool playerOwn, bool updateUi, int32_t numResults, int32_t lReputationTime, int32_t sortMethod, int32_t startIndex, std::string filterText, const SystemAddress& sysAddr) {
+	std::vector<PropertyData> entries{};
 
-	std::vector<PropertySelectQueryProperty> entries{};
-	PropertySelectQueryProperty playerEntry{};
+	if (m_UserFriendMap.find((uint32_t)entity->GetObjectID()) == m_UserFriendMap.end()) {
```
EmosewaMC, line 143: I don't believe you can cache the friends list here unfortuntately, because a player can make more friends in this world and then re-open the menu to try and visit their new friends property.  Is this scenario taken into account somewhere?

dGame/dComponents/PropertyEntranceComponent.cpp
```diff
@@ -74,261 +67,222 @@ void PropertyEntranceComponent::OnEnterProperty(Entity* entity, uint32_t index,
...
-		const auto owner = propertyEntry->getInt(2);
-		const auto cloneId = propertyEntry->getUInt64(4);
-		const auto propertyNameFromDb = std::string(propertyEntry->getString(5).c_str());
-		const auto propertyDescriptionFromDb = std::string(propertyEntry->getString(6).c_str());
-		const auto privacyOption = propertyEntry->getInt(9);
-		const auto modApproved = propertyEntry->getBoolean(10);
-		const auto dateLastUpdated = propertyEntry->getInt(11);
-		const float reputation = propertyEntry->getInt(14);
-		const auto performanceCost = (float)propertyEntry->getDouble(16);
+	std::vector<uint32_t> propertyCloneIdSlice{};
```
EmosewaMC, line 167: nit: Same as above vector

dGame/dComponents/PropertyEntranceComponent.cpp
```diff
@@ -74,261 +67,222 @@ void PropertyEntranceComponent::OnEnterProperty(Entity* entity, uint32_t index,
...
+}
 
-		auto nameResult = nameLookup->executeQuery();
 
-		if (!nameResult->next()) {
-			delete nameLookup;
-			nameLookup = nullptr;
+PropertyData PropertyEntranceComponent::GetPropertyData(uint32_t propertyID) {
+	if (m_PropertyDataCache.find(propertyID) != m_PropertyDataCache.end()) {
+		return m_PropertyDataCache[propertyID];
```
EmosewaMC, line 186: if a user updates their property name, description or has their player name updated while the data here is cached, when is this updated?

dGame/dComponents/PropertyEntranceComponent.h
```diff
@@ -59,15 +61,17 @@ class PropertyEntranceComponent : public Component {
...
+	PropertyPersonalData GetPropertyPersonalData(PropertyData& propertyData, Entity* queryingUser, bool updatePropertyDataStructure = false);
 private:
-	/**
-	 * Cache of property information that was queried for property launched, indexed by property ID
-	 */
-	std::map<LWOOBJID, std::vector<PropertySelectQueryProperty>> propertyQueries;
+	std::unordered_map<LWOOBJID, std::vector<uint32_t>> m_UserRequestedCloneMap;
+	std::unordered_map<uint32_t, PropertyData> m_PropertyDataCache;
+
+	std::unordered_map<LWOOBJID, std::unordered_map<LWOOBJID, bool>> m_UserFriendMap;
```
EmosewaMC, line 74: an unordered map of bool can be exchanged with unordered set to save memory and readability, however there are concerns above on if this can even be used due to this data not being updated after the initial query.  The same applies for the propertyDataCache. Are both of these updated if there is a change in the users friend list or if there is a change in the property status?  for example if you removed a friend they could visit your property still if you never update the cache here until the world shutsdown, and someone could make an unapproved property and then fly to it because the property was approved previously.  Are these concerns resolved somewhere else that I missed?

dGame/dComponents/PropertyEntranceComponent.cpp
```diff
@@ -74,261 +67,222 @@ void PropertyEntranceComponent::OnEnterProperty(Entity* entity, uint32_t index,
...
-		if (friendsList.at(friendsList.size() - 1) == ',') friendsList.erase(friendsList.size() - 1, 1);
-		friendsList += ") ";
 
-		// If we have no friends then use a -1 for the query.
-		if (friendsList.find("()") != std::string::npos) friendsList = " AND p.owner_id IN (-1) ";
+		propertyIds.push_back(propertyIdRes->getUInt(1));
+	}
 
-		orderBy += friendsList + "ORDER BY ci.name ASC ";
+	delete propertyIdRes;
```
EmosewaMC, line 130: nit: Unique_ptrs on all queries and results to reduce clutter sloc and improve memory safety and not need to worry about doing deletes in other return paths as you do below

dGame/dComponents/PropertyEntranceComponent.cpp
```diff
@@ -74,261 +67,222 @@ void PropertyEntranceComponent::OnEnterProperty(Entity* entity, uint32_t index,
...
+}
 
-		delete isAltQueryResults;
-		isAltQueryResults = nullptr;
+PropertyPersonalData PropertyEntranceComponent::GetPropertyPersonalData(PropertyData& propertyData, Entity* queryingUser, bool updatePropertyDataStructure) {
+	PropertyPersonalData personalData{};
 
-		delete isAltQuery;
-		isAltQuery = nullptr;
+	personalData.IsFriend = m_UserFriendMap[(uint32_t)queryingUser->GetObjectID()].find(propertyData.PrimaryData.OwnerID) != m_UserFriendMap[(uint32_t)queryingUser->GetObjectID()].end();
```
EmosewaMC, line 248: If the above logic follows, this should attempt to fill the user friend map if the user is not found in it correct?  would be a good safeguard for future users unless this is a private method upon which this can be dismissed.  Same goes for the other new methods; can these be made private/const if possible?

EmosewaMC (commented):
general feedback so far with mostly questions, and a couple small changes (see the unordered_map -> unordered_set and long winded if statements)

aronwk-aaron (approved):

dGame/dGameMessages/PropertyData.h
aronwk-aaron: we've followed this strictly for members of classes, but not for structs

dGame/dGameMessages/GameMessages.cpp
aronwk-aaron: will agree, the cast is unnecessary, since the are all already u16strings

EmosewaMC:
superceded by #1650 with an orm-like implementation.

PR #1247, opened 2023-10-30
Fix edge case where leaderboard viewing would cause unhandled exception

maxdelayer:
It was possible for the query to get the leaderboard info to cause an unhandled exception in WorldServer and crash it. Based on my testing, this one-liner mitigates it, though in the long term handling any potential database exceptions might be useful.


EmosewaMC:
Please also test empty leaderboards, and various other placements of the player in a leaderboard such as in the bottom 3, in the top 3, a leaderboard only having < 11 entries and ensure the data is displayed correctly

EmosewaMC (approved):
myRank should also not be negative and this is more congruent.

DarwinAnim8or (approved):
works around the crash, at least

PR #1252, opened 2023-11-04
chore: Notes for running as system service

Coolfeather2:
Added note for running the server as a system service

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,49 +227,35 @@ and then go to `build/masterconfig.ini` and change `use_sudo_auth` to 0.
 ### Linux Service
 If you are running this on a linux based system, it will use your terminal to run the program interactively, preventing you using it for other tasks and requiring it to be open to run the server.  
 _Note: You could use screen or tmux instead for virtual terminals_  
-To run the server non-interactively, we can use a systemctl service by creating the following file:  
-`/etc/systemd/system/darkflame.service`  
-with the contents of:
-```ini
-[Unit]
-# Description of the service.
-Description=Darkflame LEGO Universe Server
-# Wait for network to start first before starting this service.
-After=network.target
+To run the server non-interactively, we can use a systemctl service by copying the following file:  
+```shell
+cp ./systemd.example /etc/systemd/system/darkflame.service
+```  
 
-[Service]
-Type=simple
-# Services should have their own, dedicated, user account.
-# The specific user that our service will run as, you can find your current user by issuing `id -un`
-User=darkflame
-# The specific group that our service will run as, you can find your current primary group by issuing `id -gn`
-Group=darkflame
-# Full Path to the darkflame server process
-ExecStart=/PATH/TO/DarkflameServer/build/MasterServer
+Make sure to edit the file in `/etc/systemd/system/darkflame.service` and change the:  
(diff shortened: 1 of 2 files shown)


EmosewaMC (commented):
as an FYI, DLU is not a unix only server and these steps are not really explained what they are for and why.  If this is continued to be worked on, I recommend specifying this is unix only, and commenting each line and why they are necessary to keep the guide as free from assumptions of knowledge as possible.

Coolfeather2:
I think its put together better now, removed reference to root for security purposes, and explained how to use it.  
In the systemd description I use "LEGO", I am not sure if this is a trademark/legal issue and needs removing?

EmosewaMC:
As I am unfamiliar with systemctl, is there a way to have this kind of file in the DLU repo instead and direct the systemctl to look at that file for the startup instead?

Coolfeather2:
The daemon service is usually installed either during build process or package installation (when systemd is detected) into the `/lib/systemd/system` directory (as `/lib/systemd` is used for distributions and `/etc/systemd` is used for admins) but as we are not distibuting packages. <sup>[1]( https://0pointer.de/public/systemd-man/daemon.html )</sup>

We can distibute the service file in the repository, and you can link it directly using the full path via `systemctl link /FULL/PATH/DarkflameServer/systemd/darkflame.service`. <sup>[2]( https://0pointer.de/public/systemd-man/systemctl.html)</sup>
Would still need to advise people to edit/change the repo provided service still (change user and path)
Instead I could also write a script to automate/template it to fill out the required information? (set user to current run context and directory of repo)
If we automate/template it would probably be best practice to copy it to the `/etc/systemd` directory once generated.

Let me know if you want to proceed:

- Provide service file in repo and instructions on editing and linking (or copying) to systemd
- Create bash script to template the systemd (fill out path and user) and copy (or link) to `/etc/systemd`


Reference documentation researched:
1: https://0pointer.de/public/systemd-man/daemon.html 
2: https://0pointer.de/public/systemd-man/systemctl.html
3: https://fedoraproject.org/wiki/Packaging:Systemd#Unit_Files

aronwk-aaron (dismissed):

README.md
```diff
@@ -224,6 +224,58 @@ sudo setcap 'cap_net_bind_service=+ep' AuthServer
 ```
 and then go to `build/masterconfig.ini` and change `use_sudo_auth` to 0.
 
+### Linux Service
+If you are running this on a linux based system, it will use your terminal to run the program interactively, preventing you using it for other tasks and requiring it to be open to run the server.  
+_Note: You could use screen or tmux instead for virtual terminals_  
+To run the server non-interactively, we can use a systemctl service by creating the following file:  
+`/etc/systemd/system/darkflame.service`  
+with the contents of:
+```ini
```
EmosewaMC, line 233: What would be your thoughts on moving this to an exmaple file, something like systemd.example, and then having users rename/move that file to the correct name instead of having it in the readme as is?  That way they can just copy and rename the file and make the edits for the process.

aronwk-aaron: yeah, this would be good

Coolfeather2: Applied changes

aronwk-aaron (approved):

EmosewaMC (approved):
lgtm

PR #1253, opened 2023-11-04
fix: Fix foreign key constraints

EmosewaMC:
Sigh... so much for foreign keys

Tested that you can still travel between worlds and the database no longer faults for trying to insert a user of id 0 into cheat detection table


aronwk-aaron (changes requested):
Why are we inserting 0? just make it null instead.
The FK's are important. The FK doesn't have a NOT NULL on it, so null is valid input for that column

EmosewaMC:
> Why are we inserting 0? just make it null instead. The FK doesn't have a NOT NULL on it, so null is valid input for that column

by this same logic we should have done the same for the player name field.

aronwk-aaron (approved):

Wincent01 (approved):

PR #1265, opened 2023-11-06
fix: Ugc Remove async and second id usage

EmosewaMC:
Optimized writes and reads to use aligned reads and writes to save time not looping.
Removed async code as the destructor of a std::future will block at the end of the function, causing it to block and removing its purpose, the code being run by it is async anyways, so it served no purpose anyways
Removed second persistentid generation that is unnecessary for ugc models as we overwrote it anyways with something new.

image
The pictured model well above a singular sd0 chunk size was able to be built, saved and downloaded when requested without issues.

### DIFF
--- a/dGame/dGameMessages/GameMessages.cpp
+++ b/dGame/dGameMessages/GameMessages.cpp
@@ -2523,15 +2523,9 @@ void GameMessages::HandleBBBSaveRequest(RakNet::BitStream* inStream, Entity* ent
 	inStream->Read(sd0Size);
 	std::shared_ptr<char[]> sd0Data(new char[sd0Size]);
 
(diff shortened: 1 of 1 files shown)


aronwk-aaron (approved):

DarwinAnim8or (approved):
Good changes-- this code was some of the last I wrote while working on this project originally and it was more about getting *anything* working so BBB could be considered "functional" before release. 

Wrapping it in async was I think just done on a whim because it seemed logical to make a big call like this async. Aside from that, ReadAlignedBytes and similar are indeed a proper way to do this, I just hadn't thought of it at the time; makes me wonder if there's other spaces where we could do it

ANYWAY LGTM TY EMO

PR #1272, opened 2023-11-10
feat: Vanity reload and vanity spawners

aronwk-aaron:
This PR properly reload vanity things by destroying them first and clearing out the list before re-reading the xml from disk.
Additionally, support for spawners via vanity is added

### DIFF
--- a/dGame/dUtilities/VanityUtilities.cpp
+++ b/dGame/dUtilities/VanityUtilities.cpp
@@ -35,9 +35,8 @@ void VanityUtilities::SpawnVanity() {
 		for (auto& npc : m_NPCs) {
 			if (npc.m_ID == LWOOBJID_EMPTY) continue;
 			if (npc.m_LOT == 176){
-				auto* spawner = Game::zoneManager->GetSpawner(npc.m_ID);
-				if (!spawner) continue;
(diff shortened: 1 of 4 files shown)


aronwk-aaron:
tested with a large vanity xml that everything reloaded properly.
The only thing that doesn't is the Testament plaque, which it didn't before.
Will do a PR later to move that to be defined in the vanity XML so it's not a special hardcoded case

dGame/dUtilities/VanityUtilities.cpp
```diff
@@ -28,6 +32,26 @@ void VanityUtilities::SpawnVanity() {
...
+		}
+		m_NPCs.clear();
+	}
+	
+	if (!m_Parties.empty()){
+		m_Parties.clear();
+	}
+	if (!m_PartyPhrases.empty()){
+		m_PartyPhrases.clear();
+	}
```
EmosewaMC, line 54: You can remove the if wrapper on all 3 of these, clear already has cases to handle this and the for loop will just do nothing if it is empty.
nit: `const auto& npc`

EmosewaMC (approved):

DarwinAnim8or (approved):
"SpawnSpawner" excellent. no notes

PR #1273, opened 2023-11-10
feat: use more zoneTable options

aronwk-aaron:
Allow setting framerate for the zone
Allow setting if pets are allowed in the zone
Allow setting if mounts are allowed in a zone
Allow disabling saving location to a zone

Tested all of these by editing cdclient and changing them then loading a fresh zone

### DIFF
--- a/dZoneManager/dZoneManager.h
+++ b/dZoneManager/dZoneManager.h
@@ -41,6 +41,9 @@ public:
 	void Update(float deltaTime);
 	Entity* GetZoneControlObject() { return m_ZoneControlObject; }
 	bool GetPlayerLoseCoinOnDeath() { return m_PlayerLoseCoinsOnDeath; }
+	bool GetDisableSaveLocation() { return m_DisableSaveLocation; }
+	bool GetMountsAllowed() { return m_MountsAllowed; }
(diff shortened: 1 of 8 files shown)


dZoneManager/dZoneManager.h
```diff
@@ -58,7 +61,10 @@ class dZoneManager {
 
 	Zone* m_pZone = nullptr;
 	LWOZONEID m_ZoneID;
-	bool m_PlayerLoseCoinsOnDeath; //Do players drop coins in this zone when smashed
+	bool m_PlayerLoseCoinsOnDeath;
+	bool m_DisableSaveLocation;
+	bool m_MountsAllowed;
+	bool m_PetsAllowed;
```
EmosewaMC, line 67: These should be default initializing to something, not sure if false or true is correct for default however.

EmosewaMC (approved):

DarwinAnim8or (approved):
Code itself seems good, only note is that we should remember this in case there's ever a bug where people aren't getting saved in the right location. I assume this flag is only ever set in instanced worlds, but y'know

aronwk-aaron:
Only every set in lup worlds. Instances are just server logic

PR #1274, opened 2023-11-10
refactor: Database abstraction and organization of files

EmosewaMC:
Abstract away the database queries in a mini-orm with the following motivations in mind
- Allow for plug and play of databases
- Unlock even more code to be tested in the codebase that wasn't testable before due to needing an actual MySQL database connection
- Structure database files in a client a game folder so organization is simple
- Implement much easier ways to prepare queries via Fold Expressions and Parameter Packing in tandem with argument forwarding and inlining to allow for the same end result as before, but with a much clearer and easier to use interface.

Other changes:
- Removed backup password checking
- Removed a good chunk of finalize calls to sqlite queries as the destructor of result sets from cdclient queries already calls finalize in a try catch(...)
- Remove duplicated code in Character::Character
- Remove if statement in User::operator==
- Simplify CreateCharacter logic for what name is used
- PropertyManagementComponent lookups no longer use TemplateID for the property, but rather the mapId/zoneId
- Master now shutsdown on inability to find object_id_tracker value to prevent any collisions
- Use WriteAlignedBytes in bbb serialization for speed
- Moved unreferenced ugc model deletion to world shutdown instead of when saving the property
- All moved queries no longer get fields by number but rather name to prevent issues with counting
- Fixed crash when parsing a brick count larger than the uint32_t max.

Notes:
- LeaderboardManager is unchanged on purpose.  Its change will easily be at least another 600 sloc.  Will clutter the pr.
- PropertyEntranceComponent is being redone in #1246 and as such is not changed here
  - When both of the above have their queries added to the GameDatabase Tables, the prepared statement methods should be made completely private or removed and no part of the dlu codebase, aside from the MySQL orm connector itself should know that MySQL exists.  conncpp.hpp will need to remain in its current header due to generic code, however that code should only get compiler if MySQL is switched on, which with another connector will not be done due to switches.

Tests were done incrementally for each query moved to ensure functionality and prevent hard to find breaking changes.  No observed regressions were found during incremental changes.  After the main pr was done, I did a lot of moving of files and structures around and, while these shouldn't have caused a regression, it is always possible that during the re-naming a new issue will pop up.


Xiphoseer (approved):
Bit too much to verify that the queries are identical, but LGTM

EmosewaMC:
> Bit too much to verify that the queries are identical, but LGTM

The big thing would be the actual queries themselves
- list the fields they get by name
- actually get all the fields by name

EmosewaMC:
This needs comments.  Not a great PR without them

EmosewaMC (commented):
I have read through each file and did not observe any obvious issues.

aronwk-aaron (approved):

aronwk-aaron (approved):

Jettford:
Will approve upon, Emo pushing the funnies.

aronwk-aaron (approved):

Jettford (approved):

Issue #916, opened 2022-12-23
BUG: Player PvP flag Checks are inconsistent 


EmosewaMC:
### Make sure you've done the following:

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### DarkflameServer Version

HEAD

### Issue Description

Player PvP needs to properly register other PvP players as enemies for PvP combat so damage can go through.  Currently it is reported that many Area of Effect and other behaviors do not properly check for if the player is an enemy and as such either do not function or deal no damage to a player.

### Reproduction Steps

1) Have 2 players with PvP on
2) Have Player 1 hit player 2 with The Big One
3) Observe that no damage is done
4) Have Player 1 equip the full Bone Suit
5) Observe if Player 2 hits Player 1 for 2 damage, Player 1 takes no damage.

### Expected Behavior

Damage is done

### Environment

All

Idefix-Asterix:
This also occurs when a player uses an ability that actively blocks damage
Known causes:
Space ranger faction set bonus
Potion of shielding
Shield of shielding

Any ability that "blocks n damage for t  seconds"

PR #1276, opened 2023-11-11
Adding damage cooldown/"invincibility frames" as in Live

### LINKED ISSUE #1157: BUG: global hurty timer cooldown thing is not implemented

### DarkflameServer Version

all except those from after this is eventually closed

### Issue Description

reminder 4 emo boi

EmmosewaMC — 06/24/2023 8:29 PM
https://youtu.be/HAaknVJU3Ew?t=3094 there absolutely is an immunity after taking damage, likely exactly what it is in WorldConfig

ADDITIONAL NOTE

some nodes specify that they should NOT apply the immunity

example: there are weapons in ninjago which, when held, apply a maelstrom poisoning effect which does 1 damage to the player every 3 seconds. these damage ticks should not trigger the immunity, as specified by dont_apply_immune in the basicattack node:

image

https://explorer.lu-dev.net/skills/behaviors/36887/36886

via https://explorer.lu-dev.net/skills/1497

used by objects such as https://explorer.lu-dev.net/objects/14472

### Reproduction Steps

drink bone hurting juice

### Expected Behavior

the poor soul expects a limit to the hurt. a finite amount of pain. but the server has no such mercy

### Environment

genetically modified cornputer

jadebenn:
Fixes #1157. Tested by using the /buffme command in Advent Gardens survival and enabling debug mode to view the various logs, as well as just seeing the gap between damage taken. Probably a very inelegant solution given my general inexperience with the codebase. Special thanks to [EmosewaMC](https://github.com/DarkflameUniverse/DarkflameServer/commits?author=EmosewaMC) for helping to make my ideas less bad.


EmosewaMC (changes requested):
Please remove the submodule added here

dGame/dBehaviors/BasicAttackBehavior.cpp
```diff
@@ -13,8 +15,15 @@ void BasicAttackBehavior::Handle(BehaviorContext* context, RakNet::BitStream* bi
 
 		auto* destroyableComponent = entity->GetComponent<DestroyableComponent>();
 		if (destroyableComponent != nullptr) {
-			PlayFx(u"onhit", entity->GetObjectID());
+			PlayFx(u"onhit", entity->GetObjectID()); //This damage animation doesn't seem to play consistently
 			destroyableComponent->Damage(this->m_MaxDamage, context->originator, context->skillID);
+
+			//Handle player damage cooldown
+			if (entity->IsPlayer() && !this->m_DontApplyImmune) {
+				const float ImmunityTime = Game::zoneManager->GetWorldConfig()->globalImmunityTime; //DETERMINE THE DAMAGE COOLDOWN TIME
```
EmosewaMC, line 23: local variables should follow camelCase

jadebenn (commented):
Got rid of the strange submodules.

dGame/dBehaviors/BasicAttackBehavior.cpp
```diff
@@ -72,6 +81,8 @@ void BasicAttackBehavior::DoHandleBehavior(BehaviorContext* context, RakNet::Bit
 	}
 
 	if (isImmune) {
+		LOG("Target targetEntity %llu is immune! (Is this method even used?)",branch.target); //Immune is succesfully proc'd
```
EmosewaMC, line 84: Should be demoted to a debug log

dGame/dBehaviors/BasicAttackBehavior.cpp
```diff
@@ -72,6 +81,8 @@ void BasicAttackBehavior::DoHandleBehavior(BehaviorContext* context, RakNet::Bit
...
+		//Seriously, is this ever used? I never see it in the console.
```
EmosewaMC, line 85: even if it isnt used, as per the reverse engineered client code, this is the correct logic if this path is hit

dGame/dBehaviors/BasicAttackBehavior.cpp
```diff
@@ -178,11 +189,15 @@ void BasicAttackBehavior::DoBehaviorCalculation(BehaviorContext* context, RakNet
 		return;
 	}
 
-	const bool isImmune = destroyableComponent->IsImmune();
+	const float ImmunityTime = Game::zoneManager->GetWorldConfig()->globalImmunityTime; //DETERMINE THE DAMAGE COOLDOWN TIME
+	LOG_DEBUG("Damage cooldown time is  %f s",destroyableComponent->GetDamageCooldownTimer()); //ALSO PRINT CURRENT TIMER VALUE
```
EmosewaMC, line 193: nit: double space here should be removed

dGame/dBehaviors/BasicAttackBehavior.cpp
```diff
@@ -178,11 +189,15 @@ void BasicAttackBehavior::DoBehaviorCalculation(BehaviorContext* context, RakNet
...
+
+	const bool isImmune = (destroyableComponent->IsImmune()) || (destroyableComponent->GetDamageCooldownTimer() > 0.0f);
```
EmosewaMC, line 195: `GetDamageCooldownTimer() > 0.0f`could be abstracted into a IsDamageImmune possibly?  not sure if a method with a similar name exists already

dGame/dBehaviors/BasicAttackBehavior.cpp
```diff
@@ -178,11 +189,15 @@ void BasicAttackBehavior::DoBehaviorCalculation(BehaviorContext* context, RakNet
...
 
 	bitStream->Write(isImmune);
 
 	if (isImmune) {
+		LOG_DEBUG("Target targetEntity %llu is immune!",branch.target); //Immune is succesfully proc'd
```
EmosewaMC, line 200: nit: space after comma

dGame/dBehaviors/BasicAttackBehavior.h
```diff
@@ -48,6 +48,8 @@ class BasicAttackBehavior final : public Behavior
 	 */
 	void Load() override;
 private:
+	bool m_DontApplyImmune; //Added boolean for this flag
```
EmosewaMC, line 51: can remove the comment here

dGame/dComponents/DestroyableComponent.cpp
```diff
@@ -73,6 +73,8 @@ DestroyableComponent::DestroyableComponent(Entity* parent) : Component(parent) {
 	m_ImmuneToQuickbuildInterruptCount = 0;
 	m_ImmuneToPullToPointCount = 0;
 	m_DeathBehavior = -1;
+
+	m_DamageCooldownTimer = 0.0f; //Added damage cooldown timer
```
EmosewaMC, line 77: can remove this comment here

dGame/dComponents/DestroyableComponent.cpp
```diff
@@ -179,6 +181,10 @@ void DestroyableComponent::Serialize(RakNet::BitStream* outBitStream, bool bIsIn
 	}
 }
 
+void DestroyableComponent::Update(float deltaTime) {
+	m_DamageCooldownTimer -= deltaTime; //Decrement the damage cooldown timer
```
EmosewaMC, line 185: can remove this comment here

dGame/dComponents/DestroyableComponent.cpp
```diff
@@ -546,7 +552,8 @@ void DestroyableComponent::Damage(uint32_t damage, const LWOOBJID source, uint32
 		return;
 	}
 
-	if (IsImmune()) {
+	if (IsImmune() || m_DamageCooldownTimer > 0.0f) { 
+		LOG("Target targetEntity %llu is immune!",m_Parent->GetObjectID()); //Immune is succesfully proc'd
```
EmosewaMC, line 556: This should be demoted to a debug log

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -1490,6 +1490,24 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 		return;
 	}
 
+	//TESTING IMMUNITY
+	if (chatCommand == "immune" && args.size() >= 1 && entity->GetGMLevel() >= eGameMasterLevel::DEVELOPER) {
```
EmosewaMC, line 1494: If this is going to be merged as well, the name of the command should be a bit more specific as to what you are immune to (in this case basic attack immunity) and should be documented in the Docs, or should be removed, up to you which path you take

EmosewaMC (changes requested):
General feedback and formatting.  Comments are only really needed for APIs and unclear code, and if a comment just describes what its currently doing, it can be removed.  Caps lock is used quite a lot, would recommend using standard case except for logs that really need to stand out i.e. WARNING, ERROR etc.

Logic looks good though.  Could simplify the getter to just return if they are damage immune though to make the code easier to read.
I would also recommend running a formatter across the edited files to make sure they are following the format standards we currently have.

As an extra mile, but not required, you could add a test to the DestroyableComponentTests that makes sure the variable is decremented properly (see DestroyableComponentTests.cpp for the test suite we currently have for this component)

dGame/dBehaviors/BasicAttackBehavior.cpp
jadebenn: Not quite sure what you mean here.

EmosewaMC: i.e. making a public method in DestroyableComponent which is called IsDamageImmune() which does this comparison for you

jadebenn: Done!

jadebenn:
Updated in response to feedback

dGame/dBehaviors/BasicAttackBehavior.cpp
```diff
@@ -189,15 +188,15 @@ void BasicAttackBehavior::DoBehaviorCalculation(BehaviorContext* context, RakNet
 		return;
 	}
 
-	const float ImmunityTime = Game::zoneManager->GetWorldConfig()->globalImmunityTime; //DETERMINE THE DAMAGE COOLDOWN TIME
-	LOG_DEBUG("Damage cooldown time is  %f s",destroyableComponent->GetDamageCooldownTimer()); //ALSO PRINT CURRENT TIMER VALUE
+	const float immunityTime = Game::zoneManager->GetWorldConfig()->globalImmunityTime; //Determine the damage cooldown time
+	LOG_DEBUG("Damage cooldown time is %f s", destroyableComponent->GetDamageCooldownTimer()); //Print current damage cooldown time value
```
EmosewaMC, line 192: Thinking on it now, this log is going to be pretty noisy since we use it in all 3 paths and its a value that doesn't change.  Do you think we should remove this or keep it in?

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -1490,8 +1490,8 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 		return;
 	}
 
-	//TESTING IMMUNITY
-	if (chatCommand == "immune" && args.size() >= 1 && entity->GetGMLevel() >= eGameMasterLevel::DEVELOPER) {
+	//Testing basic attack immunity
+	if (chatCommand == "attackimmune" && args.size() >= 1 && entity->GetGMLevel() >= eGameMasterLevel::DEVELOPER) {
```
EmosewaMC, line 1494: since we'll be keeping this in, don't forget to add information about the command to the [docs here](https://github.com/DarkflameUniverse/DarkflameServer/blob/main/docs/Commands.md)

dGame/dComponents/DestroyableComponent.h
```diff
@@ -167,6 +167,12 @@ class DestroyableComponent : public Component {
 	 */
 	bool IsImmune() const;
 
+	/**
+	 * Returns whether this entity is currently immune to attacks due to a damage cooldown period
```
EmosewaMC, line 171: Can get rid of this first line since its a duplicate comment

dGame/dBehaviors/BasicAttackBehavior.cpp
```diff
@@ -20,9 +20,9 @@ void BasicAttackBehavior::Handle(BehaviorContext* context, RakNet::BitStream* bi
 
 			//Handle player damage cooldown
 			if (entity->IsPlayer() && !this->m_DontApplyImmune) {
-				const float ImmunityTime = Game::zoneManager->GetWorldConfig()->globalImmunityTime; //DETERMINE THE DAMAGE COOLDOWN TIME
-				destroyableComponent->SetDamageCooldownTimer(ImmunityTime); //set the cooldowntimer
-				LOG_DEBUG("TOOK BUFF DAMAGE: SETTING DAMAGE COOLDOWN TIMER TO %f s",ImmunityTime);
+				const float immunityTime = Game::zoneManager->GetWorldConfig()->globalImmunityTime; //Determine the damage cooldown time
+				destroyableComponent->SetDamageCooldownTimer(immunityTime); //set the cooldowntimer
```
EmosewaMC, line 24: Can get rid of the comments on lines similar to this one

dGame/dBehaviors/BasicAttackBehavior.cpp
```diff
@@ -81,8 +81,7 @@ void BasicAttackBehavior::DoHandleBehavior(BehaviorContext* context, RakNet::Bit
 	}
 
 	if (isImmune) {
-		LOG("Target targetEntity %llu is immune! (Is this method even used?)",branch.target); //Immune is succesfully proc'd
-		//Seriously, is this ever used? I never see it in the console.
+		LOG_DEBUG("Target targetEntity %llu is immune!", branch.target); //Immune is succesfully proc'd
```
EmosewaMC, line 84: The log and comment are the same, so only one is necessary, I'd keep the log but ill leave that up to you to decide

EmosewaMC (commented):
Just a reminder to add a doc comment and consider the rest of the feedback, or acknowledge the feedback as nominal

dGame/dBehaviors/BasicAttackBehavior.cpp
aronwk-aaron: imo, remove it

jadebenn: Just to clarify, this is actually printing the value of the timer whenever damage is taken, not the length of the immunity period (which, as you say, doesn't change). I've found it useful in debugging if the timer is functioning properly and the value is properly being set (or properly being NOT set, in the case of m_DontApplyImmune. Do you still want me to get rid of it?

jadebenn:
As far as the previously discussed changes go, only waiting on confirmation that I should delete the timer value readout given the additional context I provided. I'm fine with ditching it if it still seems like too much clutter.

jadebenn:
Also, now that I think about it, would this interfere with PVP? I could try adding another flag to not trigger the damage cooldown if we don't want it to apply to attacks made by another player.

EmosewaMC:
> Also, now that I think about it, would this interfere with PVP? I could try adding another flag to not trigger the damage cooldown if we don't want it to apply to attacks made by another player.

Yes that would be a good check to add

dGame/dBehaviors/BasicAttackBehavior.cpp
EmosewaMC: Ah in that case this makes sense to keep for debug logging

dGame/dComponents/DestroyableComponent.h
```diff
@@ -24,6 +24,7 @@ class DestroyableComponent : public Component {
 	DestroyableComponent(Entity* parentEntity);
 	~DestroyableComponent() override;
 
+	void Update(float deltaTime) override; //Add update function to DestroyableComponent
```
EmosewaMC, line 27: This comment can Be removed

tests/dGameTests/dComponentsTests/DestroyableComponentTests.cpp
```diff
@@ -536,3 +536,17 @@ TEST_F(DestroyableTest, DestroyableComponentImmunityTest) {
...
+/**
+ * Test the Damage cooldown timer of DestroyableComponent
+ */
+TEST_F(DestroyableTest, DestroyableComponentDamageCooldownTest) {
+	// Test the damage immune timer state (anything above 0.0f)
+	destroyableComponent->SetDamageCooldownTimer(1.0f);
+	EXPECT_FLOAT_EQ(destroyableComponent->GetDamageCooldownTimer(), 1.0f);
+	ASSERT_TRUE(destroyableComponent->IsCooldownImmune());
+
+	// Test the non damage immune timer state (anything below or equal to 0.0f)
```
EmosewaMC, line 548: Could also test that calling update on the component properly lowers the cooldown timer too

jadebenn:
> > Also, now that I think about it, would this interfere with PVP? I could try adding another flag to not trigger the damage cooldown if we don't want it to apply to attacks made by another player.
> 
> Yes that would be a good check to add

I loaded in two characters, buffed them both, and had one do the long-fire on the Wormholer. Whatever method is being used for PVP damage already seems to bypass the cooldown, so I think this isn't an issue after all.

EmosewaMC:
I would test with a few varying weapons; pvp is not quite consistent as per #916

jadebenn:
> I would test with a few varying weapons; pvp is not quite consistent as per #916

Of the weapons I've tested:
- Fists (Unarmed)
- The Wormholer
- The Doomslicer
- The Serratorizer
- Basic Hammer
- Buccaneer Pistol and Cutlass 3
All work as expected and do not trigger any debug logs related to the damage cooldown period, nor do I notice any changes in health/armor drain (no sign of the start-and-stop you'd expect from the immunity period). 

I was able to reproduce the issues you noted in #916, but obviously those are pre-existing and not related to any of the changes in this PR. I did notice - and I'm not sure this is logged - but the Basic Flintlock pistol projectiles just go through the corresponding player without colliding or dealing damage. I tried with the Superior Crossbow of Blasting as well, and I saw the same behavior, so it seems to be a bug common to all projectile weapons.

In summary, the working PVP weapons don't seem to be triggering the cooldown period, which is the desired behavior.

EmosewaMC (approved):
lgtm, thanks for working on this!

aronwk-aaron (approved):

PR #1278, opened 2023-11-12
Ugc: Use persistent Id for rockets and cars

EmosewaMC:
After further research, you do end up needing a persistentID for this since if two users were to have the same subkey for their rocket, the actual ugc server does not know who is requesting the model, just that an id was requested.

Tested that rockets properly get a subkey and make http requests to download their image if ugc is on


EmosewaMC:
superceded by #1279

PR #1279, opened 2023-11-12
fix: ugc Save rocket and car modular assembly data to database

EmosewaMC:
Not sure of a cleaner way to do this besides crawling the full table of user data and searching for subkey x and then further parsing for the modular assembly parts

Tested that the data is inserted into the database in the format of 1:part1+1:part2+1:part3 or for cars up to 7.

#1278 needs to be merged first in either case.


migrations/dlu/12_modular_build_ugc.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE IF NOT EXISTS ugc_modular_build (
+    id BIGINT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+	ugc_id BIGINT NOT NULL,
+    rocket_ldf VARCHAR(60) NOT NULL
```
aronwk-aaron, line 4: Rocket ldf? Isn't that supposed to work for cars too. Maybe some better generic name

migrations/dlu/12_modular_build_ugc.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE IF NOT EXISTS ugc_modular_build (
+    id BIGINT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+	ugc_id BIGINT NOT NULL,
```
aronwk-aaron, line 3: Why is Id and ugc_id needed? Ugc_id should be unique and can serve as a primary key

aronwk-aaron (changes requested):

migrations/dlu/12_modular_build_ugc.sql
EmosewaMC: I think you can figure out why I named it rocket ldf.

migrations/dlu/12_modular_build_ugc.sql
```diff
@@ -1,5 +1,4 @@
 CREATE TABLE IF NOT EXISTS ugc_modular_build (
```
aronwk-aaron, line 1: having the character_id as a FK on this table would help remove data when characters are deleted.
sooo something like...

```sql
CREATE TABLE IF NOT EXISTS ugc_modular_build (
	ugc_id BIGINT NOT NULL PRIMARY KEY,
	character_id BIGINT NOT NULL,
	ldf_config VARCHAR(60) NOT NULL,
	FOREIGN_KEY (character_id) REFERENCES charinfo(id) ON DELETE CASCADE,
);
```

Additionally, when entering modular build mode with a vehicle, is this data entry cleaned up? or is it reused somehow?

aronwk-aaron (approved):

Jettford (approved):

PR #1292, opened 2023-11-14
fix: Models leak

EmosewaMC:
Fix a massive memory leak by deleting the doc after we are done using it.  Tested that the memory does not leak in the megabytes everytime you dismantle a model to bricks.


aronwk-aaron (approved):
lol

DarwinAnim8or (approved):
can you imagine someone just forgetting to remove their raw pointer? smh :^)

PR #1294, opened 2023-11-14
feat: Darkflame Cinema

Wincent01:
> Darkflame Cinema is a way to create interactive — acted — experiences in Darkflame Universe. It aims to provide a complement to the ordinary mission structure with static NPCs and mission texts to convey story. Insperation is drawn from games likes of World of Warcraft, where NPCs frequently act out scenes, alone and in combination with both the player and other NPCs. 

See the [README](https://github.com/DarkflameUniverse/DarkflameServer/blob/dCinema/dGame/dCinema/README.md) for documentation.

The tool can be used without modded objects, but will have limited utility. With precedence of tools like VanityNPC this can fit in the main repository — introducing only optional functionality.


dGame/Player.cpp
```diff
@@ -32,9 +32,6 @@ Player::Player(const LWOOBJID& objectID, const EntityInfo info, User* user, Enti
 	m_GhostReferencePoint = NiPoint3::ZERO;
 	m_GhostOverridePoint = NiPoint3::ZERO;
 	m_GhostOverride = false;
-	m_ObservedEntitiesLength = 256;
-	m_ObservedEntitiesUsed = 0;
-	m_ObservedEntities.resize(m_ObservedEntitiesLength);
```
Wincent01, line 37: This had no right being a vector instead of a set

dGame/dCinema/Prefab.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+using namespace Cinema;
+
+std::unordered_map<std::string, Prefab> Prefab::m_Prefabs;
+std::unordered_map<size_t, Prefab::Instance> Prefab::m_Instances;
+
+size_t Prefab::AddObject(LOT lot, NiPoint3 position, NiQuaternion rotation, float scale) {
+	const auto id = ObjectIDManager::GenerateRandomObjectID();
+
+	m_Pieces.emplace(id, Prefab::Piece { lot, position, rotation, scale, {} });
```
EmosewaMC, line 18: Personally not a fan of the {} without a struct name because you dont know what it is here out of context, Id recommend adding the class name before that so its more readable

dGame/dCinema/Prefab.cpp
```diff
@@ -0,0 +1,172 @@
...
+		const auto lot = element->UnsignedAttribute("lot");
+		const auto position = NiPoint3(element->FloatAttribute("x"), element->FloatAttribute("y"), element->FloatAttribute("z"));
+
+		NiQuaternion rotation;
+		// Check if the qx attribute exists, if so the rotation is a quaternion, otherwise it's a vector
+		if (!element->Attribute("qx")) {
+			rotation = NiQuaternion::FromEulerAngles( { 
+				element->FloatAttribute("rx") * 0.0174532925f,
+				element->FloatAttribute("ry") * 0.0174532925f,
+				element->FloatAttribute("rz") * 0.0174532925f
```
EmosewaMC, line 60: These constants should be named something

dGame/dCinema/Prefab.cpp
```diff
@@ -0,0 +1,172 @@
...
+			auto* renderComponent = entity->GetComponent<RenderComponent>();
+
+			if (!renderComponent) {
+				continue;
+			}
+
+			// Generate random name
+			std::string effectName = "Effect_";
+			for (int i = 0; i < 10; ++i) {
+				effectName += std::to_string(rand() % 10);
```
EmosewaMC, line 144: I would recommend GeneralUtils::GenerateRandomNumber<int32_t>(0, 10) here.

dGame/dCinema/Prefab.h
```diff
@@ -0,0 +1,114 @@
...
+
+	/**
+	 * @brief Destroys the instance with the given ID.
+	 * 
+	 * @param instanceID The ID of the instance to destroy.
+	 */
+	static void DestroyInstance(size_t instanceID);
+	
+
+	~Prefab() = default;
```
EmosewaMC, line 90: This should be moved right next to the destructor imo

dGame/dCinema/Recorder.cpp
```diff
@@ -0,0 +1,1195 @@
...
+#include "EntityInfo.h"
+#include "ServerPreconditions.hpp"
+#include "MovementAIComponent.h"
+#include "BaseCombatAIComponent.h"
+
+using namespace Cinema::Recording;
+
+std::unordered_map<LWOOBJID, Recorder*> m_Recorders = {};
+
+std::unordered_map<int32_t, std::string> m_EffectAnimations = {};
```
EmosewaMC, line 18: These should be in an anonymous namespace as well, especially for globals so if you type m_Recorders, you are guaranteed to be referring to this m_Recorders here and not one elsewhere

dGame/dCinema/Recorder.cpp
```diff
@@ -0,0 +1,1195 @@
...
+
+Recorder::Recorder() {
+	this->m_StartTime = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch());
+	this->m_IsRecording = false;
+	this->m_LastRecordTime = this->m_StartTime;
+}
+
+Recorder::~Recorder() {
```
EmosewaMC, line 26: Can replace in the header with ~Recorder() = default;

dGame/dCinema/Recorder.cpp
```diff
@@ -0,0 +1,1195 @@
...
+
+void Recorder::ActingDispatch(Entity* actor, const std::vector<Record*>& records, size_t index, Play* variables) {
+	if (index >= records.size()) {
+		return;
+	}
+
+	auto* record = records[index];
+
+	// Check if the record is a fork
+	auto* forkRecord = dynamic_cast<ForkRecord*>(record);
```
EmosewaMC, line 62: I think I see what is going on here; I would recommend moving these into handler functions called Dispatch for each of the record types and have them overload what needs to be done based on that instead of a bunch of dynamic casts.  Would make a much smaller function and split functionality up for better maintainability

dGame/dCinema/Recorder.cpp
```diff
@@ -0,0 +1,1195 @@
...
+		return;
+	}
+	
+	Game::entityManager->GetZoneControlEntity()->AddCallbackTimer(1.0f, [actor, records, index, variables, actionTaken]() {
+		PlayerProximityDispatch(actor, records, index, variables, actionTaken);
+	});
+}
+
+void Cinema::Recording::Recorder::PathFindDispatch(Entity* actor, const std::vector<Record*>& records, size_t index, Play* variables) {
+	auto* record = dynamic_cast<PathFindRecord*>(records[index]);
```
EmosewaMC, line 334: Same thing here, this should be abstracted into overloaded methods so you can just call those on the object instead of needing to do this cast and then check the ptr

dGame/dCinema/Recorder.cpp
```diff
@@ -0,0 +1,1195 @@
...
+			result.finalize();
+
+			m_EffectAnimations.emplace(effectId, "");
+		}
+		else {
+			const auto animationName = result.getStringField(0);
+
+			m_EffectAnimations.emplace(effectId, animationName);
+
+			result.finalize();
```
EmosewaMC, line 525: finalize isnt necessary to manually call, the destructor for resultset here already calls finalize in a try {finalize()} catch (...) {};

dGame/dCinema/Recorder.cpp
```diff
@@ -0,0 +1,1195 @@
...
+
+	angularVelocity.x = element->FloatAttribute("avx");
+	angularVelocity.y = element->FloatAttribute("avy");
+	angularVelocity.z = element->FloatAttribute("avz");
+
+	onGround = element->BoolAttribute("g");
+
+	dirtyVelocity = element->BoolAttribute("dv");
+
+	dirtyAngularVelocity = element->BoolAttribute("dav");
```
EmosewaMC, line 616: Why are we saving dirty variables here?

dGame/dCinema/Recorder.cpp
```diff
@@ -0,0 +1,1195 @@
...
+
+	element->SetAttribute("text", text.c_str());
+
+	element->SetAttribute("t", m_Delay);
+
+	parent->InsertEndChild(element);
+}
+
+void SpeakRecord::Deserialize(tinyxml2::XMLElement* element) {
+	text = element->Attribute("text");
```
EmosewaMC, line 643: should check for nullptr and return early if it is

dGame/dCinema/Recorder.cpp
```diff
@@ -0,0 +1,1195 @@
...
+
+	m_Delay = element->DoubleAttribute("t");
+}
+
+AnimationRecord::AnimationRecord(const std::string& animation) {
```
EmosewaMC, line 648: May be worth splitting this into smaller files, this file is doing quite a lot of work right now with how large it is

dGame/dCinema/Recorder.cpp
```diff
@@ -0,0 +1,1195 @@
...
+}
+
+EquipRecord::EquipRecord(LOT item) {
+	this->item = item;
+}
+
+void EquipRecord::Act(Entity* actor) {
+	auto* inventoryComponent = actor->GetComponent<InventoryComponent>();
+
+	const LWOOBJID id = ObjectIDManager::Instance()->GenerateObjectID();
```
EmosewaMC, line 681: Should this be GenerateObjectID, or GenerateRandomID?

dGame/dCinema/Recorder.cpp
```diff
@@ -0,0 +1,1195 @@
...
+
+float Recorder::GetDuration() const {
+	// Return the sum of all the record delays
+	float duration = 0.0f;
+
+	for (auto* record : m_Records) {
+		duration += record->m_Delay;
+	}
+
+	return duration;
```
EmosewaMC, line 788: I would recommend std::accumulate here, not required tho

dGame/dCinema/Recorder.cpp
```diff
@@ -0,0 +1,1195 @@
...
+
+	element->SetAttribute("t", m_Delay);
+
+	parent->InsertEndChild(element);
+}
+
+void Cinema::Recording::CombatAIRecord::Deserialize(tinyxml2::XMLElement* element) {
+	enabled = element->BoolAttribute("enabled");
+
+	m_Delay = element->DoubleAttribute("t");
```
EmosewaMC, line 1194: May be worth considering not using super short variable names so the xml is more readable and changable outside of code, unless its only intended to be used in code.

dGame/dCinema/Recorder.h
```diff
@@ -0,0 +1,401 @@
...
+	static void RegisterEffectForActor(LWOOBJID actorID, const int32_t& effectId);
+
+	static void LoadRecords(tinyxml2::XMLElement* root, std::vector<Record*>& records);
+
+	static void ActingDispatch(Entity* actor, const std::vector<Record*>& records, size_t index, Play* variables);
+
+private:
+	static void PlayerProximityDispatch(Entity* actor, const std::vector<Record*>& records, size_t index, Play* variables, std::shared_ptr<bool> actionTaken);
+
+	static void PathFindDispatch(Entity* actor, const std::vector<Record*>& records, size_t index, Play* variables);
```
EmosewaMC, line 68: These could be moved to the translation unit for this file since they are static private members of a class, they dont really need to be in the header for it, what do you think?

dGame/dCinema/Recorder.h
```diff
@@ -0,0 +1,401 @@
...
+class MovementRecord : public Record
+{
+public:
+	NiPoint3 position;
+	NiQuaternion rotation;
+	NiPoint3 velocity;
+	NiPoint3 angularVelocity;
+	bool onGround;
+	bool dirtyVelocity;
+	bool dirtyAngularVelocity;
```
EmosewaMC, line 88: member variables should be moved to private if possible

dGame/dCinema/Scene.h
```diff
@@ -0,0 +1,120 @@
...
+	static void AutoLoadScenesForZone(LWOMAPID zone);
+
+private:
+	void CheckForShowings();
+
+	void CheckTicket(Entity* player);
+
+	std::vector<std::pair<LOT, std::pair<NiPoint3, NiQuaternion>>> m_Objects;
+	std::vector<std::pair<Prefab, NiPoint3>> m_Prefabs;
+	std::vector<std::pair<LOT, std::pair<Recording::Recorder*, std::string>>> m_NPCs;
```
EmosewaMC, line 99: The first and third objects here can be replaced with std::multimap if I am understanding the use case here correctly

dGame/dCinema/Scene.h
```diff
@@ -0,0 +1,120 @@
...
+
+	NiPoint3 m_Center;
+	float m_Bounds = 0.0f;
+	float m_ShowingDistance = 0.0f;
+	float m_ChanceToPlay = 1.0f;
+	bool m_Repeatable = true;
+
+	std::vector<std::pair<PreconditionExpression, bool>> m_Preconditions;
```
EmosewaMC, line 107: Can be replaced with a std::set (std::set is just std::map<T, bool> effectively)

dGame/dUtilities/ServerPreconditions.hpp
```diff
@@ -0,0 +1,109 @@
...
+ * 	      is shown to the client or not.
+ */
+namespace ServerPreconditions
+{
+
+extern std::unordered_map<LOT, std::vector<std::pair<bool, PreconditionExpression>>> m_Preconditions;
+
+extern std::unordered_map<LWOOBJID, LWOOBJID> m_SoloActors;
+
+extern std::unordered_map<LWOOBJID, std::unordered_set<LWOOBJID>> m_ExcludeForPlayer;
```
EmosewaMC, line 22: Why extern?

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -883,6 +915,150 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
...
+
+		// Construct it
+		Game::entityManager->ConstructEntity(actor);
+
+		auto* record = Cinema::Recording::Recorder::GetRecorder(entity->GetObjectID());
+
+		if (record) {
+			record->Act(actor);
+		} else {
+			Game::logger->Log("SlashCommandHandler", "Failed to get recorder for objectID: %llu", entity->GetObjectID());
```
EmosewaMC, line 956: LOG macro should be used here, applies to all other non macro usages so ill just leave thisone

EmosewaMC (commented):
First pass of review feedback, will do a second pass later

dCommon/dConfig.cpp
```diff
@@ -34,7 +34,13 @@ void dConfig::ReloadConfig() {
...
 
 const std::string& dConfig::GetValue(std::string key) {
-	return this->m_ConfigValues[key];
+	static std::string emptyString{};
+	
+	const auto& it = this->m_ConfigValues.find(key);
+
+	if (it == this->m_ConfigValues.end()) return emptyString;
+
+	return it->second;
```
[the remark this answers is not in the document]
Wincent01: When we were troubleshooting the SQL stuff this was where a crash was occuring.

dGame/dCinema/Play.cpp
```diff
@@ -0,0 +1,89 @@
...
+#include "Scene.h"
+
+#include "EntityManager.h"
+
+using namespace Cinema;
+
+void Cinema::Play::Conclude() {
+	auto* player = Game::entityManager->GetEntity(this->player);
+
+	if (player == nullptr) {
```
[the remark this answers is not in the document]
Wincent01: I've alwayed prefered the clearify of comparing with nullptr, but will consider changing it.

dGame/dCinema/Scene.h
Wincent01: Don't think so? The boolean here specifies weather or not the precondition should be negated.

dGame/dCinema/Scene.h
Wincent01: The LOT isn't a key here. This is a list of LOTs which should be spawned at a specified location/rotation.

dGame/dCinema/Recorder.cpp
Wincent01: In my experience, handling a positional packet and not passing on the dirty flags can have some odd effects. I see that's still how we handle it. Don't think the redundency hurts.

dGame/dCinema/Scene.h
EmosewaMC: In that case it could just br replaced with a map instead for lookup

dGame/dCinema/Recorder.cpp
EmosewaMC: Well, setting the values as they are above already sets these flags to dirty in the contexts where they exist, hence seeming pointless to bother saving this because you should always mark hard changes to velocity as dirty

dGame/dCinema/Recorder.cpp
```diff
@@ -0,0 +1,1200 @@
...
+	element->SetAttribute("cleanUp", true);
+
+	parent->InsertEndChild(element);
+}
+
+void Cinema::Recording::ConcludeRecord::Deserialize(tinyxml2::XMLElement* element) {
+	m_Delay = element->DoubleAttribute("t");
+
+	if (element->Attribute("cleanUp")) {
+		cleanUp = element->BoolAttribute("clean-up");
```
EmosewaMC, line 1060:
```suggestion
		cleanUp = element->BoolAttribute("cleanUp");
```

EmosewaMC: also as a note, bool attribute already does bounds checks for you, the second argument allows you to choose a default value should one not exist in the xml doc

PR #1295, opened 2023-11-15
fix: Dismantling of models taking many many more resources than necessary

EmosewaMC:
allows the user to only send a maximum of numUniqueBricksInModel gms to their client when dismantling a model instead of numBricksInModel * numModelsToDismantle gms.  The difference here is literally sending 30ish gms at the most vs 1.5 million.

Tested that the client doesn't crash due to dismantling so many models


aronwk-aaron (approved):
packets go brrrrr

Jettford (approved):

MustacheMan2:
_I_ am finding that this fix has made crashes _more_ frequent

EmosewaMC:
Are you just going to say its more frequent or are you going to say what is causing it as well?

EmosewaMC:
if you are talking about getting kicked to login, please ensure UGCUSE3DSERVICES is set to 7:0 and that your game is in a folder called client.

EmosewaMC:
I have found an item with the issue.  I would appreciate just being given the item that has an issue rather than a vague hint that there are issues still as it doesnt happen with most models.

MustacheMan2:
Sorry about my previous comment, I knew at the time that my comment wasn't especially helpful and I still posted it for some reason, I just couldn't be bothered to provide evidence. The issue I'm having is also a server crash, not a client crash, so I'm just going to quietly walk away lol

EmosewaMC:
regardless its fixed now, please confirm the fix on your end if possible

PR #1296, opened 2023-11-15
fix: Buff FX not playing and general BuffComponent improvements

### LINKED ISSUE #1284: BUG: Self-damaging "buffs" seem to apply the damage animation only once

### DarkflameServer Version

411dce7

### Issue Description

I previously noted this as a code comment in pull request #1276, but I'm making this into an actual issue for easier tracking. While taking damage in general can sometimes be inconsistent as to whether the animation is applied (I may make a separate issue for that), it's much easier to test this particular damage animation issue - even if I still don't understand what's causing it.

https://github.com/DarkflameUniverse/DarkflameServer/assets/9892985/f658c27a-22ee-408b-a615-051d3f2ae339

### Reproduction Steps

1. Pull out a Maelstrom Dagger
2. Hold it
3. Observe that health decrements without the corresponding damage animation after the first time

### Expected Behavior

The damage animation should play when the player character takes damage

### Environment

Windows client, WSL Ubuntu server

EmosewaMC:
Fixes an issue where buff effects would not play at all.

Tested that frakjaw and maelstrom dagger now play their respective effects when you are effected by them

fixes #1284 

fix serialization to use the intended members and values, also fix the serialization of AddBuff to actually be correct.


EmosewaMC:
need to double check saving is still a-ok,

aronwk-aaron (approved):

Jettford (approved):
LGTM.

PR #1312, opened 2023-11-20
feat: Add ability to toggle skipping of cinematics

EmosewaMC:
server owner can choose whether or not to allow this.

Tested that with the config option being off, no cinematics are skipped regardless of the flag value.
Tested that with the config option being on, cinematics will be skipped if the value is on and can be toggled on and off.


aronwk-aaron (approved):

Jettford (approved):
May the programming war crimes of Darwin the Anim8or, live on through EmosewaMC

PR #1316, opened 2023-11-23
fix: UB in remote input info

EmosewaMC:
Yes i should have made this first
no this wouldnt have happened with rust

I have NO clue why the new AddComponent caused this.  I dug into this for about 10 hours and it was specifically having the PossessableComponent go through the new AddComponent and nothing else.  Using the inline variant resulted in the same effect.  The RemoteInputInfo in the previous case was always initialized to 0 0 0 1, whereas with the new AddComponent for Possessable, it was UB.  

Tested that the client no longer logs HAVOK CRASH ON FRICTION multiple times per frame

### DIFF
--- a/dGame/dComponents/VehiclePhysicsComponent.h
+++ b/dGame/dComponents/VehiclePhysicsComponent.h
@@ -6,6 +6,13 @@
 #include "eReplicaComponentType.h"
 
 struct RemoteInputInfo {
+	RemoteInputInfo() {
+		m_RemoteInputX = 0;
+		m_RemoteInputY = 0;
+		m_IsPowersliding = false;
+		m_IsModified = false;
+	}
+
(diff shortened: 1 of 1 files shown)


Wincent01 (approved):
GCC/CLang doesn't initialize members without being told explicitly to do so, but MSVC does. So I don't think this issue would've happened had you compiled on Windows. Had a ton of these issues when we moved away from Windows way back when.

EmosewaMC:
> GCC/CLang doesn't initialize members without being told explicitly to do so, but MSVC does. So I don't think this issue would've happened had you compiled on Windows. Had a ton of these issues when we moved away from Windows way back when.

Yep windows i had the same experience.  Worked just fine, but more details on this behavior are in the dev discord.  Very """Fun""" bug.

aronwk-aaron (approved):

PR #1317, opened 2023-11-23
fix: Racing end of race leaderboards now work

EmosewaMC:
Tested that with two players, both players see the others time at the end of the race and all other metrics are shown correctly.

Technically the `outBitStream->Write(static_cast<uint16_t>(m_RacingPlayers.size()));` should only be written once but how we do it now it is written as we load players in and this is the cheap option compared to the number of bits we are supposed to waste at the end of races

Removed an unused function and linted the file.  Hide whitespace if you don't want to see those changes.


EmosewaMC:
has the change from #1316 in here so I could properly test and check client logs.

aronwk-aaron (approved):

Wincent01 (approved):

PR #1323, opened 2023-11-26
fix: respawn rotation

jadebenn:
Fixed off-by-one error in the PlayerReachedRespawnCheckpoint GM bitstream (thanks aron) and removed some unnecessary code that set player rotation and position (which was promptly overwritten by the client anyway).


aronwk-aaron (approved):

jadebenn:
> It's less that it was off by one and more that the wrong component was trying to be fetched in the first place, so that write never happened anyways

It was a few errors bundled into one: an off by one error where a boolean wasn't being written, a component being fetched that shouldn't have been fetched, and the function entirely ignoring the rotation value passed to it in favor of trying to pull the rotation value from said incorrect component call.

I'm guessing this was some really, really old code.

PR #1325, opened 2023-11-27
fix: visiting multiple owned by same person

EmosewaMC:
will test later, should allow you to get credit for visiting player A's block yard and player A's other properties instead of just 1 of player A's properties.


EmosewaMC:
Tested that changes work

aronwk-aaron (approved):

Jettford (approved):

PR #1326, opened 2023-11-28
fix: tacarc not using offset or checking upper/lower bounds

### LINKED ISSUE #956: BUG: Skeleton Engineers attack incorrectly

### DarkflameServer Version

e67f310

### Issue Description

This has been the case forever. Currently skeleton engineers in the ninjago section of crux prime show a spinning particle effect around them when you approach them, and then you take damage and are flung a very far distance away. Then upon approaching them again they use their gun on you, and a short time later the cycle repeats.

### Reproduction Steps

come into close range of a skeleton engineer

### Expected Behavior

Not entirely sure how they are supposed to attack, but I'm finding now that this is not the intended behavior. Live gameplay shows them just using their gun and at a higher frequency than in DLU (https://www.youtube.com/watch?v=9xLAukyfwzU&t=416s&ab_channel=jamessterV2 around the 6 minute mark). The wiki explains that when they attempt to use spinjitzu they will knock you very far back if you are in close enough range, which would explain why theres currently a spinning animation around them when they knock the player back, but I'm unsure on how accurate this is.

### Environment

ubuntu server

aronwk-aaron:
Tested in crux that enemies attack normally while on the ground, but skeleton engineers use the knockback attack when up on the tower to knock you off.

tested with other enemies around crux to make sure they felt accurate as well.

Also did some other cleanup to make it more readable

Fixes #956

### DIFF
--- a/dGame/dBehaviors/TacArcBehavior.cpp
+++ b/dGame/dBehaviors/TacArcBehavior.cpp
@@ -104,7 +104,7 @@ void TacArcBehavior::Calculate(BehaviorContext* context, RakNet::BitStream* bitS
 
 	const auto casterPosition = self->GetPosition();
 
-	auto reference = self->GetPosition(); //+ m_offset;
+	auto reference = self->GetPosition() + m_offset;
 
 	targets.clear();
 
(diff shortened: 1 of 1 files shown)


EmosewaMC (approved):
very cool

DarwinAnim8or (approved):
:^)

PR #1327, opened 2023-11-29
fix: backwards name check

EmosewaMC:
Untested.  Should fix not being able to rename a character at character create


aronwk-aaron (approved):

EmosewaMC:
Tested that this works

Wincent01 (approved):

PR #1330, opened 2023-12-01
Fix: Log Output WorldServer "Server is listening on "IP""

IraqLobsterDE:
Fixes the Log Output for Server is listening on "IP" from many 0`s to the IP set in sharedconfig.ini

before:

Screenshot 2023-12-01 081807

after:

Screenshot 2023-12-01 082114


EmosewaMC:
This is not the correct fix.  The issue is skirting around the fact that the database has failed to actually return a good value.  This change just moves the problem.  The bug is with mariadb so the fix is to downgrade to a version when the bad return value doesnt happen.

EmosewaMC:
If you'd like to possibly actually fix the issue, you could try upgrading the submodule in the repository and seeing if the issue resolves itself then, but this PR will not be merged as it goes around the actual issue here, which is a faulty database connection that wont work anyways since all queries would return 0's.

PR #1336, opened 2023-12-14
fix: general crashes

### LINKED ISSUE #1244: BUG: Windows multi threaded signal handling

### DarkflameServer Version

all

### Issue Description

Windows handles signals via a separate thread and because of this this implicitly makes dlu run multithreaded when a signal is caught on windows.  Dlu is not built with multithreading in mind and this will cause issues.

### Reproduction Steps

put a log in the master server main loop
put a sleep in the signal handler
Start master server
stop master server with ctrl+c
observe that the main loop is still running while the signal is being handled.  This applies to all servers that handle a signal.

### Expected Behavior

No dlu entity code is executed across threads that is not thread safe

### Environment

Windows

### LINKED ISSUE #1335: BUG: Sending a private chat message to player sends both players to login

### DarkflameServer Version

Latest

### Issue Description

Sending a chat message to player sends both players to login, the sender and receiver: 
In database under player_cheat_detections:
Player hey attempted to bypass chat filter with message: hi
Sending GM with a sending player that does not match their own. GM ID: 1498

### Reproduction Steps

Add another player as friends and send a private chat message.

### Expected Behavior

Log out both players.

### Environment

Ubuntu

EmosewaMC:
fix crash with chat filter
fix ldf_config being empty in database on windows debug
fix windows crashes on shutdown
fix zonemanager crash on shutdown

Tested that the ldf_config column has data in it in the database now on windows debug
tested that you cannot crash the server as per #1335 
tested that no more crashes happen on world shutdown with players in the world on windows

fixes #1335 
fixes #1244

### DIFF
--- a/dWorldServer/WorldServer.cpp
+++ b/dWorldServer/WorldServer.cpp
@@ -18,6 +18,7 @@
 #include "PerformanceManager.h"
 #include "Diagnostics.h"
 #include "BinaryPathFinder.h"
+#include "dPlatforms.h"
 
 //RakNet includes:
 #include "RakNetDefines.h"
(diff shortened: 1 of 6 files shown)


aronwk-aaron (approved):

aronwk-aaron (approved):

dWorldServer/WorldServer.cpp
```diff
@@ -1284,12 +1285,14 @@ void WorldShutdownProcess(uint32_t zoneId) {
 }
 
 void WorldShutdownSequence() {
-	if (Game::shouldShutdown || worldShutdownSequenceComplete) {
+	Game::shouldShutdown = true;
```
EmosewaMC, line 1288: This is atomic because in multithreaded handling, the main loop will start the shutdown, and setting this to true across a thread will have no effect.

Jettford (approved):

PR #1341, opened 2023-12-18
fix: Using 'leave-zone' in Return to the Venture Explorer allows players to return to the original Venture Explorer map

jadebenn:
Since fixing this bug makes me feel a little like the fun police, I also added a config option to allow a server owner to intentionally retain this bug if they so choose.


EmosewaMC:
Eh I would rather either fix the bug entirely or remove this command.  I get fun police, but this isn't a feature we really want in main, especially labeled as a bug.  Just my 2 cents.

jadebenn:
It would be very easy to strip out the config option, but it's such a niche bug that I almost feel like I'm punishing the users who showed it to me. 😓

Will go with what you all think is best.

aronwk-aaron:
don't worry, I am the fun police, I can fix it :D

Wincent01 (approved):

aronwk-aaron (approved):

PR #1344, opened 2023-12-20
feat: Improve console output to show packet enum names (magic_enum)

jadebenn:
Replaces #1340. Adds magic_enum as a dependency in order to enable enum reflection and display world and packet message enum strings in the server output.


jadebenn:
Marking as draft because I had an epiphany about how to make the performance of our magic_enum implementation equal or exceed that of the preprocessor implementation. I will test this later, but here's the idea:

Instead of directly using the enum_name() function magic_enums provides, I will create a wrapper function called "StringifiedEnum" that defines a static const std::array provided by magic_enum's similarly-named enum_names() function. This function will then pass a reference to the corresponding array entry to the calling function. So, this should load a single copy of the pre-sorted(?) std::array into memory at compile-time, and it will be templated-out so those arrays are only generated for the enums this function is called on.

This should essentially eliminate any and all performance overhead magic_enums could possibly impose, as after compilation only an array of entries corresponding to the enum would exist, and a function to access it.

jadebenn:
Implemented more or less what I said I would. The desired performance has been achieved, but I'd appreciate a look at the iteration algorithm I'm using, which was borrowed from @EmosewaMC's work on the other branch. One issue I'm having in particular is that it doesn't handle "gaps" in the enum by returning a NULL value, and I don't currently understand arrays well enough to fix that.

jadebenn:
I think this is as good as it gets now. Much better performance than using the magic_enum function directly and restored the ability to see "gaps" in the enum. Only thing I haven't tested is if thhere are any enums out of order...

jadebenn:
Actually, we're good. Magic_enum gives us a sorted array.

dWorldServer/WorldServer.cpp
```diff
@@ -1244,7 +1245,11 @@ void HandlePacket(Packet* packet) {
 	}
 
 	default:
-		LOG("Unknown world packet received: %i", int(packet->data[3]));
+		if (packet->bitSize < (16 + sizeof(uint32_t) * 8)) break;
```
EmosewaMC, line 1248: This is already bounds checked far above in the same function if it gets this low

dWorldServer/WorldServer.cpp
```diff
@@ -1244,7 +1245,11 @@ void HandlePacket(Packet* packet) {
...
+
+		uint32_t messageId = *reinterpret_cast<uint32_t*>(&packet->data[3]);
```
EmosewaMC, line 1250: This should be able to just be reinterpret_cast straight to eWorldMessageType, can it not? You could also inline these two to be `const char* messageIdString = StringifiedEnum::ToString(*reinterpret_cast<eWorldMessageType*>(&packet->data[3])).data();`

tests/dCommonTests/dEnumsTests/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+set(DENUMS_TESTS
+	"MagicEnumTests.cpp"
+)
+
+# Get the folder name and prepend it to the files above
+get_filename_component(thisFolderName ${CMAKE_CURRENT_SOURCE_DIR} NAME)
+list(TRANSFORM DENUMS_TESTS PREPEND "${thisFolderName}/")
+
+# Export to parent scope
+set(DENUMS_TESTS ${DENUMS_TESTS} PARENT_SCOPE)
```
EmosewaMC, line 10: We don't write these like this right now, so for consistency this should follow how the other folder names are gotten (CMakeLists.txt in this directory just globs its files and sends to PARENT_SCOPE, then the parent scope prepends the folder name it just added)

tests/dCommonTests/dEnumsTests/MagicEnumTests.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+// Test Game Message Enum Reflection
+TEST(MagicEnumTest, eGameMessageTypeTest) {
+	#define log_test(y, z)\
+		LOG("%s %s", StringifiedEnum::ToString(static_cast<eGameMessageType>(y)).data(), #z);\
+		ASSERT_STREQ(StringifiedEnum::ToString(static_cast<eGameMessageType>(y)).data(), #z);
+
+	#define log_test_invalid(y)\
+		LOG("%s", StringifiedEnum::ToString(static_cast<eGameMessageType>(y)).data());\
+		ASSERT_STRNE(StringifiedEnum::ToString(static_cast<eGameMessageType>(y)).data(), NULL);
```
EmosewaMC, line 82: These can be combined with the above macro and then you can either use templates to plug in the tested type, or a define which takes an argument of the string to test to reduce repeated code

EmosewaMC: The final call in log_test_invalid can also be simplified to a .empty call with no need to compare with NULL.  What do you think of that change?

tests/dCommonTests/dEnumsTests/MagicEnumTests.cpp
```diff
@@ -0,0 +1,125 @@
...
+	log_test(1772, GET_IS_ON_RAIL);
+	log_test_invalid(32);
+	log_test_invalid(1776);
+
+	auto begin = std::chrono::high_resolution_clock::now();
+	for (int i = 0; i < 10000000; ++i) {
+		volatile auto f = StringifiedEnum::ToString(static_cast<eGameMessageType>(i)).data();
+
+		// To ensure the compiler doesn't optimize out the call, I print it at random intervals
+		if (rand() % 100000 == 0) LOG("%i, %s", i, f);
```
EmosewaMC, line 116: I would recommend since I didn't do this at the moment, using srand(time(NULL)) to at least seed the rng to the current epoch

CMakeLists.txt
```diff
@@ -292,6 +292,7 @@ set(INCLUDED_DIRECTORIES
 	"dScripts/zone/PROPERTY/GF"
 	"dScripts/zone/PROPERTY/NS"
 
+	"thirdparty/magic_enum/include/magic_enum"
```
EmosewaMC, line 295: if we are vendoring magic_enum, we should be using their cmake instead of just including their directories as we do for the other third parties

EmosewaMC (commented):
left some comments that should be talked about; also is there anything more to adding a submodule others need to be aware of when updating dlu or is the README already taking care of this?

tests/dCommonTests/dEnumsTests/CMakeLists.txt
aronwk-aaron: Currently, the tests are different and other tests cmakeslists are like this.
Leave it as is and we can make them consistent in a separate PR

tests/dCommonTests/dEnumsTests/MagicEnumTests.cpp
```diff
@@ -6,55 +6,58 @@
 #include "eGameMessageType.h"
 #include "eWorldMessageType.h"
 #include <chrono>
+#include <string>
+
+#define log_test(e, y, z)\
```
EmosewaMC, line 11: nit: macro should be caps locked

tests/dCommonTests/dEnumsTests/MagicEnumTests.cpp
```diff
@@ -66,48 +69,39 @@ TEST(MagicEnumTest, eWorldMessageTypeTest) {
...
+	log_test(eGameMessageType, 1767, MARK_INVENTORY_ITEM_AS_ACTIVE);
+	log_test(eGameMessageType, 1768, UPDATE_FORGED_ITEM);
+	log_test(eGameMessageType, 1769, CAN_ITEMS_BE_REFORGED);
+	log_test(eGameMessageType, 1771, NOTIFY_CLIENT_RAIL_START_FAILED);
+	log_test(eGameMessageType, 1772, GET_IS_ON_RAIL);
+	log_test_invalid(eGameMessageType, 32);
+	log_test_invalid(eGameMessageType, 1776);
 
+	srand(time(NULL));
 	auto begin = std::chrono::high_resolution_clock::now();
```
EmosewaMC, line 105: I also noticed the timing code is repeated, could that be simplified down to one call as well?

tests/dCommonTests/dEnumsTests/MagicEnumTests.cpp
```diff
@@ -6,55 +6,58 @@
 #include "eGameMessageType.h"
 #include "eWorldMessageType.h"
 #include <chrono>
+#include <string>
+
+#define log_test(e, y, z)\
+	LOG("%s %s", StringifiedEnum::ToString(static_cast<e>(y)).data(), #z);\
+	ASSERT_STREQ(StringifiedEnum::ToString(static_cast<e>(y)).data(), #z);
```
EmosewaMC, line 13: You dont need a .data call here I dont believe?  You can just compare the stringview directly to a const char*?

jadebenn:
That's it for me. Ready to merge this when you all are.

aronwk-aaron (approved):

dCommon/dEnums/StringifiedEnum.h
```diff
@@ -0,0 +1,28 @@
...
+#define __STRINGIFIEDENUM_H__
+
+#include <string>
+#include "magic_enum.hpp"
+
+namespace StringifiedEnum {
+	template<typename T>
+	const std::string_view ToString(const T e) {
+		constexpr auto sv = &magic_enum::enum_entries<T>();
+		std::string_view output {};
```
EmosewaMC, line 11: nit: default initializer unnecessary here.  std containers are already default initialized

dCommon/dEnums/StringifiedEnum.h
```diff
@@ -0,0 +1,28 @@
...
+		const auto it = std::lower_bound(
+			sv->begin(), sv->end(), e,
+			[&](const std::pair<T, std::string_view>& lhs, const T rhs)
+			{ return lhs.first < rhs; }
+		);
+
+		if (it != sv->end() && it->first == e)
+			output = it->second;
+		else
+			output = "UNKNOWN";
```
EmosewaMC, line 22: if these are going to be new lined, then we might as well use braces for clarity

tests/dCommonTests/dEnumsTests/MagicEnumTests.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+#define ASSERT_EARRAY_SORTED(EARRAY_VAR)\
+	for (int i = 0; i < EARRAY_VAR->size(); i++) {\
+		const auto entryCurr = EARRAY_VAR->at(i).first;\
+		LOG_EARRAY(wmArray, i, entryCurr);\
+		const auto entryNext = EARRAY_VAR->at(++i).first;\
+		LOG_EARRAY(wmArray, i, entryNext);\
+		ASSERT_TRUE(entryCurr < entryNext);\
+	};\
```
EmosewaMC, line 126: Is there a reason this is a macro and not a function?

EmosewaMC: Is there also a reason its always logging the first argument as wmArray, when it should be using the first param of LOG_EARRAY as EARRAY_VAR?

EmosewaMC (commented):
more feedback

tests/dCommonTests/dEnumsTests/MagicEnumTests.cpp
jadebenn: Nope, just a typo

jadebenn: As for it being a function: It's a real pain in the behind to deal with the array type as a function argument and this code will only be used here, so...

aronwk-aaron (approved):

EmosewaMC (approved):
Some odd spacing on the edited else but im not going to block on it.  Would recommend if you don't already alt+shift+f to format a file to the repo standard if it currently doesn't meet it (for vscode on windows at least this hotkey works)

PR #1348, opened 2023-12-22
fix: Character access after deletion

jadebenn:
Fixed case where the character could be accessed after the character was deleted.


jadebenn:
I _swore_ I pulled out all the debug stuff before publishing the branch. Hang on...

jadebenn:
Okay, all cleaned-up now.

EmosewaMC (approved):

aronwk-aaron (approved):

PR #1351, opened 2023-12-23
chore: miscellaneous code, formatting, and syntax cleanup

jadebenn:
This is kind of a grab-bag of miscellaneous changes I've wanted to do. Most of them do not affect functionality, except for one fairly large change which I will elaborate on below.

I've altered the StringifiedEnum::ToString() function to attempt to evaluate as a constexpr. However, doing this requires the use of a poorly-documented compiler flag and might break things. Part of why I'm making this PR is to check how the other compiler builds handle it. If y'all want that stripped out and/or it breaks things, I can do so. Otherwise, everything in here should be pretty non-controversial.

EDIT: Ended up dropping that particular change due to compiler portability issues, as seen in commit 57e274d below. The performance benefit was pretty much purely hypothetical anyway: I couldn't even measure it.

### DIFF
--- a/dCommon/GeneralUtils.h
+++ b/dCommon/GeneralUtils.h
@@ -242,6 +242,7 @@ namespace GeneralUtils {
 	template <typename T>
 	inline constexpr typename std::underlying_type<T>::type CastUnderlyingType(const T& entry) {
 		static_assert(std::is_enum<T>::value, "Not an enum");
+
 		return static_cast<typename std::underlying_type<T>::type>(entry);
 	}
 
@@ -256,6 +257,7 @@ namespace GeneralUtils {
 	inline T GenerateRandomNumber() {
 		// Make sure it is a numeric type
 		static_assert(std::is_arithmetic<T>::value, "Not an arithmetic type");
+
 		return GenerateRandomNumber<T>(std::numeric_limits<T>::min(), std::numeric_limits<T>::max());
 	}
 }
--- a/dCommon/GeneralUtils.h
+++ b/dCommon/GeneralUtils.h
@@ -239,11 +239,11 @@ namespace GeneralUtils {
 	 * @param entry Enum entry to cast
 	 * @returns The enum entry's value in its underlying type
 	*/
-	template <typename T>
-	inline constexpr typename std::underlying_type<T>::type CastUnderlyingType(const T& entry) {
-		static_assert(std::is_enum<T>::value, "Not an enum");
+	template <typename eType>
+	inline constexpr typename std::underlying_type_t<eType> CastUnderlyingType(const eType entry) {
+		static_assert(std::is_enum_v<eType>, "Not an enum");
 
-		return static_cast<typename std::underlying_type<T>::type>(entry);
+		return static_cast<typename std::underlying_type_t<eType>>(entry);
 	}
 
 	// on Windows we need to undef these or else they conflict with our numeric limits calls
--- a/dCommon/GeneralUtils.h
+++ b/dCommon/GeneralUtils.h
@@ -234,18 +234,28 @@ namespace GeneralUtils {
 		return T();
 	}
 
-// on Windows we need to undef these or else they conflict with our numeric limits calls
-// DEVELOPERS DEVELOPERS DEVELOPERS DEVELOPERS DEVELOPERS DEVELOPERS DEVELOPERS DEVELOPERS
-#ifdef _WIN32
-#undef min
-#undef max
-#endif
+	/**
+	 * Casts the value of an enum entry to its underlying type
+	 * @param entry Enum entry to cast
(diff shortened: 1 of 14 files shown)


dCommon/dEnums/StringifiedEnum.h
```diff
@@ -6,20 +6,26 @@
 
 namespace StringifiedEnum {
 	template<typename T>
-	const std::string_view ToString(const T e) {
+	constexpr std::string_view ToString(const T& e) {
+		// Check type
+		static_assert(std::is_enum<T>::value, "Not an enum");
```
EmosewaMC, line 11: These should have an _v variant so you dont need to do ::value if memory serves me correctly

dGame/dGameMessages/GameMessages.cpp
```diff
@@ -328,7 +328,7 @@ void GameMessages::SendPlayNDAudioEmitter(Entity* entity, const SystemAddress& s
 		bitStream.Write(static_cast<char>(audioGUID[k]));
 	}
 
-	bitStream.Write(uint32_t(0));
+	bitStream.Write(static_cast<uint32_t>(0));
```
EmosewaMC, line 331: In this case just remove the static cast and change this line to `bitStream.Write<uint32_t>(0);`

EmosewaMC: there are multiple occurances of this change so this applies to all

tests/dCommonTests/dEnumsTests/MagicEnumTests.cpp
```diff
@@ -140,3 +140,24 @@ TEST(MagicEnumTest, ArraysAreSorted) {
 
 	delete Game::logger;
 }
+
+#define IS_CONSTEXPR2(...) std::integral_constant<bool, __builtin_constant_p(([] () { __VA_ARGS__; } (), 0))>::value
```
EmosewaMC, line 144: Im not sure if im a fan of using an internal compiler function for this.  This is non-standard and will cause issues on other platforms that do not implement this function

dGame/dComponents/RebuildComponent.cpp
```diff
@@ -198,9 +198,9 @@ void RebuildComponent::Update(float deltaTime) {
 			if (!destComp) break;
 
 			++m_DrainedImagination;
-			const int32_t imaginationCostRemaining = m_TakeImagination - m_DrainedImagination;
+			const auto imaginationCostRemaining = m_TakeImagination - m_DrainedImagination;
 
-			const int32_t newImagination = destComp->GetImagination() - 1;
+			const auto newImagination = destComp->GetImagination() - 1;
```
EmosewaMC, line 203: I dont think we should change these, they work currently and should be forced ints on the off chance both the operands are unsigned and the subtraction underflows.

tests/dCommonTests/dEnumsTests/MagicEnumTests.cpp
jadebenn: Ah. That wasn't even supposed to be in there.

dCommon/GeneralUtils.h
```diff
@@ -234,18 +234,30 @@ namespace GeneralUtils {
...
+	/**
+	 * Casts the value of an enum entry to its underlying type
+	 * @param entry Enum entry to cast
+	 * @returns The enum entry's value in its underlying type
+	*/
+	template <typename T>
+	inline constexpr typename std::underlying_type<T>::type CastUnderlyingType(const T& entry) {
+		static_assert(std::is_enum<T>::value, "Not an enum");
+
+		return static_cast<typename std::underlying_type<T>::type>(entry);
```
EmosewaMC, line 246: can be shortened to underlying_type_t<T>

Would also be nice to have a better template name aside from the general T

dCommon/dEnums/StringifiedEnum.h
EmosewaMC: `is_enum_v<T>`

dGame/dGameMessages/GameMessages.cpp
jadebenn: Managed to convert all but 3, because those 3 need the intermediate step due to being a conversion from an enum

EmosewaMC:
How has this been tested?

EmosewaMC (commented):
See comment

jadebenn:
I'm playing around in the client right now, changing worlds and such and smashing physics objects. I'll quickly check with some enemies, but I haven't seen any issues.

EmosewaMC (approved):

jadebenn:
Went from pet cove -> NS -> Ninjago and completed a few quests. No issues I've seen.

aronwk-aaron (approved):

jadebenn:
Thank you all and happy holidays!

PR #1355, opened 2023-12-24
fix: more than 1 zone control object causing issues with scripted minigames

EmosewaMC:
Fixes an issue when players cannot load into Avant Gardens Survival and other script based minigames with more than 1 zone control object in the level files.  (why are there 26 zone control objects in BoNS and AGS)

### DIFF
--- a/dZoneManager/Level.cpp
+++ b/dZoneManager/Level.cpp
@@ -269,13 +269,13 @@ void Level::ReadSceneObjectDataChunk(std::istream& file, Header& header) {
 				else if (gating.featureName == Game::config->GetValue("event_8")) continue;
 				else if (!featureGatingTable->FeatureUnlocked(gating)) {
 					// The feature is not unlocked, so we can skip loading this object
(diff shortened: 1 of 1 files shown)


Wincent01 (approved):

dZoneManager/Level.cpp
```diff
@@ -269,13 +269,13 @@ void Level::ReadSceneObjectDataChunk(std::istream& file, Header& header) {
 				else if (gating.featureName == Game::config->GetValue("event_8")) continue;
 				else if (!featureGatingTable->FeatureUnlocked(gating)) {
 					// The feature is not unlocked, so we can skip loading this object
-					skipLoadingObject = true;
+					skipLoadingObject |= true;
```
Xiphoseer, line 272: I would argue that this diff doesn't do anything

Xiphoseer (approved):

aronwk-aaron (approved):

Jettford (approved):

PR #1363, opened 2023-12-27
chore: make scripted activity derived from activity

aronwk-aaron:
this paves the way for other component being properly derived from the activity component
Some docs on this on this [branch](https://github.com/DarkflameUniverse/DarkflameServer/blob/components-wheeeee/dGame/dComponents/ComponentHeirachy.md)

Tested that scripted activities such as foot races and the combat challenge still work as expected


dGame/dComponents/ActivityComponent.cpp
```diff
@@ -79,16 +75,12 @@ ScriptedActivityComponent::ScriptedActivityComponent(Entity* parent, int activit
 	}
 }
 
-ScriptedActivityComponent::~ScriptedActivityComponent()
-= default;
-
-void ScriptedActivityComponent::Serialize(RakNet::BitStream* outBitStream, bool bIsInitialUpdate) {
-	outBitStream->Write(true);
+void ActivityComponent::Serialize(RakNet::BitStream* outBitStream, bool bIsInitialUpdate) {
+	outBitStream->Write(!m_ActivityPlayers.empty());
```
aronwk-aaron, line 79: from looking at RE'd client code, this could be a dirty flag, instead of like a ... Activity Active flag which is how I'm using it here (since no players means not active)
If someone thinks this should be made more like a dirty flag, I can change this

Jettford: in theory they are the same thing.

aronwk-aaron: cool deal

Jettford (dismissed):

dGame/dComponents/ActivityComponent.cpp
aronwk-aaron: this is infact a dirty flag, and we'll need to serialize when there's no players in there

Jettford (approved):

EmosewaMC (commented):
feedback given in call

EmosewaMC (approved):
I assume this has been tested in various activities like Battle of Nimbus Station, Avant Gardens Survival and Combat Challenge?

Jettford (approved):

PR #1364, opened 2023-12-28
chore: rename rebuild to quickbuild

aronwk-aaron:
this is a non-function change
tested that quickbuilds still work and spot checked some of the scripts that use QB's as well

rebuild are a legacy pre-alpha gameplay feature that got re-worked into quickbuilds
renaming this to avoid confusion with things, since quickbuild is the common way to refer to them


dGame/dBehaviors/AreaOfEffectBehavior.cpp
```diff
@@ -7,7 +7,6 @@
 #include "Logger.h"
 #include "BehaviorBranchContext.h"
 #include "BehaviorContext.h"
-#include "RebuildComponent.h"
```
EmosewaMC, line 10: To hold us to non-functional changes/renaming truly being non-functional, I would like to request that this include be added back and just renamed for consistency.

dGame/dBehaviors/TacArcBehavior.cpp
```diff
@@ -6,7 +6,6 @@
 #include "BehaviorContext.h"
 #include "BaseCombatAIComponent.h"
 #include "EntityManager.h"
-#include "RebuildComponent.h"
```
EmosewaMC, line 9: Same here.  Wont comment on includes again

dGame/dComponents/LUPExhibitComponent.h
```diff
@@ -11,7 +11,7 @@
 class LUPExhibitComponent : public Component
 {
 public:
-	inline static const eReplicaComponentType ComponentType = eReplicaComponentType::EXHIBIT;
+	inline static const eReplicaComponentType ComponentType = eReplicaComponentType::LUP_EXHIBIT;
```
EmosewaMC, line 14: Like the above, this is a functional change.  See the main comment.

dGame/dGameMessages/GameMessages.cpp
```diff
@@ -4946,17 +4946,17 @@ void GameMessages::HandleRequestPlatformResync(RakNet::BitStream* inStream, Enti
...
+void GameMessages::HandleQuickBuildCancel(RakNet::BitStream* inStream, Entity* entity) {
 	bool bEarlyRelease;
 	LWOOBJID userID;
 
 	inStream->Read(bEarlyRelease);
 	inStream->Read(userID);
 
-	RebuildComponent* rebComp = static_cast<RebuildComponent*>(entity->GetComponent(eReplicaComponentType::QUICK_BUILD));
-	if (!rebComp) return;
+	auto* quickBuildComponent = entity->GetComponent<QuickBuildComponent>();
```
EmosewaMC, line 4956: Even if I agree with this one, it is a functional change

EmosewaMC (commented):
I have left some feedback suggesting that we punt the changes not related to renaming these files, even if I think they are fine, if we are touting this as just a "rename" with "non-function" changes, we should hold ourselves to that to reduce the risk of a change this large back firing.

dGame/dComponents/LUPExhibitComponent.h
aronwk-aaron: I think this was something I noticed in passing, fixed, and forgot, since I did these changes like 3 weeks ago.
Will split out into it's own PR since it is incorrect

EmosewaMC (approved):

Jettford (approved):

PR #1365, opened 2023-12-29
fix: LUP Exhibit component type and use delta compression on serialization

aronwk-aaron:
tested that this works


Jettford (approved):

EmosewaMC (approved):
Watched it work live, we fixed undefined behavior, emmo is a happy boi

EmosewaMC (approved):

Jettford (approved):

PR #1366, opened 2023-12-29
feat: get & print std::current_exception

Xiphoseer:
This improves the linux diagnostics code to print the current exception if there is one. This would probably work on windows as well, but this matches the `std::set_terminate` call, and as far as I can tell, the windows code does something else entirely.

c.f. <https://en.cppreference.com/w/cpp/error/exception_ptr>

### DIFF
--- a/dCommon/Diagnostics.cpp
+++ b/dCommon/Diagnostics.cpp
@@ -115,6 +115,13 @@ void GenerateDump() {
 }
 
 void CatchUnhandled(int sig) {
+	std::exception_ptr eptr = std::current_exception();
+	try {
+		if (eptr) std::rethrow_exception(eptr);
+	} catch(const std::exception& e) {
+		LOG("Caught exception: '%s'", e.what());
+	}
+
 #ifndef __include_backtrace__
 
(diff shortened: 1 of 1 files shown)


dCommon/Diagnostics.cpp
```diff
@@ -115,6 +115,17 @@ void GenerateDump() {
...
+	std::exception_ptr eptr = std::current_exception();
+	try
+    {
+        if (eptr)
+            std::rethrow_exception(eptr);
+    }
+    catch(const std::exception& e)
+    {
+        LOG("Caught exception: '%s'", e.what());
+    }
```
EmosewaMC, line 127: We currently use same line curly braces for try and catch like
```
try {

} catch (...) {

}
```
and for 1 line if statements it is preferred
` if (cond) logic;`

EmosewaMC (commented):
just a formatting pass would be great, looks good otherwise

EmosewaMC:
Windows does also produce the mdmp file which contains the current exception and much more info, so its ok if this doesn't apply to windows

dCommon/Diagnostics.cpp
Xiphoseer: Done. Sorry for all the CI triggers :D

aronwk-aaron (approved):

EmosewaMC (approved):

PR #1370, opened 2023-12-30
feat: add DLU_CONFIG_DIR env var

Xiphoseer:
This introduces a new env var `DLU_CONFIG_DIR` that can be set to be an absolute path or relative to the server binary where dConfig will look for its ini files. This is mostly useful for docker/k8s setups with mounted config files, everything else can keep using the main directory.

### DIFF
--- a/dCommon/dConfig.cpp
+++ b/dCommon/dConfig.cpp
@@ -10,7 +10,7 @@ dConfig::dConfig(const std::string& filepath) {
 	LoadConfig();
 }
 
-std::filesystem::path getConfigDir() {
+std::filesystem::path GetConfigDir() {
 	std::filesystem::path config_dir = BinaryPathFinder::GetBinaryDir();
 	if (const char* env_p = std::getenv("DLU_CONFIG_DIR")) {
 		config_dir /= env_p;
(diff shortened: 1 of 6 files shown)


aronwk-aaron (dismissed):

dCommon/dConfig.cpp
```diff
@@ -10,16 +10,31 @@ dConfig::dConfig(const std::string& filepath) {
 	LoadConfig();
 }
 
+std::filesystem::path getConfigDir() {
```
EmosewaMC, line 13: PascalCase for fns

dCommon/dConfig.cpp
```diff
@@ -10,16 +10,31 @@ dConfig::dConfig(const std::string& filepath) {
...
+	std::filesystem::path config_dir = BinaryPathFinder::GetBinaryDir();
+	if (const char* env_p = std::getenv("DLU_CONFIG_DIR")) {
```
EmosewaMC, line 15: nit: what would you think about this being a string_view instead of a raw const char*?

Xiphoseer: It's the actual return of that fn and it works with the nullptr check, its a std function that we can't change, and we immediately pass it to /=

dCommon/dConfig.cpp
Xiphoseer: Also for private/local ones?

EmosewaMC (approved):
lgtm

dCommon/dConfig.cpp
EmosewaMC: Yeah, we do PascalCast across the board afaik.  Don't think i've seen an occurrence where we use camelCase for a function

aronwk-aaron (approved):

PR #1373, opened 2023-12-31
fix: improve our docker config

### LINKED ISSUE #417: BUG: Docker Compose server fails to connect to database in some cases

### LINKED ISSUE #418: BUG: Server Docker container keeps running after server exited

### Issue Description

When the server exits inside the Docker container (e.g. because it is shut down, due to setup failure or crash), the container keeps running indefinitely, due to this:
https://github.com/DarkflameUniverse/DarkflameServer/blob/f6558aa31ecf873bacbc9b4d4366a89b5f24b4e2/docker/start_server.sh#L49

This is rather confusing because it is not obvious that the server has exited, see also related #417.
Side note: Currently the server might also erroneously exit with code 0 when the initial database connection fails; this will be fixed in #416, see [changed line](https://github.com/DarkflameUniverse/DarkflameServer/pull/416/files#diff-c5aa6f99cf9b296845bd9f91122e84694fe5b5891822904ba6ed964bfc63ea53R121).

I assume the reason why `tail -f /dev/null` was added is to inspect the container content after the server crashed. However, it appears there are alternatives such as [running the container again with a different entrypoint](https://stackoverflow.com/a/48892801).

### Reproduction steps

Use the Docker Compose setup and make the server startup fail, e.g. by modifying `docker-compose.yml` to use different credentials for the database by changing changing the user name line to `MARIADB_USER=someone`.

### Expected Behavior

The Docker server container should exit when the server exited; ideally also preserving the exit code.

### Environment

WSL 2; Docker

### LINKED ISSUE #570: BUILD: Docker Compose file shows error with .env file

- [X] I have read the [installation guide](https://github.com/DarkflameUniverse/DarkflameServer/blob/main/README.md).

### DarkflameServer Version

66215da

### Platform

Windows

### Architecture

x86

### Error Logs

When I try to run "docker compose up -d --build" while being in the server directory, it says either ACCOUNT_MANAGER_SECRET or CLIENT_PATH is missing a value. That is not true, because ACCOUNT_MANAGER_SECRET is empty as it is said in the Docker.md and I did put the right path to my client (I use the unpacked version if that maybe the error). 

I get the following errors:

<details>

invalid interpolation format for services.account-manager.environment.[]: "required variable ACCOUNT_MANAGER_SECRET is missing a value: missing_account_secret". You may need to escape any $ with another $

invalid interpolation format for services.darkflame.volumes.[]: "required variable CLIENT_PATH is missing a value: missing_client_path". You may need to escape any $ with another $

</details>

aronwk-aaron:
greatly simplify docker and automaically build container and push to ghcr.io
fixes #570 
fixes #418
fixes #417

- Removes account manager as a submodule and repalces it with the nexus dashboard docker container
- Removed all the customs dockerfiles for things like mariadb
- Removed the need for users to build the image locally
- Change port config name in each servers to specific to what sever it is as to avoid env var name collisions
- Tested that the docker image works by locally building it and connecting to it with a client


Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+	cmake .. && \
+	make -j2
+
+FROM gcc:latest as runtime
+
+WORKDIR /app
+
+COPY --from=build /app/build/*Server /app/
+COPY --from=build /app/build/*.ini /app/default-configs/
+COPY --from=build /app/build/*.so /usr/lib/
```
Xiphoseer, line 24: While this is technically correct, without another fix, this will be missing the `libmariadb.so`.

README.md
```diff
@@ -347,6 +348,43 @@ certutil -hashfile <file> SHA1
...
+    -e DLU_CONFIG_DIR=/app/configs \
+    -e DUMP_FOLDER=/app/dumps \
+    -p 1001:1001/udp \
+    -p 2005:2005/udp \
+    -p 3000-3300:3000-3300/udp \
+ghcr.io/darkflameuniverse/darkfameserver:latest
+```
+You will need to replace the `/path/to/`'s to reflect the paths on your host.
+
+Any config option in the `.ini`'s can be overridden with environmental variables: Ex: `log_to_console=1` from `shared_config.ini` would be overidden like  `-e LOG_TO_CONSOLE=0`
```
Xiphoseer, line 378: Maybe mention that all servers will pick it up, so service specific settings may have trouble and should use `{world,chat,auth,master}config.ini`

Xiphoseer: Ah neat, you fixed the cases where that would have been an issue (port)

aronwk-aaron: oh yeah, I think i put that in the PR description when opening it for review

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+WORKDIR /app
+
+RUN set -ex; \
+    apt-get update; \
+    apt-get install -y cmake
+
+COPY . /app/
+
+RUN sed -i 's/MARIADB_CONNECTOR_COMPILE_JOBS__=1/MARIADB_CONNECTOR_COMPILE_JOBS__=2/' /app/CMakeVariables.txt
```
EmosewaMC, line 11: This should be looking for `MARIADB_CONNECTOR_COMPILE_JOBS__=` instead of `MARIADB_CONNECTOR_COMPILE_JOBS__=1` as the latter will not be found if we on the master repo change it from the default

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+    apt-get install -y cmake
+
+COPY . /app/
+
+RUN sed -i 's/MARIADB_CONNECTOR_COMPILE_JOBS__=1/MARIADB_CONNECTOR_COMPILE_JOBS__=2/' /app/CMakeVariables.txt 
+
+RUN mkdir build
+RUN cd build && \
+	cmake .. && \
+	make -j2
```
EmosewaMC, line 16: Can this not be replaced with `./build.sh -j2`?  That file does the same work as is done here

entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+cp -n /app/default-configs/* /app/configs/
```
Xiphoseer, line 3: Why copy in the entrypoint? Can you remove this and put master at the ENTRYPOINT directly?

Xiphoseer (dismissed):
with nits

entrypoint.sh
aronwk-aaron: this puts the files on disk. Which will be mounted, and make these files persistent.
This gives users the option to edit the configs on disk instead of using env vars if they so wish.

Same with vanity, allows users to edit them on the mounted disk.

I'm not sure if there's a better way to do this.
But I always like when images do this and I don't need to go manually grab default configs from whereever to put on my host disk

README.md
```diff
@@ -347,6 +348,43 @@ certutil -hashfile <file> SHA1
...
+    -v /path/to/dumps/:/app/dumps \
+    -v /path/to/res:/app/res:ro \
+    -v /path/to/resServer:/app/resServer \
+    -e DLU_CONFIG_DIR=/app/configs \
+    -e DUMP_FOLDER=/app/dumps \
+    -p 1001:1001/udp \
+    -p 2005:2005/udp \
+    -p 3000-3300:3000-3300/udp \
+ghcr.io/darkflameuniverse/darkfameserver:latest
+```
```
EmosewaMC, line 375: imo this should be moved to a shell.example file for people to edit instead in one place to just cut the middle man step of needing to copy this text into a file and instead of copy pasting a file and renaming it.  Plus then we can reference the example file by name and not have a huge docker container command in the readme

Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM gcc:latest as build
```
Xiphoseer, line 1: Just noticed: this should be a number, not `latest`, otherwise it'll break on its own down the line. Can be just the major semver component though, to get a bit of updates automatically.

docker-compose.yml
```diff
@@ -1,104 +1,147 @@
...
+      - HELP_2_SUMMARY=${HELP_2_SUMMARY:-Missing features?}
+      - HELP_3_SUMMARY=${HELP_3_SUMMARY:-Get smashed?}
+      - HELP_4_SUMMARY=${HELP_4_SUMMARY:-Want to contribute?}
+      - HELP_0_DESCRIPTION=${HELP_0_DESCRIPTION:-Go to the DarkflameServer repository on GitHub to view issues and discussions about the server emulator!<br/><br/><a href="https://github.com/DarkflameUniverse/DarkflameServer"><font color="#0000EE">Click Here to go there!</font></a><br/><br/>}
+      - HELP_1_DESCRIPTION=${HELP_1_DESCRIPTION:-Try switching networks, using a VPN, or using your phone's hotspot to resolve the issue.<br/><br/>}
+      - HELP_2_DESCRIPTION=${HELP_2_DESCRIPTION:-While DarkflameServer is a mostly complete emulator, there are still some features that aren't implemented. You can track these on the GitHub issues page.<br/><br/>}
+      - HELP_3_DESCRIPTION=${HELP_3_DESCRIPTION:-Skill issue!<br/><br/>}
+      - HELP_4_DESCRIPTION=${HELP_4_DESCRIPTION:-Visit Discussions on the DarkflameServer GitHub page<br/>to ask questions and collaborate with other devs!<br/><br/>}
+      - ALLOW_PLAYERS_TO_SKIP_CINEMATICS=${ALLOW_PLAYERS_TO_SKIP_CINEMATICS:-0}
```
EmosewaMC, line 96: I'm still not a fan of this huge copy paste here which is prone to errors, forgetting to update and is just messy overall.  Is there really no other possible way to do this more efficiently?

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+COPY . /app/
+
+RUN sed -i 's/MARIADB_CONNECTOR_COMPILE_JOBS__=1/MARIADB_CONNECTOR_COMPILE_JOBS__=2/' /app/CMakeVariables.txt 
+
+RUN mkdir build
+RUN cd build && \
+	cmake .. && \
+	make -j2
+
+FROM gcc:latest as runtime
```
Xiphoseer, line 18: same here, consider `debian:12`

Dockerfile
aronwk-aaron: yep, good catch. I meant to do this

Dockerfile
aronwk-aaron: Will try this

EmosewaMC (dismissed):
Because we are changing the port option name, we need to add a block of code to the CMakeLists.txt file to update a line that starts with `port=` and updates it to say auth_port etc.  This is a breaking change otherwise as some users may have had to change some of the default ports for things like WorldServer start or where MasterServer resides due to port collisions

README.md
aronwk-aaron: this is standard to have in the readme for containers

docker-compose.yml
aronwk-aaron: Just going to remove most of them and leave only the necessary ones

Dockerfile
aronwk-aaron: will need to modify the sed to just replace the whole line then

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+COPY --from=build /app/build/vanity /app/default-vanity
+COPY --from=build /app/build/navmeshes /app/navmeshes
+COPY --from=build /app/build/migrations /app/migrations
+COPY --from=build /app/build/*.dcf /app/
+COPY ./entrypoint.sh /app/
+RUN ldconfig
+# needed as the container runs with the root user
+# and therefore sudo doesn't exist
+ENV USE_SUDO_AUTH=0
+RUN chmod +x entrypoint.sh
```
EmosewaMC, line 34: Can entrypoint.sh not just have execute perms by default?

README.md
```diff
@@ -347,6 +348,43 @@ certutil -hashfile <file> SHA1
 Known good *SHA1* checksum of the Darkflame Universe client:
 - `91498e09b83ce69f46baf9e521d48f23fe502985` (packed client, zip compressed)
 
+# Docker
+
+## Standalone
+
+For standalone deployment, you can use the image provided via Github's Container Repository:
+`ghcr.io/darkflameuniverse/darkfameserver`
```
EmosewaMC, line 356: `darkfameserver`

README.md
EmosewaMC: Can't say I agree but alright

entrypoint.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+mkdir -p /app/configs/
+cp -n /app/default-configs/* /app/configs/
+mkdir -p /app/vanity/
+cp -n /app/default-vanity/* /app/vanity/
+cd /app/ && ./MasterServer
```
EmosewaMC, line 6: Should this file have `set -e` at the start?

Dockerfile
aronwk-aaron: I tried setting it to +x on my host and then building the container. And they did not work. So yes, this is required

entrypoint.sh
aronwk-aaron: Potentially, won't hurt to add it

Dockerfile
aronwk-aaron: https://stackoverflow.com/questions/27281965/docker-file-chmod-on-entrypoint-script

aronwk-aaron:
re-tested that this works and connected to a server running in docker with a client on the same host

entrypoint.sh
Xiphoseer: Some thoughts:
- I would argue this cp -n is worse than none by default because it doesn't diff, so once you upte you have stale config sitting there preventing a copy and you need to merge by hand again. I do see the appeal of default config files showing up on their own though.
- I'm worried that this is another place added to the already complex master startup that does magic. For k8s for example I need to rely on being able to replace entrypoint with WorldServers so this script must not have load bearing (fatal when skipped) side effects

aronwk-aaron: Stale configs is just part of it unfortunately. There is no remedy other than making the server possibly update the configs with new setting during startup and providing a sane default (which we could do that since we'll have the latest default configs in the container always)

All the necessary configs already exist and any new configs should only be superfluous with the default behavior that the setting would change being what happened before it was added.

This is 100% just a way to get a bootstrap on a fresh container.
Users of docker know they can modify the entrypoint in their run command if they need to do so, which I've personally rarely had to do. I'm assuming you're already modifying the entrypoint in k8's since you're deploying individual servers to separate pods (I think that terminology is correct)

aronwk-aaron: > There is no remedy other than making the server possibly update the configs with new setting during startup and providing a sane default

Having the configs be writable may be a concern as well, but that's a different discussion imo

Also, could just forsake this all and force the user to either do everything via env vars or know they need to bind mount the configs and the go grab them from the repo

And we should probably remove the dependency for the configs files needing to exist at all as well, but that can be a separate PR.

This is all just to get a MVP for docker without any server changes

aronwk-aaron:
`s/crash/exit/` on the last commit message

dMasterServer/MasterServer.cpp
```diff
@@ -81,43 +81,18 @@ int main(int argc, char** argv) {
...
-
-	if (!dConfig::Exists("worldconfig.ini")) {
-		LOG("Couldnt find worldconfig.ini");
-		return EXIT_FAILURE;
-	}
+	if (!dConfig::Exists("authconfig.ini")) LOG("Couldnt find authconfig.ini.");
+	if (!dConfig::Exists("chatconfig.ini")) LOG("Couldnt find chatconfig.ini");
+	if (!dConfig::Exists("masterconfig.ini")) LOG("Couldnt find masterconfig.ini");
+	if (!dConfig::Exists("sharedconfig.ini")) LOG("Couldnt find sharedconfig.ini");
+	if (!dConfig::Exists("worldconfig.ini")) LOG("Couldnt find worldconfig.ini");
```
EmosewaMC: Instead of removing checks like this please instead consider checking if there is an env var set which allows the bypass of the check inatead

dMasterServer/MasterServer.cpp
```diff
@@ -81,43 +81,18 @@ int main(int argc, char** argv) {
...
-	uint32_t clientNetVersion = 0;
-	if (!GeneralUtils::TryParse(Game::config->GetValue("client_net_version"), clientNetVersion)) {
-		LOG("Failed to parse (%s) as net version. Cannot start server as no clients could connect.",Game::config->GetValue("client_net_version").c_str());
-		LOG("As of version 1.1.1, client_net_version is required to be defined in sharedconfig.ini as opposed to in CMakeVariables.txt as NET_VERSION.");
-		LOG("Rerun cmake to ensure all config values exist. If client_net_version already exists in sharedconfig.ini, please ensure it is a valid number.");
-		LOG("like 171022");
-		return EXIT_FAILURE;
-	}
+	uint32_t clientNetVersion = 171022;
+	GeneralUtils::TryParse(Game::config->GetValue("client_net_version"), clientNetVersion);
```
EmosewaMC: Same here.

dMasterServer/MasterServer.cpp
aronwk-aaron: yep, already reveted these chagnes and am going to consider these changes in a different PR

Xiphoseer (dismissed):

EmosewaMC (approved):

Xiphoseer (approved):

PR #1375, opened 2024-01-01
fix: signal handling

Xiphoseer:
This PR improves graceful shutdown of the server binaries in the following ways:
- Fix signal handling (e.g. SIGTERM, Ctrl-C), so that we do as little as possible in the handler itself. This is mandated by POSIX because handlers might be called while inside functions like printf and it's not safe to call a logger or almost anything in there.
- Flush the logger in more places, including immediately before the server loops and in its destructor.
- Add SIGINT and SIGTERM handling to Auth and Chat.

WorldServer hasn't been updated completely yet, because I haven't gotten around to test that.

### DIFF
--- a/dAuthServer/AuthServer.cpp
+++ b/dAuthServer/AuthServer.cpp
@@ -1,6 +1,7 @@
 #include <iostream>
 #include <string>
 #include <ctime>
+#include <csignal>
 #include <chrono>
 #include <thread>
 
@@ -28,12 +29,17 @@ namespace Game {
 	Logger* logger = nullptr;
 	dServer* server = nullptr;
 	dConfig* config = nullptr;
-	bool shouldShutdown = false;
+	Game::signal_t lastSignal = 0;
 	std::mt19937 randomEngine;
 }
 
 Logger* SetupLogger();
 void HandlePacket(Packet* packet);
+
+void OnSignal(int signal)
+{
+    Game::lastSignal = signal;
+}
 
 int main(int argc, char** argv) {
 	constexpr uint32_t authFramerate = mediumFramerate;
@@ -41,6 +47,9 @@ int main(int argc, char** argv) {
 	Diagnostics::SetProcessName("Auth");
 	Diagnostics::SetProcessFileName(argv[0]);
 	Diagnostics::Initialize();
+
+	std::signal(SIGINT, OnSignal);
+	std::signal(SIGTERM, OnSignal);
 
 	//Create all the objects we need to run our service:
 	Game::logger = SetupLogger();
@@ -74,6 +83,7 @@ int main(int argc, char** argv) {
 		masterIP = masterInfo->ip;
 		masterPort = masterInfo->port;
 	}
+	LOG("Master is at %s:%d", masterIP.c_str(), masterPort);
 
 	Game::randomEngine = std::mt19937(time(0));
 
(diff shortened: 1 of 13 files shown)


dDatabase/GameDatabase/MySQL/Tables/Servers.cpp
```diff
@@ -14,10 +15,8 @@ std::optional<IServers::MasterInfo> MySQLDatabase::GetMasterInfo() {
...
 
-	MasterInfo toReturn;
-
-	toReturn.ip = result->getString("ip").c_str();
-	toReturn.port = result->getInt("port");
-
-	return toReturn;
+	return MasterInfo {
+		.ip = result->getString("ip").c_str(),
+		.port = static_cast<uint32_t>(result->getInt("port")),
```
EmosewaMC, line 20: this is only available in C++20

aronwk-aaron (approved):

aronwk-aaron (approved):

dAuthServer/AuthServer.cpp
```diff
@@ -28,20 +29,28 @@ namespace Game {
...
-	bool shouldShutdown = false;
+	Game::signal_t lastSignal = 0;
 	std::mt19937 randomEngine;
 }
 
 Logger* SetupLogger();
 void HandlePacket(Packet* packet);
 
+void OnSignal(int signal)
+{
```
EmosewaMC, line 40: nit: same line curly braces

dChatServer/ChatServer.cpp
```diff
@@ -33,21 +33,29 @@ namespace Game {
...
+	Game::signal_t lastSignal = 0;
 	std::mt19937 randomEngine;
 	PlayerContainer playerContainer;
 }
 
 Logger* SetupLogger();
 void HandlePacket(Packet* packet);
 
+void OnSignal(int signal)
+{
```
EmosewaMC, line 45: This block of code is repeated, would this be able to be moved to Servers.cpp or does each server require a definition like this?

dNet/dServer.cpp
```diff
@@ -39,7 +39,7 @@ class ReplicaReceiever : public ReceiveDownloadCompleteInterface {
 	}
 } ReceiveDownloadCompleteCB;
 
-dServer::dServer(const std::string& ip, int port, int instanceID, int maxConnections, bool isInternal, bool useEncryption, Logger* logger, const std::string masterIP, int masterPort, ServerType serverType, dConfig* config, bool* shouldShutdown, unsigned int zoneID) {
+dServer::dServer(const std::string& ip, int port, int instanceID, int maxConnections, bool isInternal, bool useEncryption, Logger* logger, const std::string masterIP, int masterPort, ServerType serverType, dConfig* config, Game::signal_t* lastSignal, unsigned int zoneID) {
```
EmosewaMC, line 42: Using Game namespace here kind of defeats the purpose of us passing a bunch if variables down to server, but that is something to fix later.

dNet/dServer.cpp
```diff
@@ -75,7 +75,9 @@ dServer::dServer(const std::string& ip, int port, int instanceID, int maxConnect
 	//Connect to master if we are not master:
 	if (serverType != ServerType::Master) {
 		SetupForMasterConnection();
-		ConnectToMaster();
+		if (!ConnectToMaster()) {
```
EmosewaMC, line 78: I believe this log will be a bit misleading; if I recall correctly, RakNet returning true from connect does not mean you are connected.  Is there a different way to phrase this possible failure to connect ti master?

dChatServer/ChatServer.cpp
Xiphoseer: Yes

aronwk-aaron (approved):

EmosewaMC (approved):

PR #1378, opened 2024-01-01
Feature: K8s

Xiphoseer:
This works for minikube, but is very much still WIP


Rindula:
If you need someone to test this somewhere else, I have a single node [k3s](https://k3s.io/) cluster at home where I could run this.

Rindula:
Using `hostpath` for the PVC is not quite working, is there a way to make this using the default storage provider?

Xiphoseer:
> Using `hostpath` for the PVC is not quite working, is there a way to make this using the default storage provider?

Probably yes, this was done as a first draft for use with `minikube mount /path/to/client:/luclient`. It's also raw `hostPath` in a pod not even a PVC.

For a complete setup, you're right, this should be a configurable PersistentVolume

PR #1379, opened 2024-01-02
feat: Add config update function

EmosewaMC:
Appends a new option to the specified file which contains the same contents as the old option, but with the updated name specified in the high level cmake.

Tested that I can run the cmake multiple times and a singular extra config option is generated with the same value as was in the previous option.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -122,3 +122,4 @@ docker/__pycache__
 docker-compose.override.yml
 
 !*Test.bin
+!cmake/*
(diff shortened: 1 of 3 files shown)


aronwk-aaron (dismissed):

Jettford (changes requested):
This should be in a .cmake file as a utility.

EmosewaMC:
there are you happy cmake

aronwk-aaron (dismissed):

Xiphoseer (approved):

Jettford (dismissed):

aronwk-aaron (approved):

Xiphoseer (approved):

PR #1380, opened 2024-01-02
fix: nullptr access for logger in master

EmosewaMC:
Tested that my characters position and data saves on shutdown now and that master no longer hangs on shutdown

### DIFF
--- a/dWorldServer/WorldServer.cpp
+++ b/dWorldServer/WorldServer.cpp
@@ -1291,13 +1291,16 @@ void WorldShutdownProcess(uint32_t zoneId) {
 }
 
 void WorldShutdownSequence() {
+	bool shouldShutdown = Game::ShouldShutdown() || worldShutdownSequenceComplete;
 	Game::lastSignal = -1;
 #ifndef DARKFLAME_PLATFORM_WIN32
-	if (Game::ShouldShutdown() || worldShutdownSequenceComplete)
+	if (shouldShutdown)
 #endif
 	{
 		return;
 	}
+
+	if (!Game::logger) return;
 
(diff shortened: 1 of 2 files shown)


aronwk-aaron (approved):

dWorldServer/WorldServer.cpp
```diff
@@ -1291,14 +1291,17 @@ void WorldShutdownProcess(uint32_t zoneId) {
...
 	Game::lastSignal = -1;
 #ifndef DARKFLAME_PLATFORM_WIN32
-	if (Game::ShouldShutdown() || worldShutdownSequenceComplete)
+	if (shouldShutdown)
 #endif
 	{
 		return;
 	}
 
+	if (!Game::logger) return;
```
Xiphoseer, line 1303: Can we printf when this happens? This still method still feels weird in general.

Xiphoseer (approved):

Xiphoseer:
Was this on windows? Preferably we'd sort out differences instead of silently returning on weird conditions all over the place.

aronwk-aaron:
> Was this on windows? Preferably we'd sort out differences instead of silently returning on weird conditions all over the place.

If it was with windows, it's to do with the signal spawning a separate thread to handle the shutdown, which id a windows specific thing

EmosewaMC:
> Was this on windows? Preferably we'd sort out differences instead of silently returning on weird conditions all over the place.

This is a non windows issue and causes master to hang indefinitely on my wsl instance.

EmosewaMC:
Windows must have that define.  No other way around it

PR #1385, opened 2024-01-03
change name for docker ci to be caps like the other

EmosewaMC:


aronwk-aaron (approved):

aronwk-aaron:
Should it just be renamed docker-build or something like that, as to make it different?
Since it's not the "CI" pipeline?

Fine with either decision, apologies for not thinking about this when copying it over from nexus dash

Xiphoseer (approved):

PR #1386, opened 2024-01-04
fix: dont crash if resServer doesn't exist

aronwk-aaron:
and try to make resServer as well if it doesn't exist

Tested that this works

### DIFF
--- a/dMasterServer/MasterServer.cpp
+++ b/dMasterServer/MasterServer.cpp
@@ -150,31 +150,40 @@ int main(int argc, char** argv) {
 	}
 
 	MigrationRunner::RunMigrations();
-
-	const bool cdServerExists = std::filesystem::exists(BinaryPathFinder::GetBinaryDir() / "resServer" / "CDServer.sqlite");
+	const auto resServerPath = BinaryPathFinder::GetBinaryDir() / "resServer";
+	const bool cdServerExists = std::filesystem::exists(resServerPath / "CDServer.sqlite");
 	const bool oldCDServerExists = std::filesystem::exists(Game::assetManager->GetResPath() / "CDServer.sqlite");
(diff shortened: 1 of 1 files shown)


dMasterServer/MasterServer.cpp
```diff
@@ -154,6 +154,15 @@ int main(int argc, char** argv) {
 	const bool cdServerExists = std::filesystem::exists(BinaryPathFinder::GetBinaryDir() / "resServer" / "CDServer.sqlite");
 	const bool oldCDServerExists = std::filesystem::exists(Game::assetManager->GetResPath() / "CDServer.sqlite");
 	const bool fdbExists = std::filesystem::exists(Game::assetManager->GetResPath() / "cdclient.fdb");
+	const bool resServerPathExists = std::filesystem::is_directory(BinaryPathFinder::GetBinaryDir() / "resServer");
+
+	if (!resServerPathExists) {
+		LOG("%s does not exist, creating it.", (BinaryPathFinder::GetBinaryDir() / "resServer").c_str());
```
EmosewaMC, line 160: Lets do this concatenation of binary dir and resServer once
also .c_str on a std::filesystem::path returns a wstring on windows, so this logger call should be .string().c_str()

aronwk-aaron: was following the convention below with the .c_str()

EmosewaMC: yes and that is incorrect and should be changed as well.  Unix uses std::string for its file paths and windows uses std::wstring for its

Xiphoseer (approved):

EmosewaMC (commented):
there is a bunch of extra copies being done with all the .string() calls but its not enough to block.  lgtm

EmosewaMC (approved):

PR #1387, opened 2024-01-04
fix: Dissolve more CMake dependencies

Xiphoseer:
This PR (arguably?) improves the CMake setup a bit. A bunch of folders are now `OBJECT` libraries and more cmake code was moved to the `cmake` folder. The global include list has been cut down, and each library now needs to opt in what other libraries to compile against.

Beware that circular dependencies on static libs are possible, but discouraged and can lead to pathological cases when cmake tries to figure out in what order to execute them and may end up with no parallelism at all.


tests/dCommonTests/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ list(APPEND DCOMMONTEST_SOURCES ${DENUMS_TESTS})
 add_executable(dCommonTests ${DCOMMONTEST_SOURCES})
 
 # Link needed libraries
-target_link_libraries(dCommonTests ${COMMON_LIBRARIES} GTest::gtest_main)
+target_link_libraries(dCommonTests ${COMMON_LIBRARIES} dGame GTest::gtest_main)
```
EmosewaMC, line 22: Is this addition necessary?  dCommonTests shouldn't need anything aside from common libs

Xiphoseer: Generally yes, you are meant to attach headers to INTERFACE_INCLUDE_DIRECTORIES of a target and pull that in as a link library which automatically adds the include search path and there is at least one test that included dGame headers.

I'm open for suggestions though, would just like to keep the cpp/h files unchanged in these PRs

EmosewaMC: I see what you are talking about with one test using those headers.  Those can be safely removed.  They were a result of copy and paste from a previous test and are fully unneeded for common tests as we dont even link with dGame previously, so we couldn't even use the libraries from there anyways.  I think its better to just remove the two offending headers than to add a full library to link against, but its up to you.

EmosewaMC: i just put up a pr to resolve this issue.  This link should not be necessary anymore.

Xiphoseer: Sure. I'll rebase when this one is in and remove the link

Xiphoseer:
This PR is still not cleaned up completely, but please help review and test it.

aronwk-aaron:
tested that it builds on my local dev laptop (manjaro) and a test server (Ubuntu 18.04)

Jettford (approved):
Builds and runs on Windows under x64 Clang and MSVC.

CMakeLists.txt
```diff
@@ -76,6 +76,7 @@ endif()
 
 # Our output dir
 set(CMAKE_BINARY_DIR ${PROJECT_BINARY_DIR})
+#set(CMAKE_INCLUDE_CURRENT_DIR_IN_INTERFACE ON) # unfortunately, serializes everything
```
EmosewaMC, line 79: what does 'serializes everything' mean?

CMakeLists.txt
```diff
@@ -201,40 +204,23 @@ foreach(file ${SQL_FILES})
...
-
-	"dDatabase"
-	"dDatabase/CDClientDatabase"
-	"dDatabase/CDClientDatabase/CDClientTables"
-	"dDatabase/GameDatabase"
-	"dDatabase/GameDatabase/ITables"
-	"dDatabase/GameDatabase/MySQL"
-	"dDatabase/GameDatabase/MySQL/Tables"
-
 	"dNet"
```
EmosewaMC, line 224: Any particular reason we are keeping some of these include dirs here?

CMakeLists.txt
```diff
@@ -261,27 +247,8 @@ foreach(dir ${INCLUDED_DIRECTORIES})
...
-# Glob together all headers that need to be precompiled
-file(
-	GLOB HEADERS_DDATABASE
-	LIST_DIRECTORIES false
-	${PROJECT_SOURCE_DIR}/dDatabase/CDClientDatabase/*.h
-	${PROJECT_SOURCE_DIR}/dDatabase/CDClientDatabase/CDClientTables/*.h
-	${PROJECT_SOURCE_DIR}/dDatabase/GameDatabase/ITables/*.h
-	${PROJECT_SOURCE_DIR}/thirdparty/SQLite/*.h
-)
+# link_directories(${PROJECT_BINARY_DIR})
```
EmosewaMC, line 251: We should just remove this if its unused

cmake/FindMariaDB.cmake
```diff
@@ -85,77 +86,66 @@ else() # Build from source
 			-DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0)
 	endif()
 
+	set(MARIADBCPP_INSTALL_DIR ${PROJECT_BINARY_DIR}/prefix)
+	message("MariaDB C/C++ install prefix: " ${MARIADBCPP_INSTALL_DIR})
```
EmosewaMC, line 90: Does this need STATUS?

cmake/FindMariaDB.cmake
```diff
@@ -85,77 +86,66 @@ else() # Build from source
...
-		set(MARIADB_PLUGINS_LOCATION "${BINARY_DIR}/libmariadb/RelWithDebInfo")
-	else()
-		set(MARIADB_SHARED_LIBRARY_LOCATION "${BINARY_DIR}/${MARIADB_SHARED_LIBRARY_NAME}")
-		set(MARIADB_SHARED_LIBRARY_COPY_LOCATION "${PROJECT_BINARY_DIR}")
-		set(MARIADB_PLUGINS_LOCATION "${BINARY_DIR}/libmariadb")
+	set(MARIADBCPP_SHARED_LIBRARY_LOCATION "${MARIADBCPP_LIBRARY_DIR}/${MARIADB_SHARED_LIBRARY_NAME}")
+	if(WIN32)
+		set(MARIADBC_SHARED_LIBRARY_LOCATION "${MARIADBCPP_LIBRARY_DIR}/libmariadb.lib")
+	#elseif(UNIX)
+	#	set(MARIADBC_SHARED_LIBRARY_LOCATION "${MARIADBCPP_LIBRARY_DIR}/libmariadb.so.3")
```
EmosewaMC, line 128: If these are unused they should be removed

cmake/FindMariaDB.cmake
```diff
@@ -85,77 +86,66 @@ else() # Build from source
...
-# Remove the CMakeLists.txt file from the tests folder for the maria-db-connector so we dont compile the tests.
-if(EXISTS "${CMAKE_SOURCE_DIR}/thirdparty/mariadb-connector-cpp/test/CMakeLists.txt")
-	file(REMOVE "${CMAKE_SOURCE_DIR}/thirdparty/mariadb-connector-cpp/test/CMakeLists.txt")
 endif()
 
 # Create mariadb connector library object
-add_library(mariadbConnCpp SHARED IMPORTED GLOBAL)
-set_property(TARGET mariadbConnCpp PROPERTY IMPORTED_LOCATION ${MARIADB_SHARED_LIBRARY_LOCATION})
+message("libmariadb: ${MARIADBC_SHARED_LIBRARY_LOCATION}")
+message("libmariadbcpp: ${MARIADBCPP_SHARED_LIBRARY_LOCATION}")
```
EmosewaMC, line 134: STATUS here?

dDatabase/MigrationRunner.cpp
EmosewaMC, line 1: Is there a particular reason this was moved up a directory?  It doesn't apply to the both game and cdclient databases which is why i put it under GameDatabase

dGame/CMakeLists.txt
```diff
@@ -28,7 +42,26 @@ add_subdirectory(dMission)
...
+	$<TARGET_PROPERTY:dGameBase,INTERFACE_INCLUDE_DIRECTORIES>
+	$<TARGET_PROPERTY:dBehaviors,INTERFACE_INCLUDE_DIRECTORIES>
+	$<TARGET_PROPERTY:dComponents,INTERFACE_INCLUDE_DIRECTORIES>
+	$<TARGET_PROPERTY:dEntity,INTERFACE_INCLUDE_DIRECTORIES>
+	$<TARGET_PROPERTY:dGameMessages,INTERFACE_INCLUDE_DIRECTORIES>
+	$<TARGET_PROPERTY:dInventory,INTERFACE_INCLUDE_DIRECTORIES>
+	$<TARGET_PROPERTY:dMission,INTERFACE_INCLUDE_DIRECTORIES>
+	$<TARGET_PROPERTY:dPropertyBehaviors,INTERFACE_INCLUDE_DIRECTORIES>
+	$<TARGET_PROPERTY:dUtilities,INTERFACE_INCLUDE_DIRECTORIES>
+)
```
EmosewaMC, line 67: nit: eof

dGame/dComponents/CMakeLists.txt
```diff
@@ -49,9 +49,37 @@ set(DGAME_DCOMPONENTS_SOURCES
...
 target_precompile_headers(dComponents REUSE_FROM dGameBase)
-target_link_libraries(dComponents
-	PUBLIC dPhysics dDatabase
-	INTERFACE dUtilities dCommon dBehaviors dChatFilter dMission dInventory)
+
+# INTERFACE link w/o dependency
+#set_property(TARGET dComponents APPEND PROPERTY INTERFACE_LINK_LIBRARIES
+#	dUtilities dCommon dBehaviors dChatFilter dMission dInventory
+#	dPhysics
+#)
```
EmosewaMC, line 83: is it worth keeping this around?

dGame/dComponents/CMakeLists.txt
```diff
@@ -49,9 +49,37 @@ set(DGAME_DCOMPONENTS_SOURCES
...
+
+target_link_libraries(dComponents INTERFACE dBehaviors)
```
EmosewaMC, line 85: eof

thirdparty/mariadb-connector-cpp
EmosewaMC, line 1: Need to revert these

EmosewaMC (commented):
General formatting comments, comment questions and also there is a submodule version change that may have been unintended

CMakeLists.txt
Xiphoseer: "Makes it so it compiles libraries in series as opposed to in parallel" is what I meant

EmosewaMC:
Merge from upstream will resolve the current build failure.

CMakeLists.txt
EmosewaMC: Because any extra removals would require code changes or more work than necessary to avoid code changes.

EmosewaMC:
[mariadb changes](https://github.com/mariadb-corporation/mariadb-connector-cpp/compare/8641b1453ae3ce5a70f78248a1f7fc20a048cb88...ef0873998b3f94a4f76a485fb90b14866fbb99d4#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a)

CMakeLists.txt
```diff
@@ -202,39 +205,19 @@ foreach(file ${SQL_FILES})
 	configure_file(${CMAKE_SOURCE_DIR}/migrations/cdserver/${file} ${PROJECT_BINARY_DIR}/migrations/cdserver/${file})
 endforeach()
 
+# Add system specfic includes for Apple, Windows and Other Unix OS' (including Linux)
+if (APPLE)
+	include_directories("/usr/local/include/")
+endif()
```
EmosewaMC, line 211: Is there a particular reason this comment mentioned 3 other OS' but doesnt actually include anything specific?

dDatabase/MigrationRunner.cpp
EmosewaMC: Believe I see why this was necessary.  Looks like you needed a base to build the dDatabase library off of and none existed, so we use the MigrationRunner.

CMakeLists.txt
Xiphoseer: I think this just appears as new but was moved? In general we should not include /usr/local

aronwk-aaron (dismissed):

CMakeLists.txt
EmosewaMC: Ah yeah this is in the old version.  Not going to touch it for now, but will just comment that this likely shouldnt be done.

aronwk-aaron (approved):

jadebenn (approved):
CMake is witchcraft!

...but looks good to me.

PR #1388, opened 2024-01-04
fix: don't crash if some configs values aren't present

aronwk-aaron:
remove not starting if ini's aren't present since everything can run from env vars now

tested running a server with no ini's and the db connector game proper feedback
tested running with only the MYSQL_XXXX configs set as env vars and the server ran perfect and I was able to connect with a client

### DIFF
--- a/dAuthServer/AuthServer.cpp
+++ b/dAuthServer/AuthServer.cpp
@@ -83,12 +83,15 @@ int main(int argc, char** argv) {
 	Game::randomEngine = std::mt19937(time(0));
 
 	//It's safe to pass 'localhost' here, as the IP is only used as the external IP.
-	uint32_t maxClients = 50;
+	uint32_t maxClients = 999;
 	uint32_t ourPort = 1001; //LU client is hardcoded to use this for auth port, so I'm making it the default.
-	if (Game::config->GetValue("max_clients") != "") maxClients = std::stoi(Game::config->GetValue("max_clients"));
-	if (Game::config->GetValue("auth_server_port") != "") ourPort = std::atoi(Game::config->GetValue("auth_server_port").c_str());
+	std::string ourIP = "localhost";
+	GeneralUtils::TryParse(Game::config->GetValue("max_clients"), maxClients);
+	GeneralUtils::TryParse(Game::config->GetValue("auth_server_port"), ourPort);
+	const auto externalIPString = Game::config->GetValue("external_ip");
+	if (!externalIPString.empty()) ourIP = externalIPString;
 
-	Game::server = new dServer(Game::config->GetValue("external_ip"), ourPort, 0, maxClients, false, true, Game::logger, masterIP, masterPort, ServerType::Auth, Game::config, &Game::lastSignal);
+	Game::server = new dServer(ourIP, ourPort, 0, maxClients, false, true, Game::logger, masterIP, masterPort, ServerType::Auth, Game::config, &Game::lastSignal);
 
 	//Run it until server gets a kill message from Master:
 	auto t = std::chrono::high_resolution_clock::now();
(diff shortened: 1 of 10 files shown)


dAuthServer/AuthServer.cpp
```diff
@@ -83,12 +83,15 @@ int main(int argc, char** argv) {
...
 	//It's safe to pass 'localhost' here, as the IP is only used as the external IP.
-	uint32_t maxClients = 50;
+	uint32_t maxClients = 999;
 	uint32_t ourPort = 1001; //LU client is hardcoded to use this for auth port, so I'm making it the default.
-	if (Game::config->GetValue("max_clients") != "") maxClients = std::stoi(Game::config->GetValue("max_clients"));
-	if (Game::config->GetValue("auth_server_port") != "") ourPort = std::atoi(Game::config->GetValue("auth_server_port").c_str());
+	std::string ourIP = "localhost";
 
-	Game::server = new dServer(Game::config->GetValue("external_ip"), ourPort, 0, maxClients, false, true, Game::logger, masterIP, masterPort, ServerType::Auth, Game::config, &Game::lastSignal);
+	if (!Game::config->GetValue("max_clients").empty()) maxClients = std::stoi(Game::config->GetValue("max_clients"));
```
Xiphoseer, line 90: I know it's a oneliner this way, but can we only do the GetValue once per key?

Xiphoseer (approved):

dAuthServer/AuthServer.cpp
aronwk-aaron: Was thinking about doing that for all of these I've touched. Since they're gotten twice at worst case. 
Want me to change that and cut it down to once?

aronwk-aaron: I think I'm just gong to leave it as it unless others bring it up. I'm not sure of the impact of storing it locally vs just grabbing it every time

EmosewaMC: Storing it locally as a copy is safe to do.

aronwk-aaron: (would preferably use tryparse as we discussed, so I'll try to look into why that is causing issues more later)

aronwk-aaron: move things to use tryParse where applicable

aronwk-aaron:
retested things, they work.
Was probably doing something stupid and didn't see it

dPhysics/dpWorld.cpp
```diff
@@ -10,14 +10,17 @@
 #include "dConfig.h"
 
 void dpWorld::Initialize(unsigned int zoneID, bool generateNewNavMesh) {
-	phys_sp_tilecount = std::atoi(Game::config->GetValue("phys_sp_tilecount").c_str());
-	phys_sp_tilesize = std::atoi(Game::config->GetValue("phys_sp_tilesize").c_str());
+	const auto physSpTilecount = Game::config->GetValue("phys_sp_tilecount");
```
EmosewaMC, line 13: as just an fyi, auto cannot discern between reference and value and will always create a copy.  even if you specify const it will be a const T unless you specify const auto&

dPhysics/dpWorld.h
```diff
@@ -36,7 +36,7 @@ class dpWorld : public Singleton<dpWorld> {
 
 private:
 	dpGrid* m_Grid;
-	bool phys_spatial_partitioning = 1;
```
EmosewaMC, line 39: lol

EmosewaMC (approved):

Xiphoseer (approved):

DarwinAnim8or (approved):

PR #1398, opened 2024-01-06
feat: fully reading auth packets and use stamps

aronwk-aaron:
tested that logging in still works and that stamps in the client logs are now correct


Xiphoseer (dismissed):

EmosewaMC (dismissed):

dNet/AuthPackets.cpp
```diff
@@ -60,42 +80,93 @@ void AuthPackets::SendHandshake(dServer* server, const SystemAddress& sysAddr, c
...
+	inStream.Read(platformID);
+	LOG_DEBUG("OS Info: [Size: %i, Major: %i, Minor %i, Buid#: %i, platformID: %i]", osVersionInfoSize, majorVersion, minorVersion, buildNumber, platformID);
 
 	// Fetch account details
 	auto accountInfo = Database::Get()->GetAccountInfo(username);
 
 	if (!accountInfo) {
 		LOG("No user by name %s found!", username.c_str());
-		AuthPackets::SendLoginResponse(server, packet->systemAddress, eLoginResponse::INVALID_USER, "", "", 2001, username);
+		stamps.emplace_back(Stamp(eStamps::PASSPORT_AUTH_ERROR, 1));
```
Xiphoseer, line 152: Afaik, emplace back forwards to the constructor so you need no `Stamp(...)` here. As is, this would call the copy constructor with the new value.

aronwk-aaron: oh cool

Xiphoseer (approved):

EmosewaMC (approved):
wait for ci

PR #1399, opened 2024-01-06
chore: change timers to not use ptrs

EmosewaMC:
add comments as to why logic may seem confusing.

Tested that Spider queen battle can be completed without crashing and that schooner can also be used without error

### DIFF
--- a/dGame/Entity.cpp
+++ b/dGame/Entity.cpp
@@ -1222,39 +1222,56 @@ void Entity::UpdateXMLDoc(tinyxml2::XMLDocument* doc) {
 
 void Entity::Update(const float deltaTime) {
 	uint32_t timerPosition;
-	timerPosition = 0;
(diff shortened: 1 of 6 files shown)


Wincent01 (approved):
Lgtm. 

Considering most entities will never use timers, would it be worth moving timers to a component? The entity class has quite a few members at this point, might be worth delegating some responsibility elsewhere. This would regardless be for a future PR.

aronwk-aaron (approved):

EmosewaMC:
> Lgtm.
> 
> Considering most entities will never use timers, would it be worth moving timers to a component? The entity class has quite a few members at this point, might be worth delegating some responsibility elsewhere. This would regardless be for a future PR.

Yes.  That would be ideal to make this large class smaller, just a bigger change :)

PR #1400, opened 2024-01-06
fix: activity id not using overridden id

EmosewaMC:
Resolves an issue where starting the Nimbus Station race sends you to Gnarled Forest race.

### DIFF
--- a/dGame/dComponents/ActivityComponent.cpp
+++ b/dGame/dComponents/ActivityComponent.cpp
@@ -72,9 +72,10 @@ void ActivityComponent::LoadActivityData(const int32_t activityId) {
 	CDActivitiesTable* activitiesTable = CDClientManager::Instance().GetTable<CDActivitiesTable>();
 	std::vector<CDActivities> activities = activitiesTable->Query([activityId](CDActivities entry) {return (entry.ActivityID == activityId); });
 
+	bool soloRacing = Game::config->GetValue("solo_racing") == "1";
 	for (CDActivities activity : activities) {
 		m_ActivityInfo = activity;
-		if (static_cast<Leaderboard::Type>(activity.leaderboardType) == Leaderboard::Type::Racing && Game::config->GetValue("solo_racing") == "1") {
+		if (static_cast<Leaderboard::Type>(activity.leaderboardType) == Leaderboard::Type::Racing && soloRacing) {
 			m_ActivityInfo.minTeamSize = 1;
 			m_ActivityInfo.minTeams = 1;
 		}
--- a/dGame/dComponents/ActivityComponent.cpp
+++ b/dGame/dComponents/ActivityComponent.cpp
@@ -30,6 +30,12 @@
 #include "LeaderboardManager.h"
 
 ActivityComponent::ActivityComponent(Entity* parent, int32_t activityID) : Component(parent) {
+	/*
+	* This is precisely what the client does functionally
+	* Use the component id as the default activity id and load its data from the database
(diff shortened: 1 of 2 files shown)


dGame/dComponents/ActivityComponent.cpp
```diff
@@ -30,23 +30,11 @@
...
-			m_ActivityInfo.minTeams = 1;
-		}
-		if (m_ActivityInfo.instanceMapID == -1) {
-			const auto& transferOverride = parent->GetVarAsString(u"transferZoneID");
-			if (!transferOverride.empty()) {
-				GeneralUtils::TryParse(transferOverride, m_ActivityInfo.instanceMapID);
-			}
-		}
+	m_ActivityID = activityID;
+	LoadActivityData(activityID);
```
Xiphoseer, line 34: Is there a reason not to do this only once after the `if`?

EmosewaMC: This is precisely what the client does functionally
Use the component id as the default activity id and load its data from the database
if activityID is specified and if that column exists in the activities table, update the activity info with that data.
This is what actually resolves the NT parrot foot race issue and why DLU inserted a activity 49 into the database in the first place.  It used all of the data from 39 by default, tried to load 49, found nothing, and used 49 for the leaderboards and all the data for 39.  Will add a comment.

dGame/dComponents/ActivityComponent.cpp
```diff
@@ -74,6 +68,24 @@ ActivityComponent::ActivityComponent(Entity* parent, int32_t activityID) : Compo
 		}
 	}
 }
+void ActivityComponent::LoadActivityData(const int32_t activityId) {
+	CDActivitiesTable* activitiesTable = CDClientManager::Instance().GetTable<CDActivitiesTable>();
+	std::vector<CDActivities> activities = activitiesTable->Query([activityId](CDActivities entry) {return (entry.ActivityID == activityId); });
+
+	for (CDActivities activity : activities) {
+		m_ActivityInfo = activity;
+		if (static_cast<Leaderboard::Type>(activity.leaderboardType) == Leaderboard::Type::Racing && Game::config->GetValue("solo_racing") == "1") {
```
Xiphoseer, line 77: Actually, you may want to pull this GetValue out of the loop

Xiphoseer (dismissed):

dGame/dComponents/ActivityComponent.cpp
EmosewaMC: Any particular reason?

aronwk-aaron (dismissed):

dGame/dComponents/ActivityComponent.cpp
Xiphoseer: The env reading code that might potentially allocate and it's very unlikely to change between iterations

EmosewaMC: how are you supposed to know whether to use solo racing then?

Xiphoseer:
```
bool soloRacing = Game::config->GetValue("solo_racing") == "1";
for (...) {
  if (... && soloRacing) ...
  ...
}
```
is what I mean.

EmosewaMC: Ah that was not clear.  Will change this then.

Xiphoseer (approved):

aronwk-aaron (approved):

PR #1401, opened 2024-01-06
chore: make dNet only depend on dCommon libraries

EmosewaMC:
- Add structs to communicate between the various functions so the data can be easily passed around to the correct handler
- Remove code that was dependent on other libraries linking against dNet by using the variables passed into the function instead, inlining the code to the one place it is used (A new issue should be created tracking this code so we can find a proper place for it later) and in one case inlining without a TODO or a struct due to the amount of data needed to be passed between the interfaces, and the fact that Character is a dGame class with multiple serialization types.

Tested that
- I can still load into worlds
- See other player positons update with and without mounts or racecars
- save data is in tact when delivered to my client
- bad words are rejected from civilians and gms can still bypass in global and friend chat
- remove cheat detection that isnt a cheat detection and doesn't really do anything anyways
- help issues are still the same as the strings in the inis

### DIFF
--- a/dGame/UserManager.cpp
+++ b/dGame/UserManager.cpp
@@ -216,7 +216,52 @@ void UserManager::RequestCharacterList(const SystemAddress& sysAddr) {
 		chars.push_back(character);
 	}
 
-	WorldPackets::SendCharacterList(sysAddr, u);
+	RakNet::BitStream bitStream;
+	BitStreamUtils::WriteHeader(bitStream, eConnectionType::CLIENT, eClientMessageType::CHARACTER_LIST_RESPONSE);
+
+	std::vector<Character*> characters = u->GetCharacters();
+	bitStream.Write<uint8_t>(characters.size());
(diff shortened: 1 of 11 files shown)


dGame/UserManager.cpp
```diff
@@ -216,7 +216,52 @@ void UserManager::RequestCharacterList(const SystemAddress& sysAddr) {
 		chars.push_back(character);
 	}
 
-	WorldPackets::SendCharacterList(sysAddr, u);
```
Xiphoseer, line 219: Not sure how I feel about moving all of this vs putting (most of?) Character.h in dCommon judging by its headers; and passing just the reference to the vector to dNet here. But that can be a follow up.

Xiphoseer (approved):
This is great! Approved with one optional suggestion

dGame/UserManager.cpp
EmosewaMC: will add that to the resulting issue tracking the other created work from this change.

aronwk-aaron (approved):

PR #1305, opened 2023-11-19
FIX: Make pets behave more accurately to Live

jadebenn:
Working on making the pets behave more like they did in live. Want to fix #536, #984, and #1241. As of this draft's posting, I have very roughly managed to get the treasure dig behavior prompt to appear as in Live. Still need to work on cleaning up the code and getting the functionality nailed-down better.

https://github.com/DarkflameUniverse/DarkflameServer/assets/9892985/4831f1c8-0cb4-4637-b810-faaef46e4789

PR #1402, opened 2024-01-06
chore: Assorted pet improvements

jadebenn:
Porting over some of the changes from #1305 that are currently ready to go. This PR mainly deals with loading values from CDClient regarding pet traits.


dDatabase/CDClientDatabase/CDClientTables/CDPetComponentTable.cpp
```diff
@@ -0,0 +1,63 @@
+#include "CDPetComponentTable.h"
+
+namespace {
+	// Default entries for fallback
+	CDPetComponent defaultEntry {
+		.id = static_cast<unsigned int>(-1),
```
EmosewaMC, line 6: Note to myself or aron to check what the defaults should be in the client

dDatabase/CDClientDatabase/CDClientTables/CDPetComponentTable.cpp
```diff
@@ -0,0 +1,63 @@
...
+		entry.runSpeed = tableData.getFloatField("runSpeed", defaultEntry.runSpeed);
+		entry.sprintSpeed = tableData.getFloatField("sprintSpeed", defaultEntry.sprintSpeed);
+		UNUSED_COLUMN(entry.idleTimeMin = tableData.getFloatField("idleTimeMin", defaultEntry.idleTimeMin);)
+		UNUSED_COLUMN(entry.idleTimeMax = tableData.getFloatField("idleTimeMax", defaultEntry.idleTimeMax);)
+		UNUSED_COLUMN(entry.petForm = tableData.getIntField("petForm", 0);)
+		entry.imaginationDrainRate = tableData.getFloatField("imaginationDrainRate", defaultEntry.imaginationDrainRate);
+		UNUSED_COLUMN(entry.AudioMetaEventSet = tableData.getStringField("AudioMetaEventSet", defaultEntry.AudioMetaEventSet);)
+		UNUSED_COLUMN(entry.buffIDs = tableData.getStringField("buffIDs", defaultEntry.buffIDs);)
+
+		m_entries.insert(std::make_pair(entry.id, entry));
```
EmosewaMC, line 44: Consider re-working this to instead use operator subscripting at the start of the loop to get a ref to a default constructed element and then setting the values there so no copy needs to be made

dDatabase/CDClientDatabase/CDClientTables/CDPetComponentTable.cpp
```diff
@@ -0,0 +1,63 @@
...
+		tableData.nextRow();
+	}
+
+	tableData.finalize();
+
```
EmosewaMC, line 49: Isn't necessary to call finalize manually. The result set automatically calls this in a try catch(...) via raii

EmosewaMC (commented):
General feedback

dDatabase/CDClientDatabase/CDClientDatabase.h
```diff
@@ -15,6 +15,10 @@
 
 //! The CDClient Database namespace
 namespace CDClientDatabase {
+	/**
+	 * Boolean defining the connection status of CDClient
+	*/
+	extern bool isConnected;
```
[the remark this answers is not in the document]
jadebenn: Has to be extern or a separate boolean variable is declared in every translation unit.

dDatabase/CDClientDatabase/CDClientTables/CDPetComponentTable.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+void CDPetComponentTable::LoadValuesFromDefaults() {
+	m_entries.insert(std::make_pair(defaultEntry.id, defaultEntry));
+}
+
+CDPetComponent& CDPetComponentTable::GetByID(unsigned int componentID) {
+	auto itr = m_entries.find(componentID);
+	if (itr == m_entries.end()) {
+		LOG("Unable to load pet component (ID %i) values from database! Using default values instead.", componentID);
```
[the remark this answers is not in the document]
jadebenn: It doesn't normally appear during gameplay, so something has gone very wrong if you see it at all.

dGame/dComponents/PetComponent.cpp
```diff
@@ -69,7 +69,8 @@ std::map<LOT, int32_t> PetComponent::petFlags = {
 		{ 13067, 838 }, // Skeleton dragon
 };
 
-PetComponent::PetComponent(Entity* parent, uint32_t componentId): Component(parent) {
+PetComponent::PetComponent(Entity* parentEntity, uint32_t componentId) : Component{ parentEntity },
+	m_PetInfo{ CDClientManager::Instance().GetTable<CDPetComponentTable>()->GetByID(componentId) } {
```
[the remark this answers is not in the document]
jadebenn: It can't be. The notation doesn't allow it as the reference is considered initialized without a value.

dDatabase/CDClientDatabase/CDClientTables/CDPetComponentTable.cpp
EmosewaMC: defaults here are fine.  Use 0 for the default id or change the id type to int instead of uint (-1 cast to uint is weird)

dDatabase/CDClientDatabase/CDClientManager.cpp
```diff
@@ -52,43 +52,50 @@
...
+		CDRewardsTable::Instance().LoadValuesFromDatabase();
+		CDScriptComponentTable::Instance().LoadValuesFromDatabase();
+		CDSkillBehaviorTable::Instance().LoadValuesFromDatabase();
+		CDVendorComponentTable::Instance().LoadValuesFromDatabase();
+		CDZoneTableTable::Instance().LoadValuesFromDatabase();
+	} else {
+		LOG("Unable to connect to CDClientDatabase! Using default values instead!");
+
+		CDPetComponentTable::Instance().LoadValuesFromDefaults();
+	}
```
Wincent01, line 100: Wouldn't it be catastrophic if the server failed to connect to the cdclient; feels like this should throw. It would regardless be more readable to invert the if here and get rid of the else clause with an early return.

dDatabase/CDClientDatabase/CDClientTables/CDPetComponentTable.h
```diff
@@ -0,0 +1,49 @@
...
+	UNUSED_COLUMN(unsigned int elementType;)
+	float walkSpeed;
+	float runSpeed;
+	float sprintSpeed;
+	UNUSED_COLUMN(float idleTimeMin;)
+	UNUSED_COLUMN(float idleTimeMax;)
+	UNUSED_COLUMN(unsigned int petForm;)
+	float imaginationDrainRate;
+	UNUSED_COLUMN(std::string AudioMetaEventSet;)
+	UNUSED_COLUMN(std::string buffIDs;)
```
Wincent01, line 20: Can change all unsigned ints to uint32_t.

dGame/dComponents/PetComponent.h
```diff
@@ -357,7 +352,9 @@ class PetComponent : public Component
 	PreconditionExpression* m_Preconditions;
 
 	/**
-	 * The rate at which imagination is drained from the user for having the pet out.
+	 * Pet information loaded from the CDClientDatabase
 	 */
-	float imaginationDrainRate;
+	CDPetComponent& m_PetInfo;
```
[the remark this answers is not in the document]
Wincent01: Query on if there should be such a guarantee. Feels unnecessary to deep copy every time you want to keep information from the cdclient elsewhere.

Wincent01 (changes requested):
The only blocking thing here is the question about the cdclient constructor.

dGame/dComponents/PetComponent.h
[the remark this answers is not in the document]
EmosewaMC: the cdclient database values are destructed before the server starts handling shutdown so this will cause undefined behavior keeping a reference to a destructed element.  Keeping a reference is a bad pattern and is what caused save data to be corrupted for end users when the same thing was done for mission data.

dGame/dComponents/PetComponent.h
```diff
@@ -103,7 +98,7 @@ class PetComponent : public Component
 	 * @param typeId extra information about the command, e.g. the emote to play
 	 * @param overrideObey unused
 	 */
-	void Command(NiPoint3 position, LWOOBJID source, int32_t commandType, int32_t typeId, bool overrideObey);
+	void Command(const NiPoint3& position, const LWOOBJID source, const int32_t commandType, const int32_t typeId, const bool overrideObey);
```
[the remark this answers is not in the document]
EmosewaMC: I believe it is still good to declare a variable const so that you cant accidentally modify it in your function body.  It does prevent you from accidentally changing the value mindlessly

dGame/dComponents/PetComponent.h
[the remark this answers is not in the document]
EmosewaMC: for references in a map, sure its fine to keep one around, but thats only if you can guarantee the reference will stay alive, and in the case of the current cdclient implementation this is false.  see #1183 for the details on how cdclient tables are destructed.

dDatabase/CDClientDatabase/CDClientManager.cpp
EmosewaMC: I believe it is like this for the testing, though I agree it does flow weird and doesn't make sense to continue execution if there is no connection.  Is there possibly a different way to communicate the same result jade?

dGame/dComponents/PetComponent.cpp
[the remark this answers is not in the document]
EmosewaMC: This is a further reason to not store a reference

dGame/dComponents/PetComponent.h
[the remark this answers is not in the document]
EmosewaMC: I should make it clear, I am all _for_ storing the data as a reference, but I already know this is not safe currently.  If the data in the cdclient table was indeed safe to reference, I would have no problem doing so.

dDatabase/CDClientDatabase/CDClientManager.cpp
jadebenn: It is like this for the testing. I'll see if I can figure out a better way to handle this. Perhaps I can make some sort of "testing mode" flag and only have it load the fallbacks if that is set to true. It would be useful to have a mock CDClient database for our unit testing without actually connecting to it.

dGame/dComponents/PetComponent.h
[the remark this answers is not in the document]
jadebenn: > I should make it clear, I am all _for_ storing the data as a reference, but I already know this is not safe currently. If the data in the cdclient table was indeed safe to reference, I would have no problem doing so.

Is it worth considering providing such a guarantee, then?

dGame/dComponents/PetComponent.h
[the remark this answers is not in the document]
jadebenn: If you're consistent with it you can also tell when a function is going to modify an input because said input will not be declared const

dDatabase/CDClientDatabase/CDClientManager.cpp
Xiphoseer: The "standard" way to do this might be to have one constructor that takes a filename and loads the DB from SQLite and one without parameters that just creates an empty instance. Since these are all singletons that depend on ordering anyway, it may make sense to move the loading out of the Manager constructor and into the CDClientDatabase::Connect method or something that is called closer to it.

Right now, if someone were to call CDClientManager::Instance _before_ connect, things would just break (Null pointer dereference [here](https://github.com/DarkflameUniverse/DarkflameServer/blob/e4469f997ec10c63e9e1c3972542ae6b0cfdfe94/dDatabase/CDClientDatabase/CDClientDatabase.cpp#L14) would be my guess, which may leave the static in CDClientManager::Instance in a broken state even if we were to recover from the null ptr dereference.)

dDatabase/CDClientDatabase/CDClientTables/CDPetComponentTable.cpp
jadebenn: I've tried my hand at an implementation using "try_emplace." I'll push it in a bit. Please take a look and let me know if it's what you meant.

dGame/dComponents/PetComponent.h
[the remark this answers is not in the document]
EmosewaMC: Yes it is, but we would need to research how to extend the lifetime of the orm containers to rhe binary life instead of until a signal is handled

dDatabase/CDClientDatabase/CDClientManager.cpp
jadebenn: I ended up doing something similar to that. Works a lot nicer than the old solution, IMO.

CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ set(RECASTNAVIGATION_EXAMPLES OFF CACHE BOOL "" FORCE)
 # Disabled no-register
 # Disabled unknown pragmas because Linux doesn't understand Windows pragmas.
 if(UNIX)
-	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++17 -O2 -Wuninitialized -fPIC")
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++20 -O2 -Wuninitialized -fPIC")
```
EmosewaMC, line 54: the changed flag here can just be removed.  We already set the std above this.

EmosewaMC (dismissed):

CMakeLists.txt
jadebenn: Removed

Wincent01 (approved):
Lgtm

CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ set(RECASTNAVIGATION_EXAMPLES OFF CACHE BOOL "" FORCE)
 # Disabled no-register
 # Disabled unknown pragmas because Linux doesn't understand Windows pragmas.
 if(UNIX)
-	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++17 -O2 -Wuninitialized -fPIC")
```
aronwk-aaron, line 54: just curious as to why this was changed, I thought that we were already on c++20

aronwk-aaron (approved):
Going to let @EmosewaMC give it a final once over and merge

CMakeLists.txt
EmosewaMC: It's an extraneous flag.  We already set the version above this with cxx standard.

EmosewaMC (approved):

PR #1406, opened 2024-01-10
refactor: Reduce/remove entity pointers as means of addressing entities

jadebenn:
Working toward an object ID system instead. Entity pointers should not be assumed to be constant, as it would be useful to have the capability to move their locations in memory.


jadebenn:
There is no nice way to say it: This is going to be a _major_ PITA when it comes to the function calls themselves.

dGame/EntityManager.h
```diff
@@ -20,6 +20,13 @@ class EntityManager {
 public:
 	void Initialize();
 
+	/**
+	 * Add entity to the entity lookup map
+	 * @param objectId Object ID
+	 * @param entityPtr Entity pointer 
+	*/
+	void AddEntity(const LWOOBJID objectId, Entity* entityPtr) { m_Entities.insert_or_assign(objectId, entityPtr); };
```
EmosewaMC, line 28: im not sure how much of a fan I am of exposing this API as it implies anyone and thing can add an entity when in reality every single entity must go through create entity first.

EmosewaMC (commented):
As a request it would be preferred to see a perf impact of before and after the change in real world and stress scenarios

dGame/dComponents/Component.h
```diff
@@ -50,5 +50,5 @@ class Component
 	/**
 	 * The entity that owns this component
 	 */
-	Entity* m_Parent;
```
EmosewaMC, line 53: Why is this being changed?  I had thought the agreement was to move this to a ref instead of an objectID

dGame/dGameMessages/GameMessages.cpp
```diff
@@ -659,19 +664,20 @@ void GameMessages::SendPlayEmbeddedEffectOnAllClientsNearObject(Entity* entity,
 	SEND_PACKET_BROADCAST;
 }
 
+// Compatibility shim
 void GameMessages::SendPlayFXEffect(Entity* entity, int32_t effectID, const std::u16string& effectType, const std::string& name, LWOOBJID secondary, float priority, float scale, bool serialize) {
 	SendPlayFXEffect(entity->GetObjectID(), effectID, effectType, name, secondary, priority, scale, serialize);
 }
 
-void GameMessages::SendPlayFXEffect(const LWOOBJID& entity, int32_t effectID, const std::u16string& effectType, const std::string& name, LWOOBJID secondary, float priority, float scale, bool serialize) {
+void GameMessages::SendPlayFXEffect(const LWOOBJID entityId, const int32_t effectId, const std::u16string& effectType, const std::string& name, const LWOOBJID secondary, const float priority, const float scale, const bool serialize) {
```
EmosewaMC, line 672: LWOOBJID should be passed by ref as it generally doesn't fit into a register

EmosewaMC:
There are a lot of miscelaneous changes in this PR afte doing a quick scan of it, could this be split into some more focussed ones instead to make the review process easier?

jadebenn:
I might revisit this, but if so, it probably won't be this branch. Closing.

PR #1408, opened 2024-01-12
chore: Simplify and move Player functionality to relevant component

EmosewaMC:
Already touched a few things here, so i'm stopping here for the next PR.

Detailed changes:
- Move code to CharacterComponent
- Remove extraneous interfaces
- Simplify some code greatly
- Change some types to return and take in const ref (only structs larger than 8 bytes benefit from this change.)
- Update code to use CharacterComponent for sending to zone instead of Player*.
- Remove static storage container (static containers can be destroyed before exit/terminate handler executes)

Tested that I am still able to leave cannon cove, battle of nimbus station and avant gardens survival.  Am also still able to receive mail when my inventory is full.

### DIFF
--- a/dGame/Entity.cpp
+++ b/dGame/Entity.cpp
@@ -1879,7 +1879,7 @@ const NiQuaternion& Entity::GetRotation() const {
 	return NiQuaternion::IDENTITY;
 }
 
-void Entity::SetPosition(NiPoint3 position) {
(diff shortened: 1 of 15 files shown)


aronwk-aaron (approved):
nice!

Xiphoseer (approved):

PR #1414, opened 2024-01-13
refactor: allow usage of NiPoint3 and NiQuaternion in constexpr context

jadebenn:
Allows constexpr evaluation of the NiPoint3 and NiQuaternion types, as well as all but one of their member functions (the sole exception was causing a nasty circular dependency issue and probably isn't very useful in a constexpr context anyway).


dCommon/NiPoint3.h
```diff
@@ -50,111 +62,169 @@ class NiPoint3 {
...
+		float* base = &x;
+		return base[i];
+	}
 
 	//! Operator for subscripting
-	const float& operator[](int i) const;
+	constexpr const float& operator[](const int i) const noexcept { // TODO: Is this redundant?
+		const float* base = &x;
+		return base[i];
+	}
```
EmosewaMC, line 179: yes, it is redundant.  Nothing uses this

EmosewaMC (dismissed):
I have no reason to not be in favor of this.  you convinced me otherwise.

dCommon/NiPoint3.h
jadebenn: It is part of the original code though. Maybe there's a reason for it? I'm hesitant to remove it unless there's a pressing need to.

EmosewaMC: This code is just copied I can almost 100% tell you that and ill show you why in discord

dCommon/NiPoint3.inl
```diff
@@ -0,0 +1,37 @@
+#pragma once
```
EmosewaMC, line 1: pragma once not needed for inl files, if anything the inverse is necessary
```
#ifndef NiPoint3HeaderGuard
#error "This is only to be included inline in NiPoint3.h.  Do not include directly"
#endif
```

EmosewaMC: Would probably want a combo of the two if you wanted to make sure the inl is also only included once

jadebenn: Good idea. I'll implement that in a bit.

aronwk-aaron (approved):

aronwk-aaron (dismissed):

aronwk-aaron (dismissed):

Xiphoseer (dismissed):
Don't quite get why the inl is necessary, otherwise lgtm

EmosewaMC:
> Don't quite get why the inl is necessary, otherwise lgtm

Helps separate out definitions from declarations imo, I would do the same for Entity's template definitions, except in a .tpp file instead of right in the header.

EmosewaMC (dismissed):
lgtm pending passing CI.

Xiphoseer (dismissed):

EmosewaMC:
The static vars need to be compiled once in a cpp file somewhere since they are static and not inline.

Jettford (approved):

.github/workflows/build-and-test.yml
```diff
@@ -2,9 +2,9 @@ name: CI
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, ConstexprNiTypes ]
```
EmosewaMC: remove

dCommon/NiPoint3.h
```diff
@@ -12,149 +13,144 @@ typedef NiPoint3 Vector3;       //!< The Vector3 class is technically the NiPoin
...
-	static const NiPoint3 ZERO;         //!< Point(0, 0, 0)
-	static const NiPoint3 UNIT_X;       //!< Point(1, 0, 0)
-	static const NiPoint3 UNIT_Y;       //!< Point(0, 1, 0)
-	static const NiPoint3 UNIT_Z;       //!< Point(0, 0, 1)
-	static const NiPoint3 UNIT_ALL;     //!< Point(1, 1, 1)
+	constexpr NiPoint3(const NiPoint3& point) noexcept
+		: x{ point.x }
+		, y{ point.y }
+		, z{ point.z } {
+	}
```
EmosewaMC, line 44: Why do you use this syntax for PODs?  To me it adds more clutter and is more cumbersome than just doing this in the constructor body, and since these are POD the gain from doing a chained construction like this is lost since there is no default constructor to not call first.

jadebenn: It's the syntax I'm more used to at this point, but it doesn't make a difference for basic data like floats, yeah.

jadebenn: Also, to restate what I said in the discord, you also get compiler warnings for certain conversions when you do it this way. Not really an issue here, but part of the reason I've gotten into the habit of it.

EmosewaMC: This way being chained constructors or in body construction?

jadebenn: Constructors utilizing member initialization lists in general.

jadebenn:
Can I get some new reviews on this when you all have a chance? It's been ready for a while.

aronwk-aaron (approved):

Jettford (approved):

PR #1415, opened 2024-01-13
chore: remove all raw packet reading from chat packet handler 

aronwk-aaron:
and general chat packet cleanup

Tested that things still work

### DIFF
--- a/dChatServer/ChatPacketHandler.cpp
+++ b/dChatServer/ChatPacketHandler.cpp
@@ -80,31 +80,33 @@ void ChatPacketHandler::HandleFriendRequest(Packet* packet) {
 	CINSTREAM_SKIP_HEADER;
 
 	LWOOBJID requestorPlayerID;
-	LUWString playerName;
+	LUWString LUplayerName;
 	char isBestFriendRequest{};
 
 	inStream.Read(requestorPlayerID);
 	inStream.IgnoreBytes(4);
-	inStream.Read(playerName);
+	inStream.Read(LUplayerName);
 	inStream.Read(isBestFriendRequest);
+
+	auto playerName = LUplayerName.GetAsString();
 
 	auto& requestor = Game::playerContainer.GetPlayerDataMutable(requestorPlayerID);
 	if (!requestor) {
-		LOG("No requestor player %llu sent to %s found.", requestorPlayerID, playerName.GetAsString().c_str());
+		LOG("No requestor player %llu sent to %s found.", requestorPlayerID, playerName.c_str());
 		return;
 	}
 
-	if (requestor.playerName == playerName.GetAsString()) {
+	if (requestor.playerName == playerName) {
 		SendFriendResponse(requestor, requestor, eAddFriendResponseType::MYTHRAN);
 		return;
 	};
 
-	auto& requestee = Game::playerContainer.GetPlayerDataMutable(playerName.GetAsString());
+	auto& requestee = Game::playerContainer.GetPlayerDataMutable(playerName);
 
 	// Check if player is online first
 	if (isBestFriendRequest && !requestee) {
 		for (auto& friendDataCandidate : requestor.friends) {
-			if (friendDataCandidate.friendName != playerName.GetAsString()) continue;
+			if (friendDataCandidate.friendName != playerName) continue;
 			// Setup the needed info since you can add a best friend offline.
 			requestee.playerID = friendDataCandidate.friendID;
 			requestee.playerName = friendDataCandidate.friendName;
@@ -125,8 +127,8 @@ void ChatPacketHandler::HandleFriendRequest(Packet* packet) {
 	// If at this point we dont have a target, then they arent online and we cant send the request.
 	// Send the response code that corresponds to what the error is.
 	if (!requestee) {
-		requestee.playerName = playerName.GetAsString();
-		auto responseType = Database::Get()->GetCharacterInfo(playerName.GetAsString())
+		requestee.playerName = playerName;
+		auto responseType = Database::Get()->GetCharacterInfo(playerName)
 			? eAddFriendResponseType::NOTONLINE
 			: eAddFriendResponseType::INVALIDCHARACTER;
 
@@ -285,15 +287,16 @@ void ChatPacketHandler::HandleFriendResponse(Packet* packet) {
 void ChatPacketHandler::HandleRemoveFriend(Packet* packet) {
 	CINSTREAM_SKIP_HEADER;
 	LWOOBJID playerID;
-	LUWString friendName;
+	LUWString LUFriendName;
 	inStream.Read(playerID);
 	inStream.IgnoreBytes(4);
-	inStream.Read(friendName);
+	inStream.Read(LUFriendName);
+	auto friendName = LUFriendName.GetAsString();
 
 	//we'll have to query the db here to find the user, since you can delete them while they're offline.
 	//First, we need to find their ID:
 	LWOOBJID friendID = 0;
-	auto friendIdResult = Database::Get()->GetCharacterInfo(friendName.GetAsString());
+	auto friendIdResult = Database::Get()->GetCharacterInfo(friendName);
 	if (friendIdResult) {
 		friendID = friendIdResult->id;
 	}
@@ -315,8 +318,7 @@ void ChatPacketHandler::HandleRemoveFriend(Packet* packet) {
 				break;
 			}
 		}
-		auto name = friendName.GetAsString();
-		SendRemoveFriend(goonA, name, true);
+		SendRemoveFriend(goonA, friendName, true);
 	}
 
 	auto& goonB = Game::playerContainer.GetPlayerDataMutable(friendID);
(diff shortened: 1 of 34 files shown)


Xiphoseer (dismissed):

Xiphoseer (approved):

dChatServer/ChatPacketHandler.cpp
```diff
@@ -132,8 +125,8 @@ void ChatPacketHandler::HandleFriendRequest(Packet* packet) {
 	// If at this point we dont have a target, then they arent online and we cant send the request.
 	// Send the response code that corresponds to what the error is.
 	if (!requestee) {
-		requestee.playerName = playerName;
-		auto responseType = Database::Get()->GetCharacterInfo(playerName)
+		requestee.playerName = playerName.GetAsString();
```
EmosewaMC, line 128: There is a lot of copying of the LUWString.GetAsString going on here.  Consider a `const auto nameAsStr = playerName.GetAsString();`

dChatServer/ChatPacketHandler.cpp
```diff
@@ -288,13 +285,15 @@ void ChatPacketHandler::HandleFriendResponse(Packet* packet) {
...
 	inStream.Read(playerID);
-	std::string friendName = PacketUtils::ReadString(0x14, packet, true);
+	inStream.IgnoreBytes(4);
+	inStream.Read(friendName);
 
 	//we'll have to query the db here to find the user, since you can delete them while they're offline.
 	//First, we need to find their ID:
 	LWOOBJID friendID = 0;
-	auto friendIdResult = Database::Get()->GetCharacterInfo(friendName);
+	auto friendIdResult = Database::Get()->GetCharacterInfo(friendName.GetAsString());
```
EmosewaMC, line 296: Same copy twice happens here, again consider a const declaration and just copy it once.

dChatServer/ChatPacketHandler.cpp
```diff
@@ -337,121 +337,142 @@ void ChatPacketHandler::HandleRemoveFriend(Packet* packet) {
...
-	inStream.SetReadOffset(0x14 * 8);
-
-	uint8_t channel = 0;
+	eChatChannel channel;
+	uint32_t size;
+	
+	inStream.IgnoreBytes(4);
 	inStream.Read(channel);
+	inStream.Read(size);
+	inStream.IgnoreBytes(77);
```
EmosewaMC, line 352: Why 77?  This should be commented and elaborated on for future maintainers otherwise we'll RE this again.

dChatServer/ChatPacketHandler.cpp
```diff
@@ -337,121 +337,142 @@ void ChatPacketHandler::HandleRemoveFriend(Packet* packet) {
...
-	std::string message = PacketUtils::ReadString(0x66, packet, true, 512);
-
-	LOG("Got a message from (%s) [%d]: %s", sender.playerName.c_str(), channel, message.c_str());
+	LUWString message(size);
+	inStream.Read(message);
+	
+	LOG("Got a message from (%s) via [%s]: %s", sender.playerName.c_str(), StringifiedEnum::ToString(channel).data(), message.GetAsString().c_str());
 
-	if (channel != 8) return;
+	if (channel == eChatChannel::TEAM) {
```
EmosewaMC, line 359: Why was this inverted?  Bouncer was good for early return before to reduce scope.  If its to allow for more channels, consider a switch case instead for future use intent

dChatServer/ChatPacketHandler.cpp
```diff
@@ -337,121 +337,142 @@ void ChatPacketHandler::HandleRemoveFriend(Packet* packet) {
...
-		SystemAddress sysAddr = otherMember.sysAddr;
-		SEND_PACKET;
-	}
-}
+	inStream.IgnoreBytes(4);
+	inStream.Read(channel);
+	inStream.Read(size);
+	inStream.IgnoreBytes(77);
+	inStream.Read(receiverName);
+	inStream.IgnoreBytes(2);
```
EmosewaMC, line 408: Same thing here with the magic numbers.  Try to elaborate on them if possible.

dChatServer/ChatPacketHandler.cpp
```diff
@@ -337,121 +337,142 @@ void ChatPacketHandler::HandleRemoveFriend(Packet* packet) {
...
 
-	//Get the bois:
-	const auto& goonA = Game::playerContainer.GetPlayerData(senderID);
-	const auto& goonB = Game::playerContainer.GetPlayerData(receiverName);
-	if (!goonA || !goonB) return;
+	const auto& receiver = Game::playerContainer.GetPlayerData(receiverName.GetAsString());
+	if (!receiver) return; // TODO: respond with not online or error
 
-	if (goonA.GetIsMuted()) return;
+	// TODO: check if they are friends, only friends should be able to whisper/private chat
```
EmosewaMC, line 418: container has this data already.  Why was the muted check removed?

dChatServer/ChatServer.cpp
```diff
@@ -295,7 +297,7 @@ void HandlePacket(Packet* packet) {
 			break;
 
 		default:
-			LOG("Unknown CHAT id: %i", int(packet->data[3]));
+			LOG("Unknown/Unhandled CHAT Message id: %s (%i)", StringifiedEnum::ToString(chat_message_type).data(), chat_message_type);
```
EmosewaMC, line 300: nit: add the explicitly unhandled ones as a fall through case statement such that if we plan to handle them, it logs Unhandled.

dChatServer/PlayerContainer.h
```diff
@@ -42,6 +42,8 @@ struct PlayerData {
 	std::string playerName;
 	std::vector<FriendData> friends;
 	std::vector<IgnoreData> ignoredPlayers;
+	uint8_t GMLevel = 0;
```
EmosewaMC, line 45: Why is eGameMasterLevel not used here?  In any case this is unused in this PR and should be documenated as a TODO to implement

dChatServer/ChatPacketHandler.cpp
EmosewaMC: > Same thing here with the magic numbers. Try to elaborate on them if possible.

If this code is also repeated, consider a function to deserialize and a struct to propagate the data between functions

dChatServer/ChatPacketHandler.cpp
aronwk-aaron: yeah, true

dChatServer/ChatPacketHandler.cpp
aronwk-aaron: there's no data there, so I'll make a comment to indicate as such

dChatServer/ChatPacketHandler.cpp
aronwk-aaron: yeah, there can be more channels supported, so a switch case would work well here

dChatServer/ChatPacketHandler.cpp
aronwk-aaron: yeah, was thinking about breaking it out and how to do that

dChatServer/ChatPacketHandler.cpp
aronwk-aaron: mute check is done above, was a redundant check (will double check tho)

dChatServer/PlayerContainer.h
aronwk-aaron: noted, was like 3am and i forgot we had an emun for it lol

dChatServer/ChatPacketHandler.cpp
aronwk-aaron: Also just didn't do it yet cause I wanted to keep this PR from getting to big, but can do

dChatServer/ChatPacketHandler.cpp
aronwk-aaron: erh, not the reading, but the writing.
the reading is just slightly different enough to be really annoying to consolidate to a common place like that

dChatServer/ChatPacketHandler.cpp
aronwk-aaron: yeah, goona is sender and the sender is check if muted above, so redundant check

dChatServer/ChatPacketHandler.cpp
aronwk-aaron: will consolidate the writing in a future PR

dChatServer/ChatPacketHandler.cpp
EmosewaMC: the friends check isnt even much more code
```
bool isFriends = std::find(friends.begin(), friends.end(), friendId) != friends.end();
if (friends) {}
```

aronwk-aaron: yeah, I'm just doing it right this moment lolol

dChatServer/ChatPacketHandler.cpp
```diff
@@ -346,45 +348,31 @@ void ChatPacketHandler::HandleChatMessage(Packet* packet) {
 	eChatChannel channel;
 	uint32_t size;
 	
-	inStream.IgnoreBytes(4);
+	inStream.IgnoreBytes(4); // padding/garbage
 	inStream.Read(channel);
 	inStream.Read(size);
-	inStream.IgnoreBytes(77);
+	inStream.IgnoreBytes(77); // padding/garbage
```
EmosewaMC, line 354: I would consider "the structure the client uses to send this packet is shared in many chat messages that are sent to the server.  Because of this, there is a large gap of unused data here" or something along those lines just because this is an abnormally large chunk of data to just call "padding"

dChatServer/ChatPacketHandler.cpp
```diff
@@ -398,72 +386,83 @@ void ChatPacketHandler::HandlePrivateChatMessage(Packet* packet) {
...
 
-	//To the sender:
-	{
-		CBITSTREAM;
-		BitStreamUtils::WriteHeader(bitStream, eConnectionType::CHAT_INTERNAL, eChatInternalMessageType::ROUTE_TO_PLAYER);
-		bitStream.Write(sender.playerID);
+	if (sender.isFTP) {
+		SendPrivateChatMessage(sender, receiver, sender, message, eChatMessageResponseCode::SENDERFREETRIAL);
+		return;
+	}
```
EmosewaMC, line 422: I believe any logic actually involving FTP should be removed.  We shouldn't even have work setup to support it in dlu.

dChatServer/ChatPacketHandler.cpp
```diff
@@ -398,72 +386,83 @@ void ChatPacketHandler::HandlePrivateChatMessage(Packet* packet) {
...
-		bitStream.Write(sender.playerID);
-		bitStream.Write<uint16_t>(0); // sourceID
-		bitStream.Write(sender.GMLevel);
-		bitStream.Write(LUWString(receiver.playerName));
-		bitStream.Write(receiver.GMLevel);
-		bitStream.Write(eChatMessageResponseCode::ReceivedNewWhisper);
-		bitStream.Write(message);
+	BitStreamUtils::WriteHeader(bitStream, eConnectionType::CHAT, eChatMessageType::PRIVATE_CHAT_MESSAGE);
+	bitStream.Write(sender.playerID);
+	bitStream.Write(eChatChannel::TEAM);
```
EmosewaMC, line 450: This function is called SendPrivateChatmessage, but the chat channel is Team.  Why was this changed from what it was previously?

dChatServer/PlayerContainer.h
```diff
@@ -42,7 +43,8 @@ struct PlayerData {
 	std::string playerName;
 	std::vector<FriendData> friends;
 	std::vector<IgnoreData> ignoredPlayers;
-	uint8_t GMLevel = 0;
+	// Unused TODO: make use of this
+	eGameMasterLevel GMLevel = eGameMasterLevel::CIVILIAN;
```
EmosewaMC, line 47: I would remove the include, forward declare and move the initializer to be defined in PlayerContainer.cpp to reduce the include clutter

dChatServer/ChatPacketHandler.cpp
```diff
@@ -398,72 +386,83 @@ void ChatPacketHandler::HandlePrivateChatMessage(Packet* packet) {
...
-		bitStream.Write<uint16_t>(0); // sourceID
-		bitStream.Write(sender.GMLevel);
-		bitStream.Write(LUWString(receiver.playerName));
-		bitStream.Write(receiver.GMLevel);
-		bitStream.Write(eChatMessageResponseCode::Sent);
-		bitStream.Write(message);
+	if (receiver.isFTP) {
+		SendPrivateChatMessage(sender, receiver, sender, message, eChatMessageResponseCode::RECEIVERFREETRIAL);
+		return;
+	}
```
EmosewaMC, line 427: Same thing here.  I dont like the prospect of actually starting to add code to support FTP features and locks.

dChatServer/ChatPacketHandler.cpp
aronwk-aaron: yep, it should take in the channel to send on as well. oversight on my part
teams is technically a type of private chat message as well

dChatServer/ChatPacketHandler.cpp
aronwk-aaron: fair, got carried away covering every case

EmosewaMC (approved):

EmosewaMC (approved):

Xiphoseer (approved):

PR #1416, opened 2024-01-13
fix: Remove pending timer logic

EmosewaMC:
Updated timers to use a clever erase such that we swap this timer with the last one if there are 2 and we arent deleting the last timer.

Fixes an issue where mechs in Avant Gardens Survival wouldn't attack you and probably more.

### DIFF
--- a/dGame/Entity.cpp
+++ b/dGame/Entity.cpp
@@ -1266,17 +1266,6 @@ void Entity::Update(const float deltaTime) {
 		}
 	}
 
-	// Add pending timers to the list of timers so they start next tick.
-	if (!m_PendingTimers.empty()) {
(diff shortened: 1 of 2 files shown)


aronwk-aaron (approved):

Xiphoseer (approved):
This seems logically sound to me, but from a performance perspecive it's weird to push all timers and then move up the entire tail including the new ones by one for every expired timer.

EmosewaMC:
> This seems logically sound to me, but from a performance perspecive it's weird to push all timers and then move up the entire tail including the new ones by one for every expired timer.

I believe I have fixed the possible perf issue.  lmk your thoughts.

aronwk-aaron (approved):
timer go brrrrrrr

Xiphoseer (approved):

PR #1417, opened 2024-01-14
Revert "fix: Remove pending timer logic"

aronwk-aaron:
Reverts DarkflameUniverse/DarkflameServer#1416

There's an infinite loading issue in NS that was introduced in this PR
Reverting till it can be resolved


Xiphoseer (approved):

Xiphoseer:
From discord
> I do not think it was valid to update the code such that this line https://github.com/DarkflameUniverse/DarkflameServer/blob/main/dGame%2FEntity.cpp#L1242 sees new timers

aronwk-aaron:
checked that the infinite loading issue is fixed here

Wincent01 (approved):

PR #1418, opened 2024-01-15
fix: timer iteration using two containers for each timer set

EmosewaMC:
Can still load into Nimbus Station (on a setup where I couldn't before the newest commit) and fixes the issue where mechs could not move around in avant gardens survival.

### DIFF
--- a/dGame/Entity.cpp
+++ b/dGame/Entity.cpp
@@ -1224,7 +1224,7 @@ void Entity::UpdateXMLDoc(tinyxml2::XMLDocument* doc) {
 }
 
 void Entity::Update(const float deltaTime) {
-	uint32_t timerPosition;
+	size_t timerPosition;
 	size_t currentSize = m_Timers.size();
 	for (timerPosition = 0; timerPosition < currentSize;) {
 		auto& timer = m_Timers[timerPosition];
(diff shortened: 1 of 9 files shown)


aronwk-aaron (dismissed):

dGame/Entity.cpp
```diff
@@ -1225,18 +1225,28 @@ void Entity::UpdateXMLDoc(tinyxml2::XMLDocument* doc) {
...
 			auto timerName = timer.GetName();
-			m_Timers.erase(m_Timers.begin() + timerPosition);
+			// We don't need to copy the element if there is only 1 element nor do we need to copy if we are on the last element.
+			// This is a clever removal trick that avoids having to copy the entire vector and instead replaces this now expired
+			// element with the last element in the vector and then removes the last element.
+			currentSize--;
+			if (currentSize > 1 && timerPosition < currentSize) {
+				m_Timers.at(timerPosition) = m_Timers.at(currentSize);
+			}
+			m_Timers.erase(m_Timers.begin() + currentSize);
```
Xiphoseer, line 1249: Since that is no longer real size - 1, it'll move all the new timers unless you do a second swap, right?

EmosewaMC: > Since that is no longer real size - 1, it'll move all the new timers unless you do a second swap, right?

no?  on erase, the only thing that happens is 1 copy and 1 shrink before any additions take place.

EmosewaMC:
I will be adding a test for this for the behavior is documented.

dGame/Entity.cpp
Xiphoseer: Can you elaborate on that?

EmosewaMC: im not sure what a second swap would even be for.  the only swapping and moving is anything that happens to the left of the added timers.  Added timers are not touched this frame.  Any moving/copying/swapping is only happening if the container grows in size and re-allocates.

EmosewaMC:
Re-tested and I can still load into nimbus station as well as see mechs move in avant gardens survival, even though technically there is no difference between this version and the one with the separate container.

PR #1425, opened 2024-01-18
fix: turn warnings into errors

EmosewaMC:
for the few warnings we get, at least make sure we listen to them now on unix platforms.  Windows has too many right now to enable /WX.

Adds a workaround to a compiler warning for two files which have said bug.
Replaces the gnu catch all flags in sqlite with the correct one for that compiler


dChatServer/PlayerContainer.cpp
```diff
@@ -24,7 +24,7 @@ PlayerContainer::~PlayerContainer() {
 }
 
 PlayerData::PlayerData() {
-	gmLevel = eGameMasterLevel::CIVILIAN;
+	gmLevel == eGameMasterLevel::CIVILIAN;
```
EmosewaMC, line 27: Intentional change for now to check if this will now fail CI.

aronwk-aaron (approved):

Xiphoseer (approved):

PR #1429, opened 2024-01-20
chore: Update render component and delete unused code

jadebenn:
A few minor changes I'm exporting from my entity/component refactor branch. I've also deleted the explicit destructors here, as they're (now) unnecessary and can prevent the implicit generation of move constructors.


EmosewaMC (changes requested):
I wouldn't replace the pointer with a unique ptr, instead consider just removing the ptr altogether and storing the struct as a non ptr in a vector instead.

dGame/dComponents/PropertyComponent.h
```diff
@@ -20,14 +20,13 @@ struct PropertyState {
...
  */
-class PropertyComponent : public Component {
+class PropertyComponent final : public Component {
 public:
-	inline static const eReplicaComponentType ComponentType = eReplicaComponentType::PROPERTY;
+	constexpr static eReplicaComponentType ComponentType = eReplicaComponentType::PROPERTY;
 	explicit PropertyComponent(Entity* parentEntity);
-	~PropertyComponent() override;
-	[[nodiscard]] PropertyState* GetPropertyState() const { return m_PropertyState; };
+	[[nodiscard]] PropertyState* GetPropertyState() const { return m_PropertyState.get(); };
```
EmosewaMC, line 27: This class aside from the ComponentType and constructor can be removed.  The component is unused entirely and not worth having stuff still in it.

dGame/dComponents/PropertyComponent.cpp
```diff
@@ -2,10 +2,8 @@
 #include "GameMessages.h"
 #include "dZoneManager.h"
 
-PropertyComponent::PropertyComponent(Entity* parent) : Component(parent) {
-	m_PropertyName = parent->GetVar<std::string>(u"propertyName");
-	m_PropertyState = new PropertyState();
+PropertyComponent::PropertyComponent(Entity* parentEntity)
+	: Component{ parentEntity }
+	, m_PropertyName{ parentEntity->GetVar<std::string>(u"propertyName") }
+	, m_PropertyState{ std::make_unique<PropertyState>() } {
```
EmosewaMC, line 8: personally I dont like using chained constructors like this for called code.  It would be better to have this in the body of the constructor instead.

EmosewaMC:
> In the remaining case I do think it makes sense to use unique ptrs. There needs to be a "null" case, I think.

Where is this null check done?  Can there not be an alternative check if it exists?

jadebenn:
> > In the remaining case I do think it makes sense to use unique ptrs. There needs to be a "null" case, I think.
> 
> Where is this null check done? Can there not be an alternative check if it exists? I do not believe Effect needs to be a ptr as the only case where we return one we always assume its non-null, therefore this pointer can be safely removed.

There is a null check in the serialization

EmosewaMC:
> There is a null check in the serialization. I'm not 100% sure of its purpose, but I was reluctant to touch it.

That is simply a good practice when accessing pointers.  The member does not need to be a pointer

EmosewaMC:
There are cases when you don't need to check (like when there is a strong guarantee the pointer will be valid, like the parent of a component), but for a container like this someone may mindlessly delete the ptr but not remove it from the container.  In this context it was just added as a backup.

dGame/dComponents/RenderComponent.cpp
```diff
@@ -42,100 +43,51 @@ RenderComponent::RenderComponent(Entity* parent, int32_t componentId): Component
...
 
-		outBitStream->Write<uint8_t>(eff->type.size());
-		for (const auto& value : eff->type) outBitStream->Write<uint16_t>(value);
+		outBitStream->Write<uint8_t>(eff.type.size());
+		for (const auto& value : eff.type) outBitStream->Write<uint16_t>(value);
 
-		outBitStream->Write<float_t>(eff->priority);
-		outBitStream->Write<int64_t>(eff->secondary);
+		outBitStream->Write<float_t>(eff.priority);
+		outBitStream->Write<int64_t>(eff.secondary);
```
EmosewaMC, line 64: As a way to test this serialization is still good, go to nimbus station Red Blocks and build the stuff on the stage and check for unserialize issues in the log file in the client.  If there are none then this should be good.  Just want to double check we dont have any serialization issues.

EmosewaMC (commented):
one final question

jadebenn:
I still haven't gotten around to checking the specific red blocks case but I didn't notice anything out of the ordinary running around VE and AG when I last tested. I'll try and confirm your specific example tonight.

EmosewaMC:
Ah, some context that may be helpful is red blocks is the only place in the whole game where this component is serialized, hence this test case

jadebenn:
Tried it in-game. I don't think this is normal... is it?
image

jadebenn:
Oddly, it's only the speaker that seems to have issues

jadebenn:
Marking as draft until this is solved.

EmosewaMC:
> Tried it in-game. I don't think this is normal... is it? image

Normal and live accurate behavior.  The live game used the wrong fx for that speaker (crux prime meteor instead of the intended one.)

EmosewaMC:
Check the client logs for the word unserialize and if you see any let me know.

jadebenn:
Checked, no issues in the logs. Should recompile fine if you rerun the workflow.

dGame/dComponents/RenderComponent.cpp
jadebenn: Checked and looks good

EmosewaMC (approved):
Pending ci

dGame/dComponents/PropertyComponent.h
```diff
@@ -1,34 +1,22 @@
 /*
  * Darkflame Universe
- * Copyright 2018
+ * Copyright 2024
```
aronwk-aaron, line 3: not relevant to this PR, but did we ever agree on having these or not?

aronwk-aaron (approved):

PR #1430, opened 2024-01-20
chore: Use references instead of pointers in StringifiedEnum.h

jadebenn:
Another export from the test branch. I'll be honest: I just like reference syntax better than pointer syntax.


EmosewaMC (approved):

EmosewaMC:
> Another export from the test branch. I'll be honest: I just like reference syntax better than pointer syntax.

I too like references better than pointers.  a reference cant be null and if a reference is invalid you have other critical issues to address.

aronwk-aaron (approved):

PR #1433, opened 2024-01-20
chore: Misc. component cleanup

jadebenn:
Formatting and syntax cleanup plus making eReplicaComponent values constexpr


EmosewaMC:
is inline static const not equivalent to constexpr?  it's just a number at the end of the day, so this shouldnt make a difference?

dGame/dComponents/SoundTriggerComponent.h
```diff
@@ -58,19 +59,19 @@ struct MixerProgram{
...
-	std::vector<MusicCue> m_MusicCues = {};
-	std::vector<MusicParameter> m_MusicParameters = {};
-	std::vector<GUIDResults> m_2DAmbientSounds = {};
-	std::vector<GUIDResults> m_3DAmbientSounds = {};
-	std::vector<MixerProgram> m_MixerPrograms = {};
+	std::vector<MusicCue> m_MusicCues{};
+	std::vector<MusicParameter> m_MusicParameters{};
+	std::vector<GUIDResults> m_2DAmbientSounds{};
+	std::vector<GUIDResults> m_3DAmbientSounds{};
+	std::vector<MixerProgram> m_MixerPrograms{};
```
EmosewaMC, line 74: nit: no brace needed here for a default constructor call anyways

EmosewaMC (dismissed):

jadebenn:
> is inline static const not equivalent to constexpr? it's just a number at the end of the day, so this shouldnt make a difference?

Difference is that constexpr values can be used in compile-time calculations. Having those enums available at compile-time might be needed for some functionality I want to implement, and even if it isn't, it certainly can't hurt.

EmosewaMC:
> > is inline static const not equivalent to constexpr? it's just a number at the end of the day, so this shouldnt make a difference?
> 
> Difference is that static constexpr members can be used in compile-time calculations. Having those members available at compile-time might be needed for some functionality I want to implement, and even if it isn't, it certainly can't hurt.

I still dont believe it makes a difference for integral types like an integer but sure

EmosewaMC (approved):

EmosewaMC:
image

jadebenn:
Looks like I was tired enough to mistake the "commit message" field for the "directory" field

dGame/dComponents/ScriptedActivityComponent.h
```diff
@@ -6,9 +6,9 @@
 
 class Entity;
 
-class ScriptedActivityComponent : public ActivityComponent {
+class ScriptedActivityComponent final : public ActivityComponent {
```
aronwk-aaron, line 9: this will not be final in the future.

aronwk-aaron (approved):

EmosewaMC (approved):
pending CI.

PR #1434, opened 2024-01-20
chore: Move database storage containers to be translation unit local to allow for safe references

EmosewaMC:
Allows for containers where values cannot be invalidated, them to be used by reference now since the container will now be alive for the lifetime of the binary and will only be destructed when main returns and after all exception handlers have finished their work.

Tested that various tables like loot, rebuilds and missions still load their data.

I wanted to provide a test asserting that a reference from a container will now no longer be invalid during a program shutdown, but unfortunately gtest/ctest really does not like you calling exit manually lol.

Rest assured it is now safe to store references to database values **assuming the underlying container cannot invalidate its references**.  Note: If the underlying container is a vector or a deque, it is not safe to return a reference in cdclient orm interfaces as insertion into these std containers will cause dangling references.  Wait for #1431


aronwk-aaron (dismissed):

EmosewaMC:
Even in a threaded context, having these containers be global for binary lifetime makes perfect sense.

EmosewaMC:
when this pr is eventually merged, an issue should be created tracking where in DLU we store copies of CDClientDatabase data and change them to use references instead.  doesnt matter who creates it, just matters that it is created.  Triage with P/Low.

Jettford (dismissed):

Xiphoseer (approved):

PR #1435, opened 2024-01-23
chore: upgrade MacOS build settings for better C++20 compatibility

jadebenn:
title


CMakeLists.txt
```diff
@@ -272,7 +273,7 @@ link_directories(${PROJECT_BINARY_DIR})
 
 # Load all of our third party directories
 add_subdirectory(thirdparty)
-if (UNIX)
+if (UNIX AND NOT APPLE)
```
EmosewaMC, line 276: Please enable upgrading of warnings to errors for apple.

.github/workflows/build-and-test.yml
```diff
@@ -25,9 +25,11 @@ jobs:
...
         vs-version: '[17,18)'
         msbuild-architecture: x64
-    - name: Install libssl (Mac Only)
-      if: ${{ matrix.os == 'macos-11' }}
-      run: brew install openssl@3
+    - name: Install libssl and switch to XCode 15.2 (Mac Only)
+      if: ${{ matrix.os == 'macos-13' }}
+      run: |
+        brew install openssl@3
+        sudo xcode-select -s /Applications/Xcode_15.2.app/Contents/Developer
```
EmosewaMC, line 32: Why is sudo being used here?  Is there no better way to accomplish the same work?

EmosewaMC (changes requested):
1 blocking and 1 question

.github/workflows/build-and-test.yml
jadebenn: The mode has to be switched manually, so I don't think there is a better way to do it.

CMakeLists.txt
jadebenn: Maybe you'll see something I'm not, but I'm getting a warning on the Mac build that doesn't seem... sane. In line 454 of UserManager.cpp, there is a warning about a [[nodiscard]] return value being discarded. But none of those values are being discarded at all, so I don't know what's setting it off.

.github/workflows/build-and-test.yml
EmosewaMC: I don't have experience with xcode-select so I cant confirm or deny this.

CMakeLists.txt
jadebenn: This is the other issue:
Screenshot_20240123_071403_Chrome

EmosewaMC: add
```


if (APPLE)
	set_source_files_properties("MD5.cpp" PROPERTIES COMPILE_FLAGS "-Wdeprecated-declarations")
endif()
``` to dCommon/CMakeLists.txt

File: dGame/UserManager.cpp [hunk 3]
```diff
@@ -451,13 +451,13 @@ void UserManager::RenameCharacter(const SystemAddress& sysAddr, Packet* packet)
 		"User %i tried to rename a character that it does not own!",
 		u->GetAccountID());
 
-	std::find_if(u->GetCharacters().begin(), u->GetCharacters().end(), [&](Character* c) {
+	static_cast<void>(std::find_if(u->GetCharacters().begin(), u->GetCharacters().end(), [&](Character* c) {
```
EmosewaMC, line 454: Just capture the return value, I believe this reads weird and it would be better to just capture instead of casting to void like this.

EmosewaMC: This just isn't syntaxes we use and it looks clunky to a reader, we generally just follow a pattern of catching the return value.  If you absolutely must, rework the find_if to do the same work and capture the return value like that.  Note that a cast to void does not always dismiss the return value from a function in some cases.  It's a pretty old syntax that I don't see used in modern day.

jadebenn: What about std::ignore? I just want to make it clear the return value is being discarded, because an 'auto' is ambiguous.

CMakeLists.txt
EmosewaMC: dCommon/CMakeLists.txt, not dGame/CMakeLists.txt

dGame/UserManager.cpp
EmosewaMC: Std ignore is not intended for use outside of std tuple and std tie.  The current pattern in the codebase is to just capture the return value and do nothing with it (we do this in other places without issue).  If you don't like the auto you could use the verbose name but it'd be just easier to capture it and do nothing.  If you must, a future pr could enable the warning for unused local variables.

EmosewaMC: I seem to be in the minority with my opinion on the use of casting to void, however even knowing so I believe it is better to re-write this code to use the return value than to discard it since this code could use the iterator anyways.

CMakeLists.txt
jadebenn: Updated

dGame/UserManager.cpp
jadebenn: Updated

jadebenn:
Once this is merged, it may be worth rerunning the workflow on #1429. It should compile with std::ranges support

aronwk-aaron (approved):
some warning in `thirdparty/SQLite/CppSQLite3.cpp` that i wonder if we could ignore, but otherwise looks good

EmosewaMC:
The same flags added to md5 and sha512 should be applied to that file.

EmosewaMC (approved):
Pending CI.

PR #1436, opened 2024-01-25
fix: remove pending callback timers logic

### LINKED ISSUE #1426: BUG: Stromling Mech not attacking in avant gardens survival

### DarkflameServer Version

Latest

### Issue Description

Stromling Mech aren't attacking player when nearby in survival minigame.

### Reproduction Steps

Join survival minigame in Avant Gardens.

### Expected Behavior

Stromg Mech should attack player.

### Environment

Ubuntu

EmosewaMC:
tested that mechs now work in avant gardens survival
fixes #1426

### DIFF
--- a/dGame/Entity.cpp
+++ b/dGame/Entity.cpp
@@ -1284,11 +1284,6 @@ void Entity::Update(const float deltaTime) {
 		m_PendingTimers.clear();
 	}
 
-	if (!m_PendingCallbackTimers.empty()) {
-		m_CallbackTimers.insert(m_CallbackTimers.end(), m_PendingCallbackTimers.begin(), m_PendingCallbackTimers.end());
-		m_PendingCallbackTimers.clear();
-	}
-
 	if (IsSleeping()) {
 		Sleep();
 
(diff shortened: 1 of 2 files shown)


Xiphoseer:
Doesn't this introduce the same issue with possibly changing the size of callbackTimers while the for loop on it is active?

EmosewaMC:
> Doesn't this introduce the same issue with possibly changing the size of callbackTimers while the for loop on it is active?

yes and it doesnt matter it uses a reference into the container.  It is fine.

EmosewaMC:
the code clearly shows `auto& callbackTimer = m_CallbackTimers[timerPosition];` which is fully safe given the condition for continuing the loop is if the size is < than size.

Xiphoseer:
> the code for the update loop uses references to elements in the timer container as `auto& callbackTimer = m_CallbackTimers[timerPosition];` which is fully safe given the condition for continuing the loop is if the index is < than size.
> 

Safe, yes. What I mean is the size() getting called as the loop condition which may change within the loop (causing infinite cycles)

EmosewaMC:
> > the code for the update loop uses references to elements in the timer container as `auto& callbackTimer = m_CallbackTimers[timerPosition];` which is fully safe given the condition for continuing the loop is if the index is < than size.
> 
> Safe, yes. What I mean is the size() getting called as the loop condition which may change within the loop (causing infinite cycles)

Ok then I will change back to the broken version which does not iterate over timers correctly.

EmosewaMC:
isolated the actual root cause.  will push a different pr soon.

PR #1437, opened 2024-01-26
fix: faction changes not allowing updated targets

### LINKED ISSUE #1426: BUG: Stromling Mech not attacking in avant gardens survival

EmosewaMC:
fixes an issue where enemies who would have their faction changed would not update their aggro targets since they will have already been forgotten in the previous iteration.  New version adds all colliding entities, but returns a trimmed list so that an enemy faction can be changed on the fly and not affect targetting for enemies who now have new enemies.
Returning a container not by reference will barely add any overhead since no copy or move is done since ive used copy elision to construct the vector in the previous stack frame instead of this one, so cost increase here is low.  No reference on the id as a LWOOBJID fits in a cpu register so a reference is more expensive than just copying the value.

Tested that stromling mechs and ronin/horsemen in forbidden valley still aggro on spawn as expected.

fixes #1426

### DIFF
--- a/dGame/Entity.cpp
+++ b/dGame/Entity.cpp
@@ -1357,16 +1357,10 @@ void Entity::OnCollisionPhantom(const LWOOBJID otherEntity) {
 	}
 
 	if (!other->GetIsDead()) {
-		auto* combat = GetComponent<BaseCombatAIComponent>();
-
-		if (combat != nullptr) {
+		if (GetComponent<BaseCombatAIComponent>() != nullptr) {
 			const auto index = std::find(m_TargetsInPhantom.begin(), m_TargetsInPhantom.end(), otherEntity);
 
 			if (index != m_TargetsInPhantom.end()) return;
-
-			const auto valid = combat->IsEnemy(otherEntity);
-
-			if (!valid) return;
 
 			m_TargetsInPhantom.push_back(otherEntity);
 		}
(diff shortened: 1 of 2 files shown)


dGame/Entity.cpp
```diff
@@ -1992,25 +1986,25 @@ void Entity::SetNetworkId(const uint16_t id) {
...
-
-		if (entity == nullptr) {
+	for (auto i = 0u; i < m_TargetsInPhantom.size(); ) {
+		if (Game::entityManager->GetEntity(m_TargetsInPhantom.at(i))) {
+			i++;
 			continue;
 		}
-
-		valid.push_back(id);
+		m_TargetsInPhantom.erase(m_TargetsInPhantom.begin() + i);
```
Xiphoseer, line 1996: This seems a weird way to filter a list, especially since there is another loop below that pushes to enemies?

Xiphoseer (dismissed):
LGTM

Would prefer to not erase in a loop and use a standard thing like <https://cplusplus.com/reference/algorithm/copy_if/> instead

aronwk-aaron (approved):

dGame/Entity.cpp
EmosewaMC: Well the first one checks for stale entities and prunes them, then a second loop actually does the checks for what to return.  I guess I could consolidate the loops but even then erase in a loop like this doesn't have high cost since there is no re allocation.  Best save here would just be using erase_if first and then copy_if after the null entities have been erased, but this is still 2 loops

aronwk-aaron (approved):

Jettford (approved):
yummy code

PR #1439, opened 2024-01-30
feat: Make use of CMake presets to enable easy switching between debug and release configurations on all platforms

jadebenn:
Enable some optimization options for MSVC builds and altered the CMakePresets.json to make it easier to switch between debug and release configurations on all platforms.


EmosewaMC (changes requested):
MSVC build optimizations are handled by CMakePresets.json with build types there.  I dont believe we should meddle with the optimizer in this case.  MSVC by default with build.sh builds in Release and must have a build type specified if building through CLI.  As such I dont believe this change is necessary.

If there are further details I am missing in regards to these flags please let me know, but as far as I know all this is handled by Config types in CMake by default.
further sources actually support doing the opposite, removing optimization flags from our cmake and using the [command line instead](https://stackoverflow.com/questions/41361631/optimize-in-cmake-by-default) for this

jadebenn:
I'm not sure I understand your point. While it's possible the O2 flag is being set on MSVC currently (haven't checked), based in the arifact size changing between enabling some of the other flags vs. not I'm fairly certain they aren't being applied currently.

EmosewaMC:
That's because you are manually overriding them via the cmake lists

jadebenn:
My point being that if I was manually overriding the defaults with the same flags, nothing should change. Therefore some of these options aren't normally being enabled.

jadebenn:
Moving the actual setting of flags to the .json would make sense though:

https://www.reddit.com/r/cpp_questions/comments/taxyut/what_is_the_best_way_to_set_compiler_flags_in_a/i04e1hk/

jadebenn:
Any way I can get workflow approval in the future somehow, btw? Would be really nice when testing these multi-platfor changes.

EmosewaMC:
Ideally you should test this locally for function instead of relying on CI for multi platform changes

EmosewaMC:
I have done a bunch of research and the only thing that makes a difference is /Zc:inline in the add_compile_options for MSVC.  /O2 breaks Debug builds.

jadebenn:
Possibly widening the scope of this PR to make better use of the JSON configurations for all platforms. We can have separate debug and release configurations that define the relevant flags to pass to CMake, and only keep the common flags for each platform defined in CMakeLists.txt. Thoughts?

EmosewaMC:
I'd say play with it and see how it changes compilation sizes and builds.  We haven't really touched this file since release since everything just works right now, but it may be worth experimenting and fixing that todo in the cmake file

jadebenn:
Edited the title and description of this PR to reflect its new purpose

EmosewaMC:
How does this affect final builds as far as size and crash info (crash info on Linux specifically since windows uses mini dump)?  How does one use one or the other?

EmosewaMC:
also has the readme and build script been updated to use presets?  i recall you need to provide this in the generation step and not the build step for the presets via command line.

jadebenn:
I checked that the stack traces on Linux are intact. Unsure about Windows. You're right that I need to update the readme and build script; forgot about that. I _think_ the build script will use the default settings currently.

Xiphoseer:
IIRC RelWithDebInfo was used instead of Relase not because of convenience but because of a bug somewhere.

I don't recommend changing this until that is figured out.

EmosewaMC:
> IIRC RelWithDebInfo was used instead of Relase not because of convenience but because of a bug somewhere.
> 
> I don't recommend changing this until that is figured out.

These are all fixed.  They were from a stack overflow and an odd combination of SQLStrings and std::strings.

EmosewaMC:
> > IIRC RelWithDebInfo was used instead of Relase not because of convenience but because of a bug somewhere.
> > I don't recommend changing this until that is figured out.
> 
> These are all fixed. They were from a stack overflow and an odd combination of SQLStrings and std::strings.

Hmm I think I mixed up the debug issues and the release ones.  I think it was the debug build which had those errors, not release.  Do do some playing of the game and see if you can spot any issues

jadebenn:
Persistent issues with the MacOS build right now. Not sure what the cause is, and it's hard for me to debug. Ubuntu build looks good. Trying to tweak the MSVC build to get some better integration with visual studio itself but it's hard for me because my install doesn't seem to like DLU very much.

jadebenn:
thank _christ_ I finally got the macos build working

jadebenn:
unfortunate, but for everyone else this should be ready for actual review now. the line of commits was to nail down a MacOS build, and that should be good now

.github/workflows/build-and-test.yml
```diff
@@ -33,9 +33,7 @@ jobs:
     - name: cmake
       uses: lukka/run-cmake@v10
       with:
-        configurePreset: "ci-${{matrix.os}}"
-        buildPreset: "ci-${{matrix.os}}"
-        testPreset: "ci-${{matrix.os}}"
+        workflowPreset: "ci-${{matrix.os}}"
```
Xiphoseer, line 36: Config change without version change?

build.sh
```diff
@@ -6,8 +6,7 @@ mkdir -p build
 cd build
 
 # Run cmake to generate make files
-cmake ..
+cmake -DCMAKE_BUILD_TYPE="Release" ..
```
Xiphoseer, line 9: I presume that's way to complicated for people, unfortunately. Also not sure why it's not the default

tests/CMakeLists.txt
```diff
@@ -4,8 +4,6 @@ enable_testing()
 find_package(GoogleTest REQUIRED)
 include(GoogleTest)
 
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bin)
```
Xiphoseer, line 7: Agree with this in general, but again, was added because someone had strong opinions on build directory layout to make running from build easier/possible.

Xiphoseer (commented):
This PR feels like a weird mix of a bit of everthing. Is this forward compat because our libraries are old? Is this nice to have?

It seems to be more standard in some places (build dir) and less standard in others (clang-16)

tests/CMakeLists.txt
jadebenn: For reference, this change was what solved the 40ish commit hell of builds failing on MacOS. Unless someone can figure out another way to fix that, I'd rather not touch this particular change again.

build.sh
jadebenn: Confused by what you mean here.

.github/workflows/build-and-test.yml
jadebenn: Not sure what you mean

CMakePresets.json
```diff
@@ -58,19 +96,36 @@
...
+        "description": "This preset is used to debug the build on linux",
+        "configuration": "Debug",
+        "jobs": 2
+      },
       {
         "name": "ci-windows-2022",
         "configurePreset": "ci-windows-2022",
         "displayName": "Windows CI Build",
         "description": "This preset is used by the CI build on windows",
-        "configuration": "RelWithDebInfo",
```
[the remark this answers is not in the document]
jadebenn: Did not mean to remove that line.

jadebenn:
> This PR feels like a weird mix of a bit of everthing. Is this forward compat because our libraries are old? Is this nice to have?
> 
> It seems to be more standard in some places (build dir) and less standard in others (clang-16)

I'm specifically targeting clang-16 because the build fails on clang-14, which is the default install for Ubuntu via a package manager. If it can't find 16 in particular, it falls back to just searching for _a_ clang install.

This is mostly a nice to have so we can set up multiple build configurations in subdirectories of the build folder, but it's opt-in so as to not break the reams of tutorials that presume otherwise.

.github/workflows/build-and-test.yml
Xiphoseer: Whether workflowPreset has always existed in that version of the action

build.sh
Xiphoseer: For some reason I thought this was README. But on that note it would be weird to have build sh do sth special not added to README

tests/CMakeLists.txt
Xiphoseer: Was it related to mariadb linking / rpath?

Xiphoseer:
> > This PR feels like a weird mix of a bit of everthing. Is this forward compat because our libraries are old? Is this nice to have?
> > 
> > It seems to be more standard in some places (build dir) and less standard in others (clang-16)
> 
> I'm specifically targeting clang-16 because the build fails on clang-14, which is the default install for Ubuntu via a package manager. If it can't find 16 in particular, it falls back to just searching for _a_ clang install, which may or may not be able to compile DLU successfully.
> 
> This is mostly a nice to have so we can set up multiple build configurations in subdirectories of the build folder, but it's opt-in so as to not break the reams of tutorials that presume otherwise.

Nice to have yes, but I'd presume few people install ahead of default clang packages in their distro and this logic will need patching as time passes and clang-17 becomes default.

tests/CMakeLists.txt
jadebenn: Yes

build.sh
jadebenn: I don't think it would really do something "special" per se. Running build.sh would just automatically generate a release build - which is pretty much what it does now. I think it makes much more sense for _debug_ builds to be opt-in.

jadebenn:
> Nice to have yes, but I'd presume few people install ahead of default clang packages in their distro and this logic will need patching as time passes and clang-17 becomes default.
> 
> Also: which version of ubuntu? The GH CI? LTS?

Sorry, should have clarified: The Ubuntu distribution packaged with WSL has Clang-14. You have to install more recent versions manually.

If Cmake can't find clang-16, it defaults to just grabbing the path to a generic "clang" executable. This should prevent any issues in the future if this isn't touched. The clang-16 lookup would fail, but cmake would just invoke clang, which would presumably be a version > 16 at that point.

.github/workflows/build-and-test.yml
jadebenn: I believe so? Still not quite sure what you mean by the "version" of the action. Do you mean the commit?

CMakePresets.json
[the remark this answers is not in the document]
jadebenn: Oh, wait, are you not asking why I replaced it with release? Are we using the debug info for anything on Windows currently?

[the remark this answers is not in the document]
Xiphoseer: Yes, ReWithDeb vs Release was the point. I'd have to dig through old chats to figure out why. Maybe for consistency with platforms that need it for backtrace?

.github/workflows/build-and-test.yml
Xiphoseer: The
> uses: lukka/run-cmake@v10
specifically the "v10" git tag of that repo

build.sh
Xiphoseer: Yes, debug should be opt-in so the Q is why "cmake .." is not sufficient for a release build.

.github/workflows/build-and-test.yml
jadebenn: Ah. The current v10(.7) has it. The original v10(.0) did not.

build.sh
jadebenn: The CMAKE_BUILD_TYPE flag isn't being set to "Release" for single-config generators, so no release flags are being applied if you just run cmake ..

This isn't a problem for multi-config generators, which will ignore the CMAKE_BUILD_TYPE flag anyway.

I _guess_ I could try and jerry-rig a system where if a configuration option isn't passed, CMAKE_BUILD_TYPE is set to release by default. Is that really desired?

CMakePresets.json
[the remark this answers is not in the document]
jadebenn: We're not generating debug symbols on Linux builds currently. The stack traces seem to be generated from the function symbols themselves (hence why there was an issue a while back when I set them to default=hidden).

jadebenn:
For clarification, I moved the -Werror flag into the presets.json so it doesn't trigger on builds from the command line of build.sh. This is so that if DLU is ever abandoned or idle for a prolonged period of time, users won't have to figure out how to disable it to make new builds. It's still applied to CI runs and runs from presets, though.

jadebenn:
Hey, I haven't heard from anyone on this in a while and I don't have further changes I want to make. Can we look at merging this?

aronwk-aaron (dismissed):
looks good, will let Xipho have the final say

jadebenn:
@Xiphoseer Are you still looking at this?

EmosewaMC (dismissed):
I will trust that this works and is well tested

CMakeLists.txt
```diff
@@ -8,8 +11,10 @@ endif()
 
 include(CTest)
 
+set(CMAKE_C_STANDARD 99)
```
EmosewaMC, line 14: We should consider uping this to 17 in a future PR

EmosewaMC (approved):
thanks so much for taking on this task.  I know this took a lot of your time and effort to get done :)  Only one note for the future

CMakeLists.txt
aronwk-aaron: 20 /s

aronwk-aaron (approved):

PR #1442, opened 2024-02-02
refactor: Change TryParse implementation

jadebenn:
Changed the TryParse function to use from_chars(), natively handle enums, and have a better syntax. Also did some cleanup along the way, particularly regarding the function declarations of QuickbuildComponent.h.

Need to see how these changes compile on the other platforms, but have had good experience testing on my own.


EmosewaMC:
I haven't read the changlist yet but please don't include a general cleanup here.  If it's cleanup related to try parse please be more specific as to what was cleaned up as general cleanup implies you changed code unrelated to try parsing and we are trying to refrain from mixing changes like that so bisecting is easier

jadebenn:
> Can you elaborate on general cleanup?

I can try and split out the unrelated changes later. It's kind of hard because I tend to notice and address issues as I'm working on a primary goal, but it isn't good practice to mix them up I agree.

For the most part, they're really minor. A few [[nodiscard]] tags here, a few consts there... that sort of thing.

EmosewaMC:
> For the most part, they're really minor. A few [[nodiscard]] tags here, a few consts there... that sort of thing.

Just as a note, we wont be enforcing nodiscard as an attribute in any prs, nor will we be requiring it for apis.

jadebenn:
> > For the most part, they're really minor. A few [[nodiscard]] tags here, a few consts there... that sort of thing.
> 
> Just as a note, we wont be enforcing nodiscard as an attribute in any prs, nor will we be requiring it for apis. Nothing against adding them, just wont be enforcing it

Didn't expect any different, just trying to do my part for keeping code intentions clear.

As far as the failed CI, I suspected that it wouldn't work on MacOS given our previous conversations, so I actually have the fallback floating point code ready to go (just commented out). I much prefer from_chars though, so I'll probably wrap it in a pre-processor block so it is ONLY present on Mac.

One neat thing is that the way the concept resolution works, I don't actually need to change anything in the other templates to make floating points switch over: They see the more specific constraint and immediately start using that one. At least, this is the behavior I saw on my end.

EmosewaMC:
how has this been tested?

dCommon/GeneralUtils.h
```diff
@@ -123,90 +125,117 @@ namespace GeneralUtils {
...
-	inline uint32_t Parse(const char* value) {
-		return std::stoul(value);
+		return isParsed ? static_cast<T>(result) : std::optional<T>{};
 	}
 
-	template <>
-	inline uint64_t Parse(const char* value) {
-		return std::stoull(value);
-	}
+#ifdef __APPLE__
```
EmosewaMC, line 166: I would include `dPlatforms.h` and use the macros in there for the platform type instead.

EmosewaMC (changes requested):
I am only reviewing the GeneralUtils.h header for now until the other unrelated changes are reverted.

jadebenn:
I've run around VE, AG, and Ninjago completing missions and playing normally for about 90 minutes and haven't noticed anything strange with the Ubuntu build.

dCommon/GeneralUtils.h
```diff
@@ -123,90 +125,117 @@ namespace GeneralUtils {
...
+		return x && y && z ? std::make_optional<NiPoint3>(x.value(), y.value(), z.value()) : std::nullopt;
 	}
 
-	bool TryParse(const std::string& x, const std::string& y, const std::string& z, NiPoint3& dst);
+	/**
+	 * The TryParse overload for handling NiPoint3 by passingn a reference to a vector of three strings
+	 * @param str The string vector representing the X, Y, and Xcoordinates
+	 * @returns An std::optional containing the desired NiPoint3 if it can be constructed from the string parameters
+	*/
+	template <typename T>
```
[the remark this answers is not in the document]
jadebenn: Oddly, I get substitution failures when I do so.

dCommon/GeneralUtils.h
```diff
@@ -123,90 +125,117 @@ namespace GeneralUtils {
...
-		return std::stoul(value);
-	}
+	/**
+	 * For numeric values: Parses a C-style char range (string) and returns an optional variable depending on the result
+	 * @param str The pointer to the start of the char range (string)
+	 * @param strEnd The pointer to the end of the char range (string), defaults to NULL
+	 * @returns An std::optional containing the desired value if it exists in the string
+	*/
+	template <Numeric T>
+	[[nodiscard]] std::optional<T> TryParse(const char* const str, const char* const strEnd = NULL) noexcept {
```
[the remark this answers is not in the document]
jadebenn: I noticed that too. However, cppreference also notes that the function does not use exceptions and
> Exceptions
> Throws nothing.
I do suppose that marking 'noexcept' does risk that the internal implementation could throw, however, and even if we can verify it currently doesn't it's probably best to not assume that will always be the case. I'll have to adjust the noexcept status of some of the upstream functions, but it's not too big a deal.

jadebenn:
> I am only reviewing the GeneralUtils.h header for now until the other unrelated changes are reverted. There are a lot of unrelated changes here which make the pr much larger to review than it needs to be

I don't think there were very many? I had to do a lot of changes relating to the actual TryParse invocations, but aside from QuickbuildComponent (which I will split out sometime tomorrow) it was pretty much just a few 'consts' on things I saw that didn't need to be mutable.

jadebenn:
Ah, forgot about dCommonVars.h I see what you mean a bit better now.

EmosewaMC:
> > I am only reviewing the GeneralUtils.h header for now until the other unrelated changes are reverted. There are a lot of unrelated changes here which make the pr much larger to review than it needs to be
> 
> I don't think there were very many? I had to do a lot of changes relating to the actual TryParse invocations, but aside from QuickbuildComponent (which I will split out sometime tomorrow) it was pretty much just tossing a few 'consts' on things I saw that didn't need to be mutable. Even QuickbuildComponent was just adjusting for const correctness, attrib tagging, and marking noexcept.

In the future we'd appreciate isolated pr changes for things that are very much not related.

EmosewaMC:
Yeah commonvars.h is the main one I was looking at.  With those being really compiler only changes they would be better suited in their own pr, but if you don't want to move them now it's fine, just will take more time to review

dCommon/GeneralUtils.h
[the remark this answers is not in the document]
EmosewaMC: Ah right, this is an overloaded function, not a template of a previous one.

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -1905,13 +1902,14 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 	if (chatCommand == "inspect" && entity->GetGMLevel() >= eGameMasterLevel::DEVELOPER && args.size() >= 1) {
 		Entity* closest = nullptr;
 
-		eReplicaComponentType component;
+		//eReplicaComponentType component;
```
EmosewaMC, line 1905: Should just be removed

dMasterServer/InstanceManager.cpp
```diff
@@ -17,7 +17,8 @@
 InstanceManager::InstanceManager(Logger* logger, const std::string& externalIP) {
 	mLogger = logger;
 	mExternalIP = externalIP;
-	GeneralUtils::TryParse(Game::config->GetValue("world_port_start"), m_LastPort);
+	m_LastPort =
+		GeneralUtils::TryParse<decltype(m_LastPort)>(Game::config->GetValue("world_port_start")).value_or(m_LastPort);
```
EmosewaMC, line 21: ports are always a 16 bit number, though I get the modularity.

dNet/AuthPackets.cpp
```diff
@@ -246,9 +244,9 @@ void AuthPackets::SendLoginResponse(dServer* server, const SystemAddress& sysAdd
 	uint16_t version_major = 1;
 	uint16_t version_current = 10;
 	uint16_t version_minor = 64;
-	GeneralUtils::TryParse<uint16_t>(Game::config->GetValue("version_major"), version_major);
-	GeneralUtils::TryParse<uint16_t>(Game::config->GetValue("version_current"), version_current);
-	GeneralUtils::TryParse<uint16_t>(Game::config->GetValue("version_minor"), version_minor);
+	version_major = GeneralUtils::TryParse<uint16_t>(Game::config->GetValue("version_major")).value_or(version_major);
+	version_current = GeneralUtils::TryParse<uint16_t>(Game::config->GetValue("version_current")).value_or(version_current);
+	version_minor = GeneralUtils::TryParse<uint16_t>(Game::config->GetValue("version_minor")).value_or(version_minor);
```
EmosewaMC, line 249: Is there a reason the above declarations arent just simplified to uint16_t version... .value_or(1);?

dScripts/02_server/Map/AM/AmSkullkinTower.cpp
```diff
@@ -144,13 +144,10 @@ void AmSkullkinTower::OnChildRemoved(Entity* self, Entity* child) {
 			);
 
 			for (const auto& mission : missions) {
-				int32_t missionID = 0;
+				const auto missionID = GeneralUtils::TryParse<LOT>(mission);
```
EmosewaMC, line 147: missions arent lots, should be changed to int32_t.  a LOT is a LEGO Object Template aka a pre-fab object to be placed in the world.

dZoneManager/Level.cpp
```diff
@@ -214,9 +214,9 @@ void Level::ReadSceneObjectDataChunk(std::istream& file, Header& header) {
 	gating.major = 1;
 	gating.current = 10;
 	gating.minor = 64;
-	GeneralUtils::TryParse<int32_t>(Game::config->GetValue("version_major"), gating.major);
-	GeneralUtils::TryParse<int32_t>(Game::config->GetValue("version_current"), gating.current);
-	GeneralUtils::TryParse<int32_t>(Game::config->GetValue("version_minor"), gating.minor);
+	gating.major = GeneralUtils::TryParse<int32_t>(Game::config->GetValue("version_major")).value_or(gating.major);
```
EmosewaMC, line 217: Same as authpackets feedback

dCommon/GeneralUtils.h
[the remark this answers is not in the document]
EmosewaMC: Yes, I also noticed that the function does not throw exceptions as per the std, however the key thing that confused me was the standards docs lack of the definition being noexcept, which prompted me to just check the big 3 compilers for implementation.  the floating point is generally noexcept, but the integral ones are not always so.

dNet/AuthPackets.cpp
jadebenn: I was debating the readability

dGame/dInventory/Item.cpp
```diff
@@ -468,14 +468,14 @@ void Item::DisassembleModel(uint32_t numToDismantle) {
 	// First iteration gets the count
 	std::map<int32_t, int32_t> parts;
 	while (currentBrick) {
-		auto* designID = currentBrick->Attribute("designID");
+		const char* const designID = currentBrick->Attribute("designID");
```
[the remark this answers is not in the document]
jadebenn: Honestly, good question. I think my logic was that if a variable doesn't change, I tend to make it as immutable as possible, even at the expense of verbosity.

dNet/AuthPackets.cpp
jadebenn: Went ahead and made the defaults seperate constexpr variables in the next revision for readability, and made the non-defaults seperate. Ideally this is functionally equivalent to the other approach (because they'll be substituted in at compile-time) but more readable.

jadebenn:
Actually, no I'm not going to rename that variable with the m_ prefix. That would need to be a PR of its own because it creates _so_ many problems. I will be reverting my commits to that effect.

dGame/dPropertyBehaviors/ControlBehaviorMessages/BehaviorMessageBase.cpp
```diff
@@ -5,20 +5,21 @@
 #include "dCommonVars.h"
 
 BehaviorMessageBase::BehaviorMessageBase(AMFArrayValue* arguments) {
-	behaviorId = GetBehaviorIdFromArgument(arguments);
+	this->behaviorId = GetBehaviorIdFromArgument(arguments);
```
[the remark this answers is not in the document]
jadebenn: See update for why I will not be changing this. It required the editing of 20ish files, and that was indiscriminate through a ctrl+f find and replace operation, so it changed quite a few unrelated things.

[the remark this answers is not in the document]
EmosewaMC: ok but please try to not rope random changes into random prs because you spot them as you go.  it sets a bad precedence of every pr just being effectively "general fixes"

dCommon/dEnums/dCommonVars.h
```diff
@@ -27,6 +27,13 @@ constexpr uint32_t highFrameDelta = FRAMES_TO_MS(highFramerate);
 constexpr uint32_t mediumFrameDelta = FRAMES_TO_MS(mediumFramerate);
 constexpr uint32_t lowFrameDelta = FRAMES_TO_MS(lowFramerate);
 
+//========== CLIENT VERSION DEFAULTS ===========
+namespace ClientVersion {
+	constexpr uint16_t major = 1;
+	constexpr uint16_t current = 10;
+	constexpr uint16_t minor = 64;
+}
+
```
EmosewaMC, line 36: these do not need to be included in every single source file and can be moved back to their original location.

jadebenn: Why not? They're constexpr, so they don't actually exist at runtime. It's effectively the same as something like:

#define MAJOR_VERSION 1
#define CURRENT_VERSION 10
#define MINOR VERSION 64

EmosewaMC: yes but this is a PCH.  Every source file does not need this information in it.

jadebenn: Moving it to a new header file

jadebenn:
okay pinky swear that is It. I will resist the urge to add anything more unless the CI fails again

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -776,34 +774,34 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
...
+			const auto itemLOT = GeneralUtils::TryParse<uint32_t>(args.at(0));
 
-			if (!GeneralUtils::TryParse(args[0], itemLOT)) {
+			if (!itemLOT) {
 				ChatPackets::SendSystemMessage(sysAddr, u"Invalid item LOT.");
 				return;
 			}
 
-			uint32_t count;
+			const auto count = GeneralUtils::TryParse<uint32_t>(args.at(0));
```
EmosewaMC, line 795: The index was changed here

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -1024,25 +1023,25 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 			time_t expire = 1; // Default to indefinate mute
 
 			if (args.size() >= 2) {
-				uint32_t days = 0;
-				uint32_t hours = 0;
-				if (!GeneralUtils::TryParse(args[1], days)) {
+				const auto days = GeneralUtils::TryParse<uint32_t>(args.at(1));
+				const auto hours = GeneralUtils::TryParse<uint32_t>(args.at(2));
```
EmosewaMC, line 1027: This logic was moved from its previous wrapper to verify there were enough elements in the vector and will crash if this is tried to be used

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -1024,25 +1023,25 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
...
 
 						return;
 					}
 				}
 
 				expire = time(NULL);
-				expire += 24 * 60 * 60 * days;
-				expire += 60 * 60 * hours;
+				expire += 24 * 60 * 60 * days.value_or(0);
+				expire += 60 * 60 * hours.value_or(0);
```
EmosewaMC, line 1044: If we got to here, days must have been valid

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -1278,22 +1278,22 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
...
 		// MODERATION should work but it doesn't.  Relog to see uscore changes
 
 		eLootSourceType lootType = eLootSourceType::MODERATION;
 
-		int32_t type;
-		if (args.size() >= 2 && GeneralUtils::TryParse(args[1], type)) {
-			lootType = static_cast<eLootSourceType>(type);
+		const auto type = GeneralUtils::TryParse<eLootSourceType>(args[1]);
+		if (args.size() >= 2 && type) {
+			lootType = type.value();
```
EmosewaMC, line 1296: The index check here was moved to be after the value is indexed

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -1432,71 +1433,63 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
...
 		return;
 	}
 
 	if (chatCommand == "buff" && args.size() >= 2 && entity->GetGMLevel() >= eGameMasterLevel::DEVELOPER) {
 		auto* buffComponent = entity->GetComponent<BuffComponent>();
 
-		int32_t id = 0;
-		int32_t duration = 0;
+		const auto id = GeneralUtils::TryParse<int32_t>(args[0]);
+		const auto duration = GeneralUtils::TryParse<int32_t>(args[1]);
```
EmosewaMC, line 1466: nit: no point in parsing duration if id is invalid

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -1507,10 +1500,12 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 				force = true;
 			}
 
-			if (args.size() > index && !GeneralUtils::TryParse(args[index], cloneId)) {
+			const auto cloneIdOptional = GeneralUtils::TryParse<LWOCLONEID>(args[index]);
+			if (args.size() > index && !cloneIdOptional) {
```
EmosewaMC, line 1504: index check was changed to be after the access here

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -1673,20 +1666,20 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 	}
 
 	if (chatCommand == "reforge" && entity->GetGMLevel() >= eGameMasterLevel::DEVELOPER && args.size() >= 2) {
-		LOT baseItem;
-		LOT reforgedItem;
+		const auto baseItem = GeneralUtils::TryParse<LOT>(args[0]);
+		const auto reforgedItem = GeneralUtils::TryParse<LOT>(args[1]);
```
EmosewaMC, line 1670: nit: no point in parsing reforgedItem is baseItem didnt parse

dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -1752,19 +1745,19 @@ void SlashCommandHandler::HandleChatCommand(const std::u16string& command, Entit
 	}
 
 	if (chatCommand == "rollloot" && entity->GetGMLevel() >= eGameMasterLevel::OPERATOR && args.size() >= 3) {
-		uint32_t lootMatrixIndex = 0;
-		uint32_t targetLot = 0;
-		uint32_t loops = 1;
+		const auto lootMatrixIndex = GeneralUtils::TryParse<uint32_t>(args[0]);
+		const auto targetLot = GeneralUtils::TryParse<uint32_t>(args[1]);
+		const auto loops = GeneralUtils::TryParse<uint32_t>(args[2]);
```
EmosewaMC, line 1750: These should be unrolled so we arent constantly constructing and destructing optional objects in the loop below

EmosewaMC (changes requested):

dNet/AuthPackets.cpp
```diff
@@ -39,10 +40,9 @@ void AuthPackets::LoadClaimCodes() {
 	auto rcstring = Game::config->GetValue("rewardcodes");
 	auto codestrings = GeneralUtils::SplitString(rcstring, ',');
 	for(auto const &codestring: codestrings){
-		uint32_t code = -1;
-		if(GeneralUtils::TryParse(codestring, code) && code != -1){
-			claimCodes.push_back(code);
-		}
+		const auto code = GeneralUtils::TryParse<uint32_t>(codestring);
+
+		if (code) claimCodes.push_back(code.value());
```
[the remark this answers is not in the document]
jadebenn: Are you saying that I need to check if the value is -1 as well?

dGame/dUtilities/SlashCommandHandler.cpp
jadebenn: Can you clarify?

dGame/dUtilities/SlashCommandHandler.cpp
jadebenn: I hate that I'm basically replying to all these with "I'm not sure what you mean," but I'm not sure what you mean here.

dGame/dUtilities/SlashCommandHandler.cpp
jadebenn: Not sure how to avoid that while also getting the return value

dGame/dUtilities/SlashCommandHandler.cpp
jadebenn: Same issue as the other one. If I fold that into the if statement, I'm not sure how to get the value out without calling the function a second time.

dGame/dUtilities/SlashCommandHandler.cpp
jadebenn: It's not though?

dGame/dUtilities/SlashCommandHandler.cpp
jadebenn: Oh, see what you mean now.

dGame/dUtilities/SlashCommandHandler.cpp
EmosewaMC: This one is fine, pretty sure I saw a copy in the loop below at some point or something idk, i dont see a copy happening in the loop at a second glance

dGame/dUtilities/SlashCommandHandler.cpp
jadebenn: Figured it out.

dGame/dUtilities/SlashCommandHandler.cpp
jadebenn: Ah, the hours could potentially be.

dGame/dUtilities/SlashCommandHandler.cpp
jadebenn: I'll double check for you

jadebenn: Think I found a potential copy in a for-ranged loop. Not related to my changes, but I'll fix it just in case.

dNet/AuthPackets.cpp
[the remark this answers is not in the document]
EmosewaMC: Yes, as it was before since that what the original did

EmosewaMC (approved):
pending ci

aronwk-aaron (approved):

PR #1443, opened 2024-02-03
chore: remove SHA512, move md5 to thirdparty

EmosewaMC:
We don't use SHA512 anymore, no point in keeping it around since it was copy pasted code anyways.

Tested that Auth runs and I can login still.


aronwk-aaron (approved):

jadebenn (approved):
Lgtm

PR #1444, opened 2024-02-03
chore: General cleanup roundup

jadebenn:
Split out unrelated changes from #1442. General cleanup of things I noticed while I was working on the TryParse branch.


dChatServer/PlayerContainer.h
```diff
@@ -45,7 +43,7 @@ struct PlayerData {
 	std::string playerName;
 	std::vector<FriendData> friends;
 	std::vector<IgnoreData> ignoredPlayers;
-	eGameMasterLevel gmLevel;
+	eGameMasterLevel gmLevel = eGameMasterLevel::CIVILIAN;
```
EmosewaMC, line 46: This was left out of the header on purpose to keep it small.  Is there a reason we must transitively include this header now?  Headers should be kept as small and have as few includes as possible in them, so changing this just adds another include for seemingly no reason?

dCommon/dEnums/dCommonVars.h
```diff
@@ -39,86 +39,84 @@ constexpr uint32_t lowFrameDelta = FRAMES_TO_MS(lowFramerate);
...
-
-const float PI = 3.14159f;
+using LOT = int32_t;                        //!< A LOT
+using LWOOBJID = int64_t;                   //!< An object ID (should be unsigned actually but ok)
+using TSkillID = int32_t;                   //!< A skill ID
+using LWOCLONEID = uint32_t;                 //!< Used for Clone IDs
+using LWOMAPID = uint16_t;                   //!< Used for Map IDs
+using LWOINSTANCEID = uint16_t;              //!< Used for Instance IDs
+using PROPERTYCLONELIST = uint32_t;          //!< Used for Property Clone IDs
+using StripId = uint32_t;
```
EmosewaMC, line 49: using x = y and typedef y x are exactly the same, not really a change that is necessary.

dGame/dComponents/QuickBuildComponent.h
```diff
@@ -24,17 +24,17 @@ class QuickBuildComponent final : public Component {
...
 	void Serialize(RakNet::BitStream* outBitStream, bool bIsInitialUpdate) override;
-	void Update(float deltaTime) override;
+	void Update(const float deltaTime) override;
 
 	/**
 	 * Handles a OnUse event from some entity, initiating the quick build
 	 * @param originator the entity that triggered the event
 	 */
-	void OnUse(Entity* originator) override;
+	void OnUse(Entity* const originator) override;
```
EmosewaMC, line 37: adding const here makes it not match the function its overloading, same with Update.  These not matching semantically is odd and not something I think would be good design, the overloads should match the base function it is overloading imo.  Thoughts?

dChatServer/PlayerContainer.h
jadebenn: I wanted to avoid specifying a non-default constructor, even if said constructor only contains one assignment, and I figured the cost of #include-ing one enum class was worth it

dCommon/dEnums/dCommonVars.h
jadebenn: Not _exactly_ the same, though yes, in this particular context they are. Still, I like standardizing our notation and typedef's really C legacy.

dGame/dComponents/QuickBuildComponent.h
jadebenn: I agree. I forgot this was an overload.

dChatServer/PlayerContainer.h
EmosewaMC: While I would normally agree, the pattern we are following is making headers as small as possible.  Not specifying a constructor is a want here rather than a need and I believe it is good practice to not include something you don't need.

jadebenn: What about using a cast? That won't require an include

jadebenn: Ended up using the cast

EmosewaMC: I mean that _works_ but let me ask you this, in the future, if you saw `eGameMasterLevel gmLevel = static_cast<eGameMasterLevel>(0); // CIVILIAN`, I believe it would not be too far fetched to believe someone would want to remove the cast and just either include the file or make a constructor and do the work in the translation unit.  I believe it was fine before as a constructor as it initially was.  I know other would just say include the file, but my stance in headers has always been keep them small so transitive includes and rebuilds happen infrequently.  I won't block this change, but if I saw `eGameMasterLevel gmLevel = static_cast<eGameMasterLevel>(0); // CIVILIAN` I would just make a constructor and remove the comment and assignment.

jadebenn: My reasoning is the moment you define a constructor - even if that's just to define it _as_ default - you open Pandora's box of all sorts of seemingly unrelated effects to copies, moves... Is it really worth it in this case solely to spare an include of a header file that only holds one single enum class?

EmosewaMC:
I will wait until you finish adding in random changes to review this.

jadebenn:
All the changes I have added to this since the initial commit have been to respond to your requests.

dChatServer/PlayerContainer.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 #include <map>
 #include "dCommonVars.h"
+#include "eGameMasterLevel.h"
```
EmosewaMC, line 4: still including the header, if you are going the static_cast route

jadebenn: whoops. fixing...

jadebenn:
looks like I finally got the instantiation problem fixed. should be ready to go unless you all have any other feedback here

dCommon/Amf3.h
```diff
@@ -2,6 +2,7 @@
 #define __AMF3__H__
 
 #include "dCommonVars.h"
+#include "dPlatforms.h"
```
EmosewaMC, line 5: no longer need this

EmosewaMC (approved):

EmosewaMC (dismissed):

jadebenn:
aaaand false alarm. We're good.

aronwk-aaron (dismissed):

jadebenn:
...wow, they really require re-review for fixing merge conflicts?

jadebenn:
annoying.

aronwk-aaron (approved):
Pending ci

EmosewaMC (approved):

PR #1449, opened 2024-02-04
feat: Cross-platform compiler attributes

jadebenn:
Add support for the 'inline' compiler flags by way of the C preprocessor


EmosewaMC:
But why?

PR #1453, opened 2024-02-11
chore: make predefined name unique when the name files are not provided

aronwk-aaron:
Multiple characters would not be allowed to be made since they would both have the name INVALID, that is no longer the case

### DIFF
--- a/dGame/UserManager.cpp
+++ b/dGame/UserManager.cpp
@@ -301,6 +301,7 @@ void UserManager::CreateCharacter(const SystemAddress& sysAddr, Packet* packet) 
 
 	const auto name = LUWStringName.GetAsString();
 	std::string predefinedName = GetPredefinedName(firstNameIndex, middleNameIndex, lastNameIndex);
+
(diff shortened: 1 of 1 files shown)


aronwk-aaron:
tested that it works

dGame/UserManager.cpp
```diff
@@ -323,7 +324,7 @@ void UserManager::CreateCharacter(const SystemAddress& sysAddr, Packet* packet)
 	}
 
 	//Now that the name is ok, we can get an objectID from Master:
-	ObjectIDManager::RequestPersistentID([=, this](uint32_t objectID) {
+	ObjectIDManager::RequestPersistentID([=, this](uint32_t objectID) mutable {
```
jadebenn, line 327: Anyway to avoid the mutable specification here? I'm fine with it if there's no other option, but...

aronwk-aaron: no other option, need the objid for the name and that's only gotten inside the lambda

jadebenn (approved):

EmosewaMC (approved):

PR #1454, opened 2024-02-11
fix: Remove hard coded groups for spawners

EmosewaMC:
Tested that the target next to the plunger gun dude now spawns its target without needing to leave the area if it is spawned
Tested that spiders still spawn in Spider Queen battle
Tested that macros work

### DIFF
--- a/dScripts/02_server/Map/AG/NpcCowboyServer.cpp
+++ b/dScripts/02_server/Map/AG/NpcCowboyServer.cpp
@@ -1,6 +1,7 @@
 #include "NpcCowboyServer.h"
 #include "eMissionState.h"
 #include "InventoryComponent.h"
+#include "dZoneManager.h"
 
 void NpcCowboyServer::OnMissionDialogueOK(Entity* self, Entity* target, int missionID, eMissionState missionState) {
 	if (missionID != 1880) {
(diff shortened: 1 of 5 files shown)


aronwk-aaron (approved):
thank you father for cleansing our sins

dScripts/02_server/Map/AG/NpcCowboyServer.cpp
```diff
@@ -23,4 +24,17 @@ void NpcCowboyServer::OnMissionDialogueOK(Entity* self, Entity* target, int miss
...
 	}
+
+	// Next up hide or show the samples based on the mission state
+	int32_t visible = 1;
+	if (missionState == eMissionState::READY_TO_COMPLETE || missionState == eMissionState::COMPLETE_READY_TO_COMPLETE) {
+		visible = 0;
+	}
+
+	auto spawners = Game::zoneManager->GetSpawnersByName("PlungerGunTargets");
+	for (auto* spawner : spawners) {
```
jadebenn, line 35: Nitpick: Does auto* spawner need to be mutable here?

EmosewaMC: ehh i get it but its overkill tbqh

jadebenn: tbf I am very much in the 'const all the things!' school of thought. Shows you whether your access is read/only or not and makes debugging easier. But I won't block if you really don't want to do it.

jadebenn (approved):

dScripts/02_server/Map/AG/NpcCowboyServer.cpp
EmosewaMC: I get the mindset and will keep it in mind

PR #1457, opened 2024-02-11
fix: close trade window when one player ends the trade

EmosewaMC:
Tested that someone closing the trade window now echo's this to the other player


EmosewaMC:
truly the end of an era

aronwk-aaron (approved):

jadebenn (approved):

PR #1458, opened 2024-02-12
chore: Reduce manual cmake include directories

jadebenn:
Testing a way to remove most of the CMake directory include statements. Right now I have it compiling with only the thirdparty directories listed manually in the main cmake file


jadebenn:
Please let me know what you all think about this approach versus our current one.

EmosewaMC (changes requested):
headers do not need to be in sources as they are not compiled.

EmosewaMC:
> Please let me know what you all think about this approach versus our current one.

This approach was already being done in #1387 and that pr is nearly finished just missing some final touches for mac builds

jadebenn:
They are different approaches, and my approach does require the headers to be in the libraries.

jadebenn:
Closed after further discussion. There is a reason that #1387 is doing things the way it does.

PR #1459, opened 2024-02-20
fix: EntityManager incorrectly inverted statement

EmosewaMC:
if we DON'T find it, we want to kill/delete it.  not the other way around where if we find it we try to delete it again.

tested that you no longer crash when trying to login
tested that bird monument issues are fixed

### DIFF
--- a/dGame/EntityManager.cpp
+++ b/dGame/EntityManager.cpp
@@ -575,13 +575,13 @@ void EntityManager::ScheduleForKill(Entity* entity) {
 
 	const auto objectId = entity->GetObjectID();
 
-	if (std::find(m_EntitiesToKill.begin(), m_EntitiesToKill.end(), objectId) != m_EntitiesToKill.end()) {
+	if (std::find(m_EntitiesToKill.begin(), m_EntitiesToKill.end(), objectId) == m_EntitiesToKill.end()) {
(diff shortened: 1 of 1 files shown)


Jettford (approved):
LGTM Mr EmosewaMC

Wincent01 (approved):

Wincent01:
Could these be made into sets?

EmosewaMC:
After some more thought a set wouldn't quite work here due to the nature of the set being sorted on insertion to prevent iterator invalidation and to make sure the set is sorted.  Because of this, we may fall into the following example
entities to delete are objId 2 and 3
start delete loop
delete 2
delete 3 | while destructing 2, a third entity with id 1 is added to be deleted
end delete loop
clear entities to delete

Because the set kept the set sorted, we failed to delete entity 1 and that entity was effectively leaked (not fully lost, but it is unlikely to be on any client and will sit there perpetually on servers taking up ram).  Because of this, I believe a vector is still the correct container here since we can push into the entities to delete container and safely know that the entity will be deleted.  unordered containers are the same except its hashing, so they wouldn't work either.  I did begin to grab some numbers before realizing that front to back iteration isnt possible, especially given the wide array of IDs we know exist (spawner flag, persistent flag, character flag etc.)

PR #1461, opened 2024-02-24
feat: achievement vendor and vendor feedback

aronwk-aaron:
Remove hardcoded achievement vendor logic to now work with any achievements vendor
Check if a player should be able to buy an item from the achievement vendor
Log cheat detection for trying to buy items that can't be bought
Better feedback when trying to buy or sell something and can't

Tested that buying from normal vendors and achievement vendors works


aronwk-aaron:
Lot 13732 (TEST COMMENDATION VENDOR) now works since it's no longer hardcoded to be 13806

dDatabase/CDClientDatabase/CDClientTables/CDMissionsTable.cpp
```diff
@@ -79,7 +79,7 @@ void CDMissionsTable::LoadValuesFromDatabase() {
 		entries.push_back(entry);
 		tableData.nextRow();
 	}
-
+	std::ranges::sort(entries, [](const CDMissions& lhs, const CDMissions& rhs) { return lhs.id < rhs.id; });
```
Xiphoseer, line 82: Yknow, CDClient is sorted by ID if we were to start mmapping it.

dDatabase/CDClientDatabase/CDClientTables/CDMissionsTable.cpp
```diff
@@ -118,3 +118,25 @@ const CDMissions& CDMissionsTable::GetByMissionID(uint32_t missionID, bool& foun
 	return Default;
 }
 
+const std::set<int32_t> CDMissionsTable::GetMissionsForReward(LOT lot) {
+	std::set<int32_t> toReturn {};
+	for (const auto& entry : GetEntries()) {
```
Xiphoseer, line 123: This seems expensive

Xiphoseer (approved):
Lgtm

dDatabase/CDClientDatabase/CDClientTables/CDMissionsTable.cpp
EmosewaMC: was left over from a previous thing we were doing to speed up the searches.  Could be removed now mr aron

dDatabase/CDClientDatabase/CDClientTables/CDMissionsTable.cpp
aronwk-aaron: added a cache for this... it is indeed expensive, but this is also what the client does on every load of the UI for the achievement vendor, but instead we are just doing it on buy

Xiphoseer (approved):

jadebenn (approved):

PR #1469, opened 2024-02-25
chore: Convert LeaderboardManager to use BitStream refs

jadebenn:
Pointer reduction


aronwk-aaron (approved):

EmosewaMC (approved):
If this was covered in your other testing, can you just drop that comment here as well so its documented what was tested

jadebenn:
Merge is pending test with minigame.

jadebenn:
Leaderboards look good.

PR #1471, opened 2024-02-25
chore: Convert DoClientProjectileImpact to use bitstream refs

jadebenn:
raw pointer reduction


EmosewaMC (approved):
If this was covered in your other testing, can you just drop that comment here as well so its documented what was tested

jadebenn:
Does the Space Ranger kit weapon count as a projectile? I'll try and test with another gun if it doesn't.

jadebenn:
confirmed that guns still do the shooty-hurty

PR #1476, opened 2024-02-25
chore: Change dServer and related code to use BitStream references

jadebenn:
This is the last of the changes as far as I can tell. I will be testing the combined reference changes in game and will report back on their status.

### DIFF
--- a/dChatServer/ChatIgnoreList.cpp
+++ b/dChatServer/ChatIgnoreList.cpp
@@ -59,7 +59,7 @@ void ChatIgnoreList::GetIgnoreList(Packet* packet) {
 		bitStream.Write(LUWString(ignoredPlayer.playerName, 36));
 	}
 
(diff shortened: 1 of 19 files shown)


jadebenn:
Combined all the bitstream reference branches into one local branch and ran the game for about an hour where I went from AG to NS to GF following the main questline. Seems stable to me.

aronwk-aaron (approved):

EmosewaMC:
Can you also check the ignore list functionality (add, remove and fetch on world load)?  doesnt sound like that was checked

EmosewaMC (approved):
pending the above test request

jadebenn:
Yeah, I'll do that when I have time later today.

jadebenn:
tested by blocking an alt, reloading, and seeing they were still blocked

PR #1479, opened 2024-02-26
chore: Change AMF to use references

jadebenn:
Yet another feature punched out from PR #1452. Part of the current pointer reduction obsess- er, I mean, push.


jadebenn (commented):
ah, since I merged in the branch to get the changes, looks like I brought the history along with it. let me try and squash that...
nope. no good way to. oh well.

dGame/dPropertyBehaviors/ControlBehaviors.h
```diff
@@ -19,15 +19,15 @@ class SystemAddress;
...
 		: arguments{ args }, modelComponent{ modelComponent }, modelOwner{ modelOwner } {
 	};
 
 	operator bool() const {
-		return arguments != nullptr && modelComponent != nullptr && modelOwner != nullptr;
+		return modelComponent != nullptr && modelOwner != nullptr;
 	}
 
-	AMFArrayValue* arguments;
+	std::reference_wrapper<const AMFArrayValue> arguments;
```
EmosewaMC, line 30: there a particular reason this cant be a reference?  the initializer list above should allow this correct?

EmosewaMC (dismissed):
1 question on the functional object

dGame/dPropertyBehaviors/ControlBehaviors.h
jadebenn: Using an actual const reference would make the struct non-moveable, only copyable, and a use of reference wrapper here doesn't require "get()," so it essentially has the semantic upside of representing non-nullability without any of the downsides of an actual reference

EmosewaMC: I think its fine to have this be non move able.  This should be created once per request anyways

jadebenn: Why change it? We're not even using .get() here. It's a perfect use case.

EmosewaMC: I'll trust your judgment here in that case then

aronwk-aaron (approved):

EmosewaMC (approved):

PR #1480, opened 2024-02-26
chore: remove extra speed class

EmosewaMC:
Speed is used in more waypoints than not so we may as well reduce repeated references.

tested that the data is still loaded as normal in avant gardens


aronwk-aaron (approved):

jadebenn (approved):
Lgtm

PR #1481, opened 2024-02-26
fix: FrameStats serialization

EmosewaMC:
The only flags are if any data in the FrameStats has changed, you write them again.  Velocities also do not use dirty flags for their values, they use a flag to determine if their velocity if zero or non-zero.  If any velocity changes, re-write FrameStats. TeleportFrameStats also have not been written correctly and only hasn't caused issues because we never undirtied the FrameStats.  This write has been moved to the correct location, dirty flags are cleared and no unserialize issues are arising in the client.

Tested that 2 players can see each other move as before, enemies move as before and players racing is identical as before.


aronwk-aaron (approved):

jadebenn (dismissed):
Lgtm

Jettford (dismissed):

aronwk-aaron:
merge conflicts :(

aronwk-aaron (approved):

jadebenn (approved):

PR #1486, opened 2024-03-02
fix: plaques would crash when interacting with no data provided

aronwk-aaron:
Also docs

tested that normal plaques work
tested that ones `/spawn`'d in don't crash when interacting anymore
tested that DLU vanity plaques work

Feel free to give feedback and criticism on the docs as I wasn't exactly sure what all to write, but hopefully that covers most use cases and we can adapt as user have questions or confusion.


jadebenn (approved):

jadebenn:
Lgtm

Jettford (approved):

EmosewaMC (approved):

PR #1490, opened 2024-03-03
chore: Set default symbol visibility to hidden in CMAKE

jadebenn:
Change the default C++ visibility to hidden, since we aren't using them at all. This shrinks WorldServer by about 0.8 MB without any impact to functionality. Tested by moving around Gnarled Forest for about 30 minutes, completing side missions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,8 @@ include(CTest)
 
 set(CMAKE_CXX_STANDARD 20)
 set(CXX_STANDARD_REQUIRED ON)
+set(CMAKE_POLICY_DEFAULT_CMP0063 NEW)
+set(CMAKE_CXX_VISIBILITY_PRESET hidden)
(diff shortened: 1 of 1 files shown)


jadebenn:
Before change:
image

After change:
image

aronwk-aaron:
Will want to get the offer cmake PR I think. To avoid more merge conflict. But looks good otherwise

CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ include(CTest)
 
 set(CMAKE_CXX_STANDARD 20)
 set(CXX_STANDARD_REQUIRED ON)
+set(CMAKE_POLICY_DEFAULT_CMP0063 NEW)
```
EmosewaMC, line 7: Please leave a comment here kn what this means since I know we will forget this immediately, waiting for other pr in any case.

jadebenn:
Looks like it can still automerge.

aronwk-aaron (approved):

PR #1491, opened 2024-03-04
feat: Provide SerializeEntity constant reference overload

jadebenn:
Add a constant reference override to SerializeEntity


EmosewaMC (approved):
i mean sure

aronwk-aaron (approved):

PR #1492, opened 2024-03-04
chore: Table Loading Improvements

jadebenn:
Make more use of emplace() in some of the CDClient tables


dDatabase/CDClientDatabase/CDClientTables/CDLootTableTable.cpp
```diff
@@ -79,10 +79,10 @@ const LootTableEntries& CDLootTableTable::GetTable(uint32_t tableId) {
 
 	while (!tableData.eof()) {
 		CDLootTable entry;
-		entries[tableId].push_back(ReadRow(tableData));
+		entries.at(tableId).emplace_back(ReadRow(tableData));
```
EmosewaMC, line 82: This is already a container with safe subscripting.  At is not necessary here and should be reverted.

jadebenn: Are we intentionally keeping the possibility of it creating a new entry here?

EmosewaMC: yes, that is the whole point of this and why i wrote it like that.

dDatabase/CDClientDatabase/CDClientTables/CDMissionTasksTable.h
```diff
@@ -25,7 +27,7 @@ class CDMissionTasksTable : public CDTable<CDMissionTasksTable, std::vector<CDMi
 	// Queries the table with a custom "where" clause
 	std::vector<CDMissionTasks> Query(std::function<bool(CDMissionTasks)> predicate);
 
-	std::vector<CDMissionTasks*> GetByMissionID(uint32_t missionID);
+	std::vector<CDMissionTasks*> GetByMissionID(const uint32_t missionID);
```
EmosewaMC, line 30: The corresponding function in the cpp file was left unchanged and no compiler error was detected.  This function can be removed and re-added if necessary.

dDatabase/CDClientDatabase/CDClientTables/CDObjectsTable.cpp
```diff
@@ -1,7 +1,7 @@
 #include "CDObjectsTable.h"
 
 namespace {
-	CDObjects m_default;
+	CDObjects objDefault;
```
EmosewaMC, line 4: file scope should be PascalCase.  Helps to discern what is a global and what isnt.  only reason this wasnt is because I was editing too many files at that time to warrant variable name changes.

dGame/dComponents/PetComponent.cpp
```diff
@@ -70,8 +70,8 @@ std::map<LOT, int32_t> PetComponent::petFlags = {
 		{ 13067, 838 }, // Skeleton dragon
 };
 
-PetComponent::PetComponent(Entity* parentEntity, uint32_t componentId) : Component{ parentEntity } {
-	m_PetInfo = CDClientManager::GetTable<CDPetComponentTable>()->GetByID(componentId); // TODO: Make reference when safe
+PetComponent::PetComponent(Entity* parentEntity, uint32_t componentId) : Component{ parentEntity },
+	m_PetInfo{ CDClientManager::GetTable<CDPetComponentTable>()->GetByID(componentId) } {
```
EmosewaMC, line 74: Please pull out to another PR.  On this note, the object in question was not updated to be a reference which should be done as well.

dDatabase/CDClientDatabase/CDClientTables/CDPackageComponentTable.cpp
```diff
@@ -19,12 +19,11 @@ void CDPackageComponentTable::LoadValuesFromDatabase() {
 	// Now get the data
 	auto tableData = CDClientDatabase::ExecuteQuery("SELECT * FROM PackageComponent");
 	while (!tableData.eof()) {
-		CDPackageComponent entry;
+		auto &entry = entries.emplace_back();
```
EmosewaMC, line 22: format: we use left aligned ref and ptr.

dDatabase/CDClientDatabase/CDClientTables/CDPhysicsComponentTable.cpp
```diff
@@ -4,32 +4,31 @@ void CDPhysicsComponentTable::LoadValuesFromDatabase() {
...
+		UNUSED_ENTRY(entry.rotSpeed = tableData.getFloatField("rotSpeed", -1),)
+		entry.playerHeight = static_cast<float>(tableData.getFloatField("playerHeight")),
+		entry.playerRadius = static_cast<float>(tableData.getFloatField("playerRadius")),
+		entry.pcShapeType = tableData.getIntField("pcShapeType"),
+		entry.collisionGroup = tableData.getIntField("collisionGroup"),
+		UNUSED_ENTRY(entry.airSpeed = tableData.getFloatField("airSpeed"),)
+		UNUSED_ENTRY(entry.boundaryAsset = tableData.getStringField("boundaryAsset"),)
+		UNUSED_ENTRY(entry.jumpAirSpeed = tableData.getFloatField("jumpAirSpeed"),)
+		UNUSED_ENTRY(entry.friction = tableData.getFloatField("friction"),)
+		UNUSED_ENTRY(entry.gravityVolumeAsset = tableData.getStringField("gravityVolumeAsset"),)
```
EmosewaMC, line 25: why was every semi colon replaced with a comma?

EmosewaMC (changes requested):
good improvements to construction of objects in containers.  As you know unnecessary copies are good to remove.  The pet component changes can be moved to their own pr (small enough to anyways and arent really related to any code changes here) and the .at revert is necessary as the table was complete before as far as functionality goes, and this change will cause crashes when they are not necessary.

dDatabase/CDClientDatabase/CDClientTables/CDPhysicsComponentTable.cpp
jadebenn: Genuinely unsure what happened there

dDatabase/CDClientDatabase/CDClientTables/CDPackageComponentTable.cpp
jadebenn: My spacebar evidently is a C purist 😅

dGame/dComponents/PetComponent.cpp
jadebenn: I no longer want to include a reference member due to it inhibiting move constructors and reducing data locality.

dDatabase/CDClientDatabase/CDClientTables/CDObjectsTable.cpp
jadebenn: Not sure I agree with that formatting decision. Isn't it much more important to preserve the clear distinction between functions and objects than to convey the file scope? One is a very easy to catch compiler error (if not quickly found by the red squiggles in an IDE). The other could introduce very subtle and hard to diagnose bugs if a PascalCase variable was somehow used in a context where a function or object was expected instead.

In particular, I really don't like how your proposed formatting rule would interact with constructor calls. ObjName objName{}; is much clearer than ObjName ObjName{}; If you use parenthesis, the latter might even look like a function call: ObjName ObjName().

EmosewaMC: Resolution agreed upon is f_ for file scope as per discord conversation

dDatabase/CDClientDatabase/CDClientTables/CDPhysicsComponentTable.cpp
jadebenn: I actually remember now: These used to be in an aggregate constructor. Guess I forgot to update the line endings.

jadebenn:
Updated to account for feedback

jadebenn:
Discovered an issue with this branch (#1493). Recommend holding off on review for now.

jadebenn (commented):
Well. That was faster than I expected.

dDatabase/CDClientDatabase/CDClientTables/CDObjectsTable.cpp
```diff
@@ -57,14 +56,16 @@ const CDObjects& CDObjectsTable::GetByID(uint32_t LOT) {
...
-		return m_default;
+		entries.emplace(LOT, ObjDefault);
+		return ObjDefault;
 	}
 
 	// Now get the data
 	while (!tableData.eof()) {
-		CDObjects entry;
-		entry.id = tableData.getIntField("id", -1);
+		const uint32_t LOT = tableData.getIntField("id", 0);
```
EmosewaMC, line 65: LOT is a typedef, not sure how compilers handle a variable having the same name as a typedef

dDatabase/CDClientDatabase/CDClientTables/CDPhysicsComponentTable.cpp
```diff
@@ -4,32 +4,31 @@ void CDPhysicsComponentTable::LoadValuesFromDatabase() {
...
-		entry.speed = tableData.getFloatField("speed", -1);
-		UNUSED(entry->rotSpeed = tableData.getFloatField("rotSpeed", -1));
-		entry.playerHeight = tableData.getFloatField("playerHeight");
-		entry.playerRadius = tableData.getFloatField("playerRadius");
+		UNUSED_COLUMN(entry.jump = tableData.getIntField("jump", -1) != 0;)
+		UNUSED_COLUMN(entry.doubleJump = tableData.getIntField("doublejump", -1) != 0;)
+		entry.speed = static_cast<float>(tableData.getFloatField("speed", -1));
+		UNUSED_COLUMN(entry.rotSpeed = tableData.getFloatField("rotSpeed", -1);)
+		entry.playerHeight = static_cast<float>(tableData.getFloatField("playerHeight"));
+		entry.playerRadius = static_cast<float>(tableData.getFloatField("playerRadius"));
```
EmosewaMC, line 18: These are already floats in the entry struct, not sure why we're casting it to the type that is already returned here

EmosewaMC (approved):

dDatabase/CDClientDatabase/CDClientTables/CDPhysicsComponentTable.cpp
EmosewaMC: and that is of course because getFloatField does not return a float but a double.  Very cool api

dDatabase/CDClientDatabase/CDClientTables/CDObjectsTable.cpp
jadebenn: Seems the typedef gets shadowed, but I see your point.

aronwk-aaron (approved):

EmosewaMC (approved):

PR #1494, opened 2024-03-06
chore: Less verbose name for enum underlying type casts

jadebenn:
Having actually needed to use this helper function fairly extensively for a commit on the PetFixes branch, I propose switching it to use a slightly shorter name that's still descriptive

### DIFF
--- a/dNet/AuthPackets.cpp
+++ b/dNet/AuthPackets.cpp
@@ -229,7 +229,7 @@ void AuthPackets::SendLoginResponse(dServer* server, const SystemAddress& sysAdd
 	RakNet::BitStream loginResponse;
 	BitStreamUtils::WriteHeader(loginResponse, eConnectionType::CLIENT, eClientMessageType::LOGIN_RESPONSE);
 
-	loginResponse.Write<uint8_t>(GeneralUtils::CastUnderlyingType(responseCode));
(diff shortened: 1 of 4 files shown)


EmosewaMC (dismissed):

dNet/AuthPackets.cpp
```diff
@@ -229,7 +229,7 @@ void AuthPackets::SendLoginResponse(dServer* server, const SystemAddress& sysAdd
 	RakNet::BitStream loginResponse;
 	BitStreamUtils::WriteHeader(loginResponse, eConnectionType::CLIENT, eClientMessageType::LOGIN_RESPONSE);
 
-	loginResponse.Write<uint8_t>(GeneralUtils::CastUnderlyingType(responseCode));
```
EmosewaMC, line 232: Is there a particular reason this is cast in the first place?  the enum should be the correct base type already no?

aronwk-aaron (dismissed):

dNet/AuthPackets.cpp
jadebenn: You're right. Looks like the call is redundant here.

aronwk-aaron (approved):

EmosewaMC (approved):

EmosewaMC:
Have you tested that you can still login?

jadebenn:
Logins work normally

PR #1497, opened 2024-03-06
fix: commendation vendor cant accept missions

EmosewaMC:
We never set the HasStandardCost flag in the AchievementVendor.

"Why are you calling RefreshVendor again in the most derived class constructor?  Wont the one in VendorComponent call it?"
No because at that point in time, the vftable for the AchievementVendor has not been assigned yet and as such a call to the virtual function is impossible.  We do need the function to be virtual to prevent the regular refresh from going through as the work is unnecessary for this entity

Tested that you can complete missions from the AchievementVendor and still buy still.

### DIFF
--- a/dGame/dComponents/AchievementVendorComponent.cpp
+++ b/dGame/dComponents/AchievementVendorComponent.cpp
@@ -9,11 +9,15 @@
 #include "UserManager.h"
 #include "CDMissionsTable.h"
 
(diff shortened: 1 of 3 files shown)


aronwk-aaron (approved):
Ty

jadebenn (approved):

PR #1500, opened 2024-03-07
fix: plaques not giving progress

EmosewaMC:
tested that a plaque in nimbus station now loses its blue face when interacted with.


aronwk-aaron (approved):
whoops

jadebenn (approved):

PR #1501, opened 2024-03-10
fix: update serialization, remove unused variable

EmosewaMC:
Tested that players show up as normal on each others screens, tested that money magnet still works with item 8600, tested that gravity still works in Moon Base.
This should also resolve an issue (if one exists) where sometimes gravity or money magnet would not be active for a specific player based on the order they loaded in.  Now the correct data will be sent every time.


aronwk-aaron (approved):

EmosewaMC:
thanks github ci

EmosewaMC:
also tested that enemies in gnarled forest still move in their normal patterns and cannot climb up walls

jadebenn (approved):
Lgtm

PR #1502, opened 2024-03-14
chore: supress warnings on external library headers and actually get rid of the last old-style casts

jadebenn:
adding the 'SYSTEM' keyword to subdirectories is actually a relatively new CMake feature from what I gather, so it might be worthwhile to discuss bumping up the required version in the main CMakeLists.txt as I'm not sure how this would interact with older versions


EmosewaMC:
Include directories has supported SYSTEM syntax in this manner since cmake 3.16, which is the oldest version wsl 20.04 uses which isn't a platform which builds dlu anymore.  All the SYSTEM marker does is say to search the angle beacket include path first instead of quote path.

jadebenn:
I couldn't figure out how to get the SYSTEM flag consistently applied to the the dependencies. If there's a way to do that, it would avoid the issue.

CMakeLists.txt
```diff
@@ -213,29 +213,29 @@ if (APPLE)
 endif()
 
 # Load all of our third party directories
-add_subdirectory(thirdparty)
+add_subdirectory(thirdparty SYSTEM)
```
EmosewaMC, line 216: Since this line requires cmake version 3.25, regardless of if this is merged or not, i believe the minimum required version should be updated to require 3.25 to ensure no weird build issues occur on older clients

jadebenn: Yeah. I can't think of another way to accomplish this, so I've bumped up the required version.

aronwk-aaron (approved):

EmosewaMC:
please update the readme as well

aronwk-aaron (approved):

EmosewaMC (approved):
Keep this pr in mind for issues relating to out of date cmake builds

PR #1504, opened 2024-03-24
chore: Remove `new`s in Behavior members

EmosewaMC:
Could remove all the this-> since everything is prefixed with m_, but meh, too much change for this pr, getting rid of the news is a better fix

Tested that GrowingFlowers still have their SkillEvent fired with the correct parameters, gftikitorch works, sharks eating stinky fish still work

### DIFF
--- a/dGame/dBehaviors/Behavior.h
+++ b/dGame/dBehaviors/Behavior.h
@@ -41,9 +41,9 @@ public:
 	uint32_t m_behaviorId;
 	BehaviorTemplates m_templateId;
 	uint32_t m_effectId;
-	std::string* m_effectHandle = nullptr;
-	std::unordered_map<std::string, std::string>* m_effectNames = nullptr;
-	std::string* m_effectType = nullptr;
+	std::string m_effectHandle;
+	std::unordered_map<std::string, std::string> m_effectNames;
+	std::string m_effectType;
 
 	/*
 	 * Behavior parameters
@@ -88,5 +88,5 @@ public:
 	 */
 
 	explicit Behavior(uint32_t behaviorId);
-	virtual ~Behavior();
+	virtual ~Behavior() = default;
 };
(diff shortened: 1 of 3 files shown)


dGame/dBehaviors/Behavior.h
```diff
@@ -88,5 +88,5 @@ class Behavior
 	 */
 
 	explicit Behavior(uint32_t behaviorId);
-	virtual ~Behavior();
+	virtual ~Behavior() = default;
```
jadebenn, line 91: If the destructor is being defaulted, it should just be removed in order to avoid the weird side effects of having an explicitly defined destructor

jadebenn: Ah. Right. It's virtual. Hm... I need to check if a defaulted destructor needs to be declared virtual or not. I think you might actually be right to keep it in this case.

jadebenn: It might be worthwhile to explicitly define the move constructors as defaulted so they aren't implicitly deleted, then. I don't know if we actually make use of any move optimizations here, but some classes will, like vector, and it's good to future-proof even if we don't.

aronwk-aaron (dismissed):

dGame/dBehaviors/Behavior.h
EmosewaMC: Behaviors themselves are never moved or copied, they are always a pointer, a copy or move could be defined but they will never be used since the ptr will be copied, not the data.  Just going foe the low hanging fruits here since they didn't require much modification (it's very clear to me this was c# code originally ported to c++)

jadebenn: I went ahead and did it myself then. Feel free to revert if you think it shouldn't be there.

jadebenn (approved):

dGame/dBehaviors/Behavior.h
EmosewaMC: im not going to use time reverting them, but those will never be used because Behaviors are never used in that manner as I stated above.  whether those are defined or not will make zero difference since they wont be used

aronwk-aaron (approved):

PR #1507, opened 2024-03-25
fix: missing stack trace function definitions

EmosewaMC:
fixes an issue where the stack traces no longer had any information.  The space savings came from not having the information there which is necessary for stack traces.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,8 +4,6 @@ include(CTest)
 
 set(CMAKE_CXX_STANDARD 20)
 set(CXX_STANDARD_REQUIRED ON)
-set(CMAKE_POLICY_DEFAULT_CMP0063 NEW) # Set CMAKE visibility policy to NEW on project and subprojects
-set(CMAKE_CXX_VISIBILITY_PRESET hidden) # Set C++ symbol visibility to default to hidden
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 
 # Read variables from file


aronwk-aaron (approved):

jadebenn:
I think this is using a hammer instead of a scapel. There should be a way to generate the symbols we need for traces without generating symbols for every single function (including templates and inlines) by not suppressing symbol visibility. Also, MSVC defaults to hidden symbols, so if this is an issue, this isn't a portable solution.

One question I have as well is whether we intend for stack traces to be accessible outside debug builds. If we do, then fixing this makes sense. If not, then I think it might make more sense to just use a debug configuration for development.

jadebenn (changes requested):
I think this is using a hammer instead of a scapel. There should be a way to generate the symbols we need for traces without generating symbols for every single function (including templates and inlines) by not suppressing symbol visibility. Also, MSVC defaults to hidden symbols, so if this is an issue, this isn't a portable solution.

One question I have as well is whether we intend for stack traces to be accessible outside debug builds. If we do, then fixing this makes sense. If not, then I think it might make more sense to just use a debug configuration for development.

jadebenn:
One thing I would try prior to this is explicitly enabling debug symbol generation with -g. I _think_ that should restore stack traces while minimizing bloat to the library.

EmosewaMC:
the previous change broke stack trace info and is preventing stack traces on non-msvc crashes.  If you want the optimziation then you can look into it, but i will not be and just want working stack traces and the small binary size increase is insignficant when compared to a functional feature.

jadebenn:
I strongly disagree that the solution here is to just rip out the change. When we encounter other bugs, we don't just rip out the offending function without attempting to solve the issue.

There are a few questions I have:

1. Since we already know that the code isn't portable, do we need this in release builds? If not, why is the solution here removal?
2. Have you tried setting the visibility to hidden-inline as an interim measure?
3. Have you tried generating debug symbols?

EmosewaMC:
> I strongly disagree that the solution here is to just rip out the change. When we encounter other bugs, we don't just rip out the offending function without attempting to solve the issue.
> 
> There are a few questions I have:
> 
> 1. Since we already know that the code isn't portable, do we need this in release builds? If not, why is the solution here removal?
> 2. Have you tried setting the visibility to hidden-inline as an interim measure?
> 3. Have you tried generating debug symbols?

No, no and no because I don't want to work in cmake.  Last time i broke something the code was reverted without giving me time to look at the issue to give the person with more knowledge on the feature time to work on it

jadebenn:
Sorry things got heated earlier. I've got a fix going up now that partially rolls back the change I made, and I've tested that it restores trace functionality.

PR #1510, opened 2024-03-25
feat: refactor slash commands system into more scalable system

aronwk-aaron:
Spot checked a several of the commands.
Things still work and there is a new help command.
A bulk of the changes are breaking up the big if-else-if block into functions and registering each command.

Future PR will break out each set of command into it's own file. (dev, gm zero, and gm greater than zero command)


aronwk-aaron:
Note to self: fix failing test

aronwk-aaron:
> Note to self: fix failing test

Thanks emo

aronwk-aaron:
OKAY, this is the final form of this PR, I need to check some things with the help command not displaying correctly in-game that may be due to some char limit of that message, but that's a tangent. 
General feedback and suggestions welcome. Emo help me with this impl, so I may need to refer to him for some questions

EmosewaMC:
Declining to review since I contributed to the pr and am biased.  I will keep cleaning it up and helping work on it though

EmosewaMC:
text prints correctly and properly utilizes the asterisk found in reverse engineering.  Its used for denoting where a localized string starts and ends.

EmosewaMC:
afaict, we are sending the correct data, however whats being printed is not correct.  Will have to debug further to figure out the issue.

EmosewaMC:
tested so far
gmlevel 0, 9, 7, 9, 0 sets my gm level to the correct value
makegm and setgmlevel both set the gm level as above
togglenameplate is locked out when disabled still, gm bypasses.
reloadconfig works

EmosewaMC (approved):
finished a code passthrough.  Don't see anything missed on my end, but will not be explicitly approving as I contributed to the PR.

EmosewaMC (commented):
finished a code passthrough. Don't see anything missed on my end, but will not be explicitly approving as I contributed to the PR.

DarwinAnim8or (approved):
props for actually undertaking this :')

dWorldServer/WorldServer.cpp
```diff
@@ -313,6 +314,9 @@ int main(int argc, char** argv) {
 	uint32_t sqlPingTime = 10 * 60 * currentFramerate; // 10 minutes in frames
 	uint32_t emptyShutdownTime = (cloneID == 0 ? 30 : 5) * 60 * currentFramerate; // 30 minutes for main worlds, 5 for all others.
 
+	// Register slash commands
+	SlashCommandHandler::Startup();
```
Jettford, line 318: Genuine like, insane nit pick, this in theory doesn't need to be done on zone 0 worlds servers, so if you are going to make some changes, maybe just put a `if (worldId != 0)` at the start of this line, but really doesn't matter if you don't.

Jettford (approved):
One tiny note, completely non-blocking (presuming it isn't a dumb comment to begin with). Awesome stuff other than that.

dWorldServer/WorldServer.cpp
aronwk-aaron: I'm going to do a follow up PR and break out SlashCommand.cpp into multiple files. I'll add this in that PR

PR #1514, opened 2024-03-26
chore: Hide C symbol visibility

jadebenn:
Update C symbol visibility to default to hidden as well. Marking as draft for now as while this unifies the default symbol behaviors of gcc, clang, and MSVC (as MSVC defaults to hidden), I'm not actually sure if we have enough C code in the project for this to affect anything and so need to check if there's any difference in the compiled binaries.


jadebenn:
Taking this down due to the recently-discovered issues.

PR #1516, opened 2024-03-27
chore: unbreak the stacktraces

### LINKED ISSUE #1515: BUG: crash dumps no longer contain function definitions

### DarkflameServer Version

Main

### Issue Description

Title.  Crash dumps used to contain function definitions but they were removed when visibility was standardized.  

### Reproduction Steps

Crash the server with /crash
Observe there are no function names

### Expected Behavior

Function names from the stack exist

### Environment

Non-windows builds

jadebenn:
Fixes #1515 (which I broke) by partially undoing the symbol visibility changes to only hide inlined functions. Also, CMake now dumps the compile commands into compile_commands.json because that was useful to make sure the correct flags were being applied here and it would probably be useful enough to keep it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,8 +4,9 @@ include(CTest)
 
 set(CMAKE_CXX_STANDARD 20)
 set(CXX_STANDARD_REQUIRED ON)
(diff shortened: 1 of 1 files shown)


aronwk-aaron (approved):

Xiphoseer (approved):
Lgtm

EmosewaMC (commented):
+1

PR #1521, opened 2024-03-30
fix: stewblaster stopping for non-players

EmosewaMC:
fixes an issue when stew blaster would stop for non-players and would stand still permanently due to enemy hitboxes being removed.  The multiple stoppings from players would also cause unnecessary work to be done, so checks for only being paused or unpaused for pausing or resuming movement were added.

Tested that stewblaster only stops for players and starts moving when there are no players in the vicinity

### DIFF
--- a/dScripts/02_server/Map/AM/WanderingVendor.cpp
+++ b/dScripts/02_server/Map/AM/WanderingVendor.cpp
@@ -1,6 +1,7 @@
 #include "WanderingVendor.h"
 #include "MovementAIComponent.h"
 #include "ProximityMonitorComponent.h"
+#include <ranges>
 
 void WanderingVendor::OnStartup(Entity* self) {
 	auto movementAIComponent = self->GetComponent<MovementAIComponent>();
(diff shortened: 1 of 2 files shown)


aronwk-aaron (commented):
Lgtm

aronwk-aaron (approved):

dScripts/02_server/Map/AM/WanderingVendor.cpp
```diff
@@ -19,7 +20,16 @@ void WanderingVendor::OnProximityUpdate(Entity* self, Entity* entering, std::str
 		if (!proximityMonitorComponent) self->AddComponent<ProximityMonitorComponent>();
 
 		const auto proxObjs = proximityMonitorComponent->GetProximityObjects("playermonitor");
-		if (proxObjs.empty()) self->AddTimer("startWalking", 1.5);
+		bool foundPlayer = false;
+		for (const auto id : proxObjs | std::views::keys) {
```
Wincent01, line 24: new syntax just dropped

Wincent01 (approved):

PR #1523, opened 2024-03-30
fix: update InterruptBehavior to current knowledge

EmosewaMC:
Should be 100% live accurate as far as logic and bitstream reads goes.

Tested with all valiant weapons and crux prime weapons (drops from dragons) that combat does not desync and that the client reports the same level and amount of skill deserialize issues as before.

Would like Wincent or the last person who worked on this part of the behavior system to give this a once over due to that guess... and it dating back to pre-foss DLU.


aronwk-aaron (approved):

Wincent01 (approved):
The original guesses here were based on my old Ghidra work. Good to see them updated by someone more experienced in that department 😅

Lgtm

PR #1529, opened 2024-04-04
chore: Remove dpEntity pointers from collision checking

### LINKED ISSUE #1522: ENH: Remove usage of dpEntity objects being used for collision checks

The only thing the pointer is used for is its ID, as such the dpEntity itself does not need to track the pointer, but rather just the objectID, adding safety to the containers and simplicity to them as well.  

https://github.com/DarkflameUniverse/DarkflameServer/blob/main/dPhysics/dpEntity.h#L83-L85
These three lines should be updated to replace the dpEntity with an ObjectID, with the map being changed to a set of objectIDs.
i.e. `std::vector<LWOOBJID>` is all that is necessary for collision checks.

Could break physics collisions.

asking jeff bezos for an open source copy of Havok

### Additional context

_No response_

jadebenn:
Fixes #1522. Tested by going into AG and checking that the bus raises and lowers in the Sentinel base

### DIFF
--- a/dGame/dComponents/PhantomPhysicsComponent.cpp
+++ b/dGame/dComponents/PhantomPhysicsComponent.cpp
@@ -187,7 +187,7 @@ PhantomPhysicsComponent::PhantomPhysicsComponent(Entity* parent) : PhysicsCompon
 			//add fallback cube:
 			m_dpEntity = new dpEntity(m_Parent->GetObjectID(), 2.0f, 2.0f, 2.0f);
 		}
-	
+
 		m_dpEntity->SetScale(m_Scale);
 		m_dpEntity->SetRotation(m_Rotation);
 		m_dpEntity->SetPosition(m_Position);
@@ -323,14 +323,14 @@ void PhantomPhysicsComponent::Update(float deltaTime) {
 	if (!m_dpEntity) return;
 
 	//Process enter events
-	for (auto en : m_dpEntity->GetNewObjects()) {
+	for (const auto en : m_dpEntity->GetNewObjects()) {
 		if (!en) continue;
-		ApplyCollisionEffect(en->GetObjectID(), m_EffectType, m_DirectionalMultiplier);
-		m_Parent->OnCollisionPhantom(en->GetObjectID());
+		ApplyCollisionEffect(en, m_EffectType, m_DirectionalMultiplier);
+		m_Parent->OnCollisionPhantom(en);
 
 		//If we are a respawn volume, inform the client:
 		if (m_IsRespawnVolume) {
-			auto entity = Game::entityManager->GetEntity(en->GetObjectID());
+			const auto entity = Game::entityManager->GetEntity(en);
 
 			if (entity) {
 				GameMessages::SendPlayerReachedRespawnCheckpoint(entity, m_RespawnPos, m_RespawnRot);
@@ -341,10 +341,10 @@ void PhantomPhysicsComponent::Update(float deltaTime) {
 	}
 
 	//Process exit events
-	for (auto en : m_dpEntity->GetRemovedObjects()) {
+	for (const auto en : m_dpEntity->GetRemovedObjects()) {
 		if (!en) continue;
-		ApplyCollisionEffect(en->GetObjectID(), m_EffectType, 1.0f);
-		m_Parent->OnCollisionLeavePhantom(en->GetObjectID());
+		ApplyCollisionEffect(en, m_EffectType, 1.0f);
+		m_Parent->OnCollisionLeavePhantom(en);
 	}
 }
 
--- a/dGame/dComponents/PhantomPhysicsComponent.cpp
+++ b/dGame/dComponents/PhantomPhysicsComponent.cpp
@@ -330,7 +330,7 @@ void PhantomPhysicsComponent::Update(float deltaTime) {
 
 		//If we are a respawn volume, inform the client:
 		if (m_IsRespawnVolume) {
-			const auto entity = Game::entityManager->GetEntity(en);
+			auto* const entity = Game::entityManager->GetEntity(en);
 
 			if (entity) {
 				GameMessages::SendPlayerReachedRespawnCheckpoint(entity, m_RespawnPos, m_RespawnRot);
--- a/dGame/dComponents/PhantomPhysicsComponent.cpp
+++ b/dGame/dComponents/PhantomPhysicsComponent.cpp
@@ -323,14 +323,13 @@ void PhantomPhysicsComponent::Update(float deltaTime) {
 	if (!m_dpEntity) return;
 
 	//Process enter events
-	for (const auto en : m_dpEntity->GetNewObjects()) {
-		if (!en) continue;
-		ApplyCollisionEffect(en, m_EffectType, m_DirectionalMultiplier);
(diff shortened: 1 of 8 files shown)


dPhysics/dpEntity.cpp
```diff
@@ -76,16 +77,25 @@ void dpEntity::CheckCollision(dpEntity* other) {
...
+		m_CurrentlyCollidingObjects.emplace_back(other->GetObjectID());
+		m_NewObjects.push_back(other->GetObjectID());
 	} else if (!isColliding && wasFound) {
-		m_CurrentlyCollidingObjects.erase(other->GetObjectID());
-		m_RemovedObjects.push_back(other);
+		// Erase object ID from currently colliding objects by swapping and popping vector
+		const auto objIdx = std::distance(m_CurrentlyCollidingObjects.cbegin(), objItr);
+		auto& lastObj = m_CurrentlyCollidingObjects.back();
+		m_CurrentlyCollidingObjects[objIdx] = std::move(lastObj);
+		m_CurrentlyCollidingObjects.pop_back();
```
EmosewaMC, line 96: why if objItr not used?  it already is known to be the element we want to remove and has to exist to get into this if block, so it should be simpler to just do 
```cpp
*objItr = m_CurrentlyCollidingObjects.back();
m_CurrentlyCollidingObjects.pop_back();
```

dPhysics/dpEntity.cpp
```diff
@@ -76,16 +77,25 @@ void dpEntity::CheckCollision(dpEntity* other) {
 		return;
 	}
 
-	bool wasFound = m_CurrentlyCollidingObjects.contains(other->GetObjectID());
-
-	bool isColliding = m_CollisionShape->IsColliding(other->GetShape());
+	const auto objItr = std::find(
```
EmosewaMC, line 80: don't forget about std::ranges::find

dScripts/ai/AG/AgBusDoor.cpp
```diff
@@ -23,13 +23,13 @@ void AgBusDoor::OnProximityUpdate(Entity* self, Entity* entering, std::string na
 	m_Counter = 0;
 	m_OuterCounter = 0;
 
-	for (const auto& pair : proximityMonitorComponent->GetProximityObjects("busDoor")) {
-		auto* entity = Game::entityManager->GetEntity(pair.first);
+	for (const auto id : proximityMonitorComponent->GetProximityObjects("busDoor")) {
+		const auto entity = Game::entityManager->GetEntity(id);
```
EmosewaMC, line 27: we use auto* for deduced pointers

aronwk-aaron (approved):

dPhysics/dpEntity.cpp
jadebenn: I am tired. This is a better solution. I will use it. 👍

dScripts/ai/AG/AgBusDoor.cpp
jadebenn: the const auto* const shall make its return i suppose

aronwk-aaron (dismissed):

dGame/dComponents/PhantomPhysicsComponent.cpp
```diff
@@ -323,14 +323,14 @@ void PhantomPhysicsComponent::Update(float deltaTime) {
 	if (!m_dpEntity) return;
 
 	//Process enter events
-	for (auto en : m_dpEntity->GetNewObjects()) {
+	for (const auto en : m_dpEntity->GetNewObjects()) {
 		if (!en) continue;
```
Wincent01, line 327: Should never resolve false?

dGame/dComponents/PhantomPhysicsComponent.cpp
```diff
@@ -341,10 +341,10 @@ void PhantomPhysicsComponent::Update(float deltaTime) {
 	}
 
 	//Process exit events
-	for (auto en : m_dpEntity->GetRemovedObjects()) {
+	for (const auto en : m_dpEntity->GetRemovedObjects()) {
 		if (!en) continue;
```
Wincent01, line 345: Same here

dPhysics/dpEntity.h
```diff
@@ -80,7 +80,7 @@ class dpEntity {
 
 	bool m_IsGargantuan = false;
 
-	std::vector<dpEntity*> m_NewObjects;
-	std::vector<dpEntity*> m_RemovedObjects;
-	std::map<LWOOBJID, dpEntity*> m_CurrentlyCollidingObjects;
+	std::vector<LWOOBJID> m_NewObjects;
+	std::vector<LWOOBJID> m_RemovedObjects;
+	std::vector<LWOOBJID> m_CurrentlyCollidingObjects;
```
Wincent01, line 85: This should be a set for O(1)* time.

Wincent01 (changes requested):
Changing currently colliding to a set is required, the other things are nitpicks

dPhysics/dpEntity.h
EmosewaMC: a set is actually slower here.  The only reason I would support a set is to prevent double registration of currently colliding objects.

EmosewaMC: This set is generally very small (really only players are colliding with others, and there are only a few doing so) and because of this, the locality of reference overpowers the lookup speed for a set (unordered_set is the O(1) set which I think you meant to suggest, set in c++ is ordered and O(log base2 n)) in the grand scheme of things we wont really notice a difference, but with small containers under around 6-7 elements, a vector search in linear time is faster than a tree search in logn time.  for an unordered set however, they are roughly equal depending on the under the hood implementation of the items (the actual elements may be stored in a vector and provided via an unordered interface).

EmosewaMC: in our context given what a LWOOBJID is and how msvc/gcc/clang implement the hash for a unsigned long long, an unordered_set should be the fastest option here, as you suggested Wincent.  the average number of elements in a collision can be high in populated worlds, so a set is correct here.

Wincent01: Yeah apologizes for the ambiguity. I almost always resort to unordered_set when I talk about sets, almost always what you want.

EmosewaMC: the naming of sets should have been reversed (default set unordered and an ordered_set) but alas a decision made 45 years ago echos to this day :)

jadebenn: Are we actually sure we want it to be unordered too? We are doing a lot of searches, insertions, and removals, aren’t we? Using an ordered set penalizes the insertions and removals but it makes the searches faster.

jadebenn: Also, I am somewhat confused about the benefits of an unordered_set vs the vector approach I was using initially. I was using a "swap and pop" O(1) deletion scheme as well as appending new elements to the back. While finding an element required iterating until a match was found, that iteration was also occurring over continuous memory. Is an unordered_set really more advantageous than that?

EmosewaMC: They are about equal given the hash scheme for a lwoobjid.  When taking into account social worlds, they may have a size of 20+, which is well beyond where a vector locality overpowers the lookup speed

EmosewaMC (approved):

Wincent01 (approved):

PR #1534, opened 2024-04-05
chore: disable non-standard volatile behavior on MSVC

jadebenn:
Discovered this at work the other day. By default, MSVC treats all volatile variables as if they were atomics on x86 and x64 architectures, and requires this use of an obscure compiler flag to disable this.

https://learn.microsoft.com/en-us/cpp/build/reference/volatile-volatile-keyword-interpretation?view=msvc-170

This is, needless to say, very much not portable and it makes me worried it could potentially cause odd behavior between the platforms. While there's no evidence it has at the moment, I'd rather nip this in the bud. This PR adds said obscure compiler flag to CMake and thus brings MSVC inline with gcc and clang's treatment of volatile in accordance to the C++ standard.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,7 +73,8 @@ if(UNIX)
 	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c99 -O2 -fPIC")
 elseif(MSVC)
 	# Skip warning for invalid conversion from size_t to uint32_t for all targets below for now
-	add_compile_options("/wd4267" "/utf-8")
+	# Also disable non-portable MSVC volatile behavior
+	add_compile_options("/wd4267" "/utf-8" "/volatile:iso")
 elseif(WIN32)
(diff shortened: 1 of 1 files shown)


jadebenn:
I still can't quite get the windows build working on my machine, but there's a [compiler warning](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-c4746?view=msvc-170) that should let us know if this change might cause any issues in our windows specific dependencies. Otherwise, this should not affect anything shared across platforms and should in fact remove a potential bug source from play.

jadebenn:
> So to make sure i understand. this would make primitives, like size_t, be consistent across compiles?

No, this just has to do with `volatile` types

EmosewaMC (approved):

Xiphoseer:
I disagree this is a portability concern and see no downside to volatile access being atomic?

EmosewaMC:
> I disagree this is a portability concern and see no downside to volatile access being atomic?

i dont either but dont have the energy to refute something like this

PR #1535, opened 2024-04-07
chore: Use string to access SQLite columns

### LINKED ISSUE #1531: ENH: Replace numeric indexing of SQLite columns with column names

Currently, column indexing is prone to issues with off by one, copy paste and ambiguity.  This has caused issues in the past and has caused issues to this day and will cause issues in the future if a column is added to the query or removed in an update.

Replace any calls of, for example a query with a column `banana` and a getter `getIntField(0)` with the getter `getIntField("banana")`.  there are other similar api calls which should have this done as well.

Binary size will grow but we're talking a few hundred bytes.  Worth the zero error and runtime error if a column doesnt exist in an updated query.

Why are these not just using the ORM in these random queries?  this should be done in the future.  No file should need to link with sqlite.

### Additional context

i ate a carrot today.  tasted good.

EmosewaMC:
~~Untested right now.  Will try to test later.  May do some other small improvements near this code, like removing finalizes and some char* -> string -> copy to data stuff, may not.  idk atm.~~

fixes #1531


Wincent01 (commented):
It probably wouldn't be a measurable improvement, but we could utilize static variables to store a cashed index of the fields.

EmosewaMC:
I was planning on doing more improvements in a separate PR.  this one is mostly just a number->text replacement and ideally any queries to the cdclient are replaced with an ORM table for caching

EmosewaMC:
tested the following
combat in nimbus station with a stromling and nuckal is working and effects by the stromling play
lookup works
having a pirate hat on and transferring to another zone works (nimbus station ->gnarled forest)
set bonuses work (used 6 piece engineer with valiants)
buffs work(equipped ninjago chest and pants and reduced damage by 2)
projectiles still land and do damage
character create still chooses the correct shirts the player chose
switch multiple works (tested with valiant summoner weapon chargeup)
claiming and placing a model works
paradox teleporter in nexus tower has the correct animation time
adding the nuckal mission and smashing them gives the quest item and does not when you do not have the mission
can use mounts still
can dismantle models to bricks in the inventory

EmosewaMC:
lookup still works

aronwk-aaron (dismissed):

dGame/dInventory/ItemSet.cpp
```diff
@@ -27,14 +27,16 @@ ItemSet::ItemSet(const uint32_t id, InventoryComponent* inventoryComponent) {
 		return;
 	}
 
-	for (auto i = 0; i < 5; ++i) {
-		if (result.fieldIsNull(i)) {
+	const std::vector<std::string_view> rowNames = { "skillSetWith2", "skillSetWith3", "skillSetWith4", "skillSetWith5", "skillSetWith6" };
```
jadebenn, line 30: nit: could this be a constexpr std::array?

dGame/dInventory/ItemSet.cpp
```diff
@@ -27,14 +27,16 @@ ItemSet::ItemSet(const uint32_t id, InventoryComponent* inventoryComponent) {
...
+	for (auto i = 0; i < rowNames.size(); ++i) {
```
jadebenn, line 31: nit: could this be a range-for loop?

jadebenn (dismissed):
nitpicks aside, looks good. you can address them if you want, or don't if you don't

dGame/dInventory/ItemSet.cpp
jadebenn: Ah, never-mind. I see why it can't now.

dGame/dInventory/ItemSet.cpp
jadebenn: Changed it to be one.

jadebenn (approved):

jadebenn:
well, frick

jadebenn:
welp, reverting my attempted changes and approving. that is a mystery to solve later.

jadebenn (approved):

jadebenn (approved):

dGame/dInventory/ItemSet.cpp
EmosewaMC: I would appreciate not editing my PRs unless I have asked someone to, or worked together with someone on it.  I did not think those changes were necessary and would have left them alone personally as they would not change anything.

aronwk-aaron (approved):

dGame/dInventory/ItemSet.cpp
jadebenn: I was not aware of that. Will do moving forward.

PR #1536, opened 2024-04-08
fix: ignore whitespace in TryParse

EmosewaMC:
tested that Not the Boss of Me progresses normally (its cdclient entry is `7805,11225,11226, 11217, 11220, 11214, 11982, 11987, 11989, 12005, 11999, 12467, 12468, 12469, 12591, 12600, 12609, 14381,16047,16048,16049,16050`

### DIFF
--- a/dCommon/GeneralUtils.h
+++ b/dCommon/GeneralUtils.h
@@ -156,8 +156,11 @@ namespace GeneralUtils {
 	 * @returns An std::optional containing the desired value if it is equivalent to the string
 	*/
 	template <Numeric T>
-	[[nodiscard]] std::optional<T> TryParse(const std::string_view str) {
+	[[nodiscard]] std::optional<T> TryParse(std::string_view str) {
 		numeric_parse_t<T> result;
+		if (!str.empty()) {
+			while (std::isspace(str.front())) str.remove_prefix(1);
+		}
 
 		const char* const strEnd = str.data() + str.size();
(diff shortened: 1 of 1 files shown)


aronwk-aaron (approved):

Jettford (approved):

jadebenn:
Is there any possibility here that this could overrun the string_view .size()? I think you'd need a pretty choice confluence of circumstances (string made of whitespace, no null termination, following byte matching a whitespace character), but I think it's possible.

EmosewaMC:
> Is there any possibility here that this could overrun the string_view .size()? I think you'd need a pretty choice confluence of circumstances (string made of whitespace, no null termination, following byte matching a whitespace character), but I think it's possible to invoke UB. Would it make sense to add a length check to the loop?

.empty already does this check.

EmosewaMC:
I can see how this may overrun (it wont because null termination would result in false and just return).  Feel free to make a patch with a working check.

EmosewaMC:
> > Is there any possibility here that this could overrun the string_view .size()? I think you'd need a pretty choice confluence of circumstances (string made of whitespace, no null termination, following byte matching a whitespace character), but I think it's possible to invoke UB. Would it make sense to add a length check to the loop?
> 
> .empty already does this check and .front will throw before allowing ub.

correction, front does not throw, which is very different from the behavior I had thought front did.  not sure where I got that information from.

PR #1537, opened 2024-04-08
fix: player jittering when picking up loot

EmosewaMC:
tested that players no longer jitter when picking up items by killing 20 admirals and picking up all their loot at once.


EmosewaMC:
Yes, setting rotation to all zero and not IDENTITY is correct.

aronwk-aaron (approved):

Jettford (approved):

PR #1541, opened 2024-04-09
chore: update check to check size in loop

EmosewaMC:
safer code.  tested that missions still progress


aronwk-aaron (approved):

jadebenn (approved):

jadebenn:
Ah, I get it now. I just tried to do an overrun on godbolt with a view to a non-terminated slice of an underlying string, and the str.remove_prefix() method actually checks that it's within bounds already and will prevent any overruns.

jadebenn:
Disregard my previous (now deleted) comment. It's the empty() check preventing overrun. I didn't understand that. 😅

PR #1544, opened 2024-04-10
fix: Add exception inheritance for CppSQLite3Exception, harder try catch re-throw in exception handler

EmosewaMC:
Allows this to be caught in our exception handler by name.  Fixes the try catch re-throw to catch any exception if the exception re-thrown does not inherit std::exception.

Tested that replacing a dummy query.bind with a non-existent field
- prints `Caught unknown exception.` when not inheriting from std::exception
- prints `SQLITE3_ERROR...` when inheriting from std::exception


aronwk-aaron (approved):

Xiphoseer (approved):
Oh 😅

PR #1545, opened 2024-04-10
feat: showall, findplayer, get/openhttpmoninfo, debug

aronwk-aaron:
`/showall` and `/findplayer` use custom packets and are live accurate `/who` variations

`/gethttpmoninfo` and `/openhttpmoninfo` are planned to be used later, just putting in info since it's been RE-d and don't want it lost

Additionally add debug worldPacket for future dev use

Tested all new commands and variation of command arguments


aronwk-aaron:
image
There is a quirk with httpmoninfo commands that displays the supported pages 3 times. It's been like this the entire time but i am not sure why exactly this is happening. will need further investigation

dChatServer/ChatPacketHandler.cpp
```diff
@@ -354,6 +354,75 @@ void ChatPacketHandler::HandleGMLevelUpdate(Packet* packet) {
...
+	inStream.Read(playerID);
+
+	const auto& sender = Game::playerContainer.GetPlayerData(playerID);
+	if (!sender) return;
+
+	LUWString playerName;
+	inStream.Read(playerName);
+	const auto& player = Game::playerContainer.GetPlayerData(playerName.GetAsString());
+	bool online = true;
+	if (!player) online = false;
```
EmosewaMC, line 370: bool online = player;

dChatServer/ChatPacketHandler.cpp
```diff
@@ -354,6 +354,75 @@ void ChatPacketHandler::HandleGMLevelUpdate(Packet* packet) {
...
+	LWOOBJID playerID;
+	inStream.Read(playerID);
+
+	const auto& sender = Game::playerContainer.GetPlayerData(playerID);
+	if (!sender) return;
+
+	LUString displayZoneData(1);
+	inStream.Read(displayZoneData);
+	LUString displayIndividualPlayers(1);
+	inStream.Read(displayIndividualPlayers);
```
EmosewaMC, line 398: a 1 character LUString?  just read a character in that case?

dChatServer/ChatPacketHandler.cpp
```diff
@@ -354,6 +354,75 @@ void ChatPacketHandler::HandleGMLevelUpdate(Packet* packet) {
...
+
+	CBITSTREAM;
+	BitStreamUtils::WriteHeader(bitStream, eConnectionType::CHAT, eChatMessageType::WORLD_ROUTE_PACKET);
+	bitStream.Write(playerID);
+
+	BitStreamUtils::WriteHeader(bitStream, eConnectionType::CLIENT, eClientMessageType::SHOW_ALL_RESPONSE);
+	bitStream.Write<uint8_t>(displayZoneData.string != "1" && displayIndividualPlayers.string != "1");
```
EmosewaMC, line 405: do these comparisons once and use those booleans instead

dChatServer/ChatPacketHandler.cpp
```diff
@@ -354,6 +354,75 @@ void ChatPacketHandler::HandleGMLevelUpdate(Packet* packet) {
...
+	bitStream.Write(Game::playerContainer.GetPlayerCount());
+	bitStream.Write(Game::playerContainer.GetSimCount());
+	bitStream.Write<uint8_t>(displayIndividualPlayers.string == "1");
+	bitStream.Write<uint8_t>(displayZoneData.string == "1");
+	if (displayZoneData.string == "1" || displayIndividualPlayers.string == "1"){
+		for (auto& [playerID, playerData ]: Game::playerContainer.GetAllPlayers()){
+			if (!playerData) continue;
+			bitStream.Write<uint8_t>(0); // padding
```
EmosewaMC, line 413: Is this truly padding?  Please be more descriptive if this is structure packing filler specifically or if it is just unknown.

dChatServer/PlayerContainer.h
```diff
@@ -71,6 +71,9 @@ class PlayerContainer {
 	const PlayerData& GetPlayerData(const std::string& playerName);
 	PlayerData& GetPlayerDataMutable(const LWOOBJID& playerID);
 	PlayerData& GetPlayerDataMutable(const std::string& playerName);
+	uint32_t GetPlayerCount() { return m_PlayerCount; };
+	uint32_t GetSimCount() { return m_SimCount; };
+	std::map<LWOOBJID, PlayerData> GetAllPlayers() { return m_Players; };
```
EmosewaMC, line 76: const ref

dGame/dUtilities/SlashCommands/GMGreaterThanZeroCommands.cpp
```diff
@@ -287,4 +287,29 @@ namespace GMGreaterThanZeroCommands {
 		std::string name = entity->GetCharacter()->GetName() + " - " + args;
 		GameMessages::SendSetName(entity->GetObjectID(), GeneralUtils::UTF8ToUTF16(name), UNASSIGNED_SYSTEM_ADDRESS);
 	}
+
+	void ShowAll(Entity* entity, const SystemAddress& sysAddr, const std::string args) {
+		const auto splitArgs = GeneralUtils::SplitString(args, ' ');
+		CBITSTREAM;
+		BitStreamUtils::WriteHeader(bitStream, eConnectionType::CHAT, eChatMessageType::SHOW_ALL);
+		bitStream.Write(entity->GetObjectID());
+		if (splitArgs.size() > 0 && splitArgs.at(0).size() > 0) bitStream.Write(LUString(splitArgs.at(0), 1));
```
EmosewaMC, line 296: !.empty() and !.empty()

dGame/dUtilities/SlashCommands/GMGreaterThanZeroCommands.cpp
```diff
@@ -287,4 +287,29 @@ namespace GMGreaterThanZeroCommands {
...
+		else bitStream.Write(LUString("1", 1));
+		if (splitArgs.size() > 1) bitStream.Write(LUString(splitArgs.at(1), 1));
+		else bitStream.Write(LUString("1", 1));
+		Game::chatServer->Send(&bitStream, SYSTEM_PRIORITY, RELIABLE, 0, Game::chatSysAddr, false);
```
EmosewaMC, line 300: move this stream writing and probably send to a dNet ChatPackets function

dGame/dUtilities/SlashCommands/GMGreaterThanZeroCommands.cpp
```diff
@@ -287,4 +287,29 @@ namespace GMGreaterThanZeroCommands {
...
+	}
+
+	void FindPlayer(Entity* entity, const SystemAddress& sysAddr, const std::string args) {
+		if (args.size() == 0) {
```
EmosewaMC, line 304: .empty()

dGame/dUtilities/SlashCommands/GMGreaterThanZeroCommands.cpp
```diff
@@ -287,4 +287,29 @@ namespace GMGreaterThanZeroCommands {
...
+			GameMessages::SendSlashCommandFeedbackText(entity, u"No player Given");
+			return;
+		}
+
+		CBITSTREAM;
+		BitStreamUtils::WriteHeader(bitStream, eConnectionType::CHAT, eChatMessageType::WHO);
+		bitStream.Write(entity->GetObjectID());
+		bitStream.Write(LUWString(args));
+		Game::chatServer->Send(&bitStream, SYSTEM_PRIORITY, RELIABLE, 0, Game::chatSysAddr, false);
```
EmosewaMC, line 313: to ChatPackets

dNet/WorldPackets.h
```diff
@@ -10,6 +10,21 @@ struct SystemAddress;
 enum class eGameMasterLevel : uint8_t;
 enum class eCharacterCreationResponse : uint8_t;
 enum class eRenameResponse : uint8_t;
+namespace RakNet {
+	class BitStream;
+};
+
+#pragma pack(push, 1)
```
EmosewaMC, line 17: Why?  The only reason you would use this is so you could write the struct as is to the stream, but you have a serialize anyways, as such this isnt necessary and can be removed.

dNet/WorldPackets.cpp
```diff
@@ -160,3 +169,18 @@ void WorldPackets::SendGMLevelChange(const SystemAddress& sysAddr, bool success,
...
+	BitStreamUtils::WriteHeader(bitStream, eConnectionType::CLIENT, eClientMessageType::HTTP_MONITOR_INFO_RESPONSE);
+	info.Serialize(bitStream);
+	SEND_PACKET;
+}
+
+void WorldPackets::SendDebugOuput(const SystemAddress& sysAddr, const std::string data){
+	CBITSTREAM;
+	BitStreamUtils::WriteHeader(bitStream, eConnectionType::CLIENT, eClientMessageType::DEBUG_OUTPUT);
+	bitStream.Write<uint32_t>(data.size());
+	bitStream.Write(data);
```
EmosewaMC, line 184: don't we overload std::string writing already to do exactly this?

EmosewaMC:
also remember reference (for structures larger than 8 bytes) in function arguments when possible if a copy is not necessary :)

dNet/WorldPackets.cpp
EmosewaMC: only for writing the string, not the length.

EmosewaMC:
> image There is a quirk with httpmoninfo commands that displays the supported pages 3 times. It's been like this the entire time but i am not sure why exactly this is happening. will need further investigation

will look into this

dChatServer/ChatPacketHandler.cpp
aronwk-aaron: Truly padding from when I RE-d this with you. Will double check

dNet/WorldPackets.h
aronwk-aaron: Iirc, you showed me this months ago cause it was cool lol

dChatServer/ChatPacketHandler.cpp
EmosewaMC: lets use structure packing instead then, padding just feels a big vague and structure packing lands a bit more on the why bit

dNet/WorldPackets.h
EmosewaMC: While it is cool (and live accurate) let's go for the safer programmatic approach

dChatServer/ChatPacketHandler.cpp
aronwk-aaron: this was stupid, I just need to read the text in the slash command and write 0 or 1 instead of passing the text along internally and then doing the comparison here, will fix

dChatServer/ChatPacketHandler.cpp
aronwk-aaron: writing bools to the packet will make this unnecessary

dGame/dUtilities/SlashCommands/GMGreaterThanZeroCommands.cpp
aronwk-aaron: having issues moving these to chatpackets, will discuss more in discord

aronwk-aaron: due to the linker issue with chat server, compromised and made (de)serialize functions to contain the data and reading/writing

EmosewaMC (approved):

EmosewaMC:
> > image There is a quirk with httpmoninfo commands that displays the supported pages 3 times. It's been like this the entire time but i am not sure why exactly this is happening. will need further investigation
> 
> will look into this

Confirmed to be a bug in the client.  This bug is not solvable without client mods.  Please utilize another method from the server to communicate the same information from the server.

aronwk-aaron:
>  Please utilize another method from the server to communicate the same information from the server.

Will just have to be how it is, since there is other functionality that does work in the handle for that packet (such as opening the webpage for `/openhttpmoninfo`)
From the discussion in discord, this is a bug in the client. but other than the undefined behavior quirkiness which causes the chat messages to display incorrectly, it's not a big issue?

EmosewaMC:
> > Please utilize another method from the server to communicate the same information from the server.
> 
> Will just use this packet with the bug, since there is other functionality that does work in the handle for that packet (such as opening the webpage for `/openhttpmoninfo`) From the discussion in discord, this is a bug in the client. but other than the undefined behavior quirkiness which causes the chat messages to display incorrectly, it's not a big issue?

Http monitor isn't even implemented, so this is introducing a new bug for zero benefit.  If you implement the http monitor then you can use the packet, aside from that there's no point.  There are probably other ways to get the shell window open anyways that don't involve that gm.

aronwk-aaron:
when ahead and removed the commands for now, will add them back with a workaround for this bug once we actually have the http enpoint

jadebenn (approved):
lgtm

EmosewaMC (approved):

EmosewaMC:
dont forget, there is a way to do exactly the behavior of the httpmon work on the server only to fix the client bug.

aronwk-aaron:
yeet, got it written down

PR #1548, opened 2024-04-11
fix: Unset MAXIMUM_OUTGOING_BANDWIDTH by default in docker-compose

Xiphoseer:


aronwk-aaron (approved):

EmosewaMC (changes requested):
This is a thing i need to help research and figure out why the rubber banding happens when this is unlimited.  Requesting changes just so its marked as not being mergeable for now.

EmosewaMC (approved):

PR #1550, opened 2024-04-21
feat: chat web api mvp

### LINKED ISSUE #450: ENH: Add live player count tracking

No

Add the ability for the number of currently online players to be tracked in the database. This would allow for easier moderation and allow for an external program to grab the current player count for display purposes. 

Having the server update a new column in the `accounts` table called something like `is_online` which would be a boolean 0 for offline and 1 for online. 

Each time a user logs in, their `is_online` would be changed to a `1`
When a user logs off, their `is_online` value would be changed to a `0`

The `accounts` table would need an additional column.

A couple alternatives, 

- Have a seperate table called `online_accounts` that is updated with the accounts that are currently online. (this is inefficient, and not ideal)
- Write the total number of online accounts to a file (downside of not getting who is online, and a bit unelegent having to write to a file)

### Additional context

Briefly going through the code, its already tracked when user logs on/off, see https://github.com/DarkflameUniverse/DarkflameServer/blob/42f6f2f10b5971dd13faa18e2018892ce21ce3c3/dWorldServer/WorldServer.cpp#L800 
and https://github.com/DarkflameUniverse/DarkflameServer/blob/d9d27a88fc3306f89e0cfb72e01de3e1bfe203e9/dGame/UserManager.cpp#L139

aronwk-aaron:
Web API for Chat server Minimum Viable Product

## Features:

- Send announcements in-game
- Get online players
- Get Teams 

resolves #450


aronwk-aaron:
TODO: Add wait for startup check

aronwk-aaron:
Also need to add nlohmann as a 3rd party dep in a separate PR, cmake or header only

dChatServer/PlayerContainer.h
```diff
@@ -7,6 +7,9 @@
 #include "dServer.h"
 #include <unordered_map>
 
+#include <nlohmann/json.hpp>
+using json = nlohmann::json;
```
EmosewaMC, line 11: This should be moved to cpp files.  Using Using in a header is bad practice and pollutes the namespace

dChatServer/PlayerContainer.h
```diff
@@ -88,6 +93,7 @@ class PlayerContainer {
 	LWOOBJID GetId(const std::u16string& playerName);
 	uint32_t GetMaxNumberOfBestFriends() { return m_MaxNumberOfBestFriends; }
 	uint32_t GetMaxNumberOfFriends() { return m_MaxNumberOfFriends; }
+	const std::vector<TeamData*> GetAllTeams() { return mTeams;};
```
EmosewaMC, line 96:
```suggestion
	const std::vector<TeamData*>& GetAllTeams() { return mTeams;};
```

resources/chatconfig.ini
```diff
@@ -6,3 +6,11 @@ max_number_of_best_friends=5
...
 # going over 50 will be allowed in some secnarios, but proper handling will require client modding
 max_number_of_friends=50
+
+# Enable or disable the chat web API, disabled by default
+# It will run on the same port the chat server is running on, defined in shardconfig.ini
+enable_chat_web_api=0
+
+# If that chat web api is enabled, it will only listen for connections on this ip address
+# 127.0.0.1 is localhost
+chat_web_api_listen_address=127.0.0.1
```
EmosewaMC, line 16:
```suggestion
chat_web_api_listen_address=localhost
```
Raknet should translate local host to this ip, more readable than the ip imo.

dChatServer/PlayerContainer.cpp
```diff
@@ -1,7 +1,8 @@
-#include "PlayerContainer.h"
-#include "dNetCommon.h"
 #include <iostream>
 #include <algorithm>
+
+#include "PlayerContainer.h"
```
EmosewaMC, line 4: This should be moved back up.  This files include should always be at the top

dChatServer/ChatWebApi.h
```diff
@@ -0,0 +1,6 @@
+#include "httplib.h"
```
EmosewaMC, line 1: This can be moved to the cpp file.  Nothing in the header uses this.

dChatServer/ChatWebApi.cpp
```diff
@@ -0,0 +1,93 @@
...
+	m_APIServer.Get("/teams", [](const httplib::Request& req, httplib::Response& res) {
+		auto data = json::array();
+		for (auto& teamData: Game::playerContainer.GetAllTeams()){
+			if (!teamData) continue;
+			json toInsert;
+			toInsert["id"] = teamData->teamID;
+			toInsert["loot_flag"] = teamData->lootFlag;
+			toInsert["local"] = teamData->local;
+
+			auto leader = Game::playerContainer.GetPlayerData(teamData->leaderID);
```
EmosewaMC, line 69:
```suggestion
			auto& leader = Game::playerContainer.GetPlayerData(teamData->leaderID);
```

dChatServer/ChatWebApi.cpp
```diff
@@ -0,0 +1,93 @@
...
+			toInsert["leader"] = leader.to_json();
+
+			json members;
+			for (auto& member : teamData->memberIDs){
+				auto playerData = Game::playerContainer.GetPlayerData(member);
```
EmosewaMC, line 74:
```suggestion
				auto& playerData = Game::playerContainer.GetPlayerData(member);
```

dChatServer/ChatWebApi.cpp
```diff
@@ -0,0 +1,93 @@
...
+		}
+		res.set_content(data.dump(), "application/json");
+		if (data.empty()) res.status = 204;
+	});
+
+	m_APIServer.listen(Game::config->GetValue("chat_web_api_listen_address").c_str(), port);
+};
+
+void ChatWebApi::Stop(){
+	if (Game::config->GetValue("enable_chat_web_api") == "1") {
```
EmosewaMC, line 89: Does this check matter?  I'd like to imagine the api server class doesn't allow stop to be called on an already stopped server.

dChatServer/PlayerContainer.h
aronwk-aaron: I have an idea about something to help with this an only need to do this once

resources/chatconfig.ini
aronwk-aaron: httplib needs and IP to listen on, localhost does not work

dChatServer/ChatWebApi.cpp
aronwk-aaron: true, probably doesn't need this check here

dChatServer/PlayerContainer.h
aronwk-aaron: lmk if this solution I have pushed is better

EmosewaMC: this is doing the same thing just moved to a second file

each file that uses json should include the header manually and the cpp file can do `using json = nlohmann::json;`.

EmosewaMC: a #include just takes the contents of the included file and pastes it right there, so moving it to another header just moves the work

EmosewaMC: the purpose of unnamed namespaces is to make it to the contents within are guaranteed to be the ones in this file.  placing an unnamed namespace in a header just creates a new version of its contents in each file that includes it, likely not what was intended

aronwk-aaron: Yeah, that makes sense, honestly don't even really need the using. So I may just not do that at all

EmosewaMC: its fine to use it but only in a cpp file

aronwk-aaron: Would it not be inconsistent between the h and cpp then? I'd rather it be consistent

EmosewaMC: its not inconsistent though, in one you use the fully qualified name to prevent namespace pollution, and in the implementation you can use a shorthand.  for example, one could do
```cpp
namespace f {
    void foo();
}

void f::foo() {};
// or
namespace f {
    void foo() {}
};
```
i see it as perfectly fine and reasonable to use fully qualified name for the return in the header and the shorthand for the return and usage in the implementation.

aronwk-aaron: hmmmm okay

EmosewaMC (dismissed):

aronwk-aaron:
NOTHING IS THREADSAFE sdlfnsdljfhasd;jklfsdljfnsdjklfnosdjklnf

aronwk-aaron:
super seeded by #1715

PR #1553, opened 2024-04-24
chore: Convert heap allocation to optional

jadebenn:
swap out the petcomponent precondition expression for an optional
doing this here because the petfixes PR is already big enough

### DIFF
--- a/dGame/dComponents/PetComponent.cpp
+++ b/dGame/dComponents/PetComponent.cpp
@@ -40,7 +40,7 @@ std::unordered_map<LWOOBJID, LWOOBJID> PetComponent::activePets{};
  * Maps all the pet lots to a flag indicating that the player has caught it. All basic pets have been guessed by ObjID
  * while the faction ones could be checked using their respective missions.
  */
-std::map<LOT, int32_t> PetComponent::petFlags = {
+const std::map<LOT, int32_t> PetComponent::petFlags{
 		{ 3050, 801 },  // Elephant
 		{ 3054, 803 },  // Cat
 		{ 3195, 806 },  // Triceratops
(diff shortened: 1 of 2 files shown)


aronwk-aaron (approved):

dGame/dComponents/PetComponent.cpp
```diff
@@ -158,7 +157,7 @@ void PetComponent::OnUse(Entity* originator) {
 		return;
 	}
 
-	if (m_Preconditions != nullptr && !m_Preconditions->Check(originator, true)) {
+	if (m_Preconditions.has_value() && !m_Preconditions->Check(originator, true)) {
```
EmosewaMC, line 160:
```suggestion
	if (m_Preconditions && !m_Preconditions->Check(originator, true)) {
```
optional should be implicitly convertible to bool so the has_value can be dropped

dGame/dComponents/PetComponent.h
```diff
@@ -349,7 +349,7 @@ class PetComponent final : public Component
 	/**
 	 * Preconditions that need to be met before an entity can tame this pet
 	 */
-	PreconditionExpression* m_Preconditions;
+	std::optional<PreconditionExpression> m_Preconditions = std::nullopt;
```
EmosewaMC, line 352:
```suggestion
	std::optional<PreconditionExpression> m_Preconditions;
```
This should achieve the same result?

jadebenn: If I'm going to default-initialize I think I should actually go with
```suggestion
	std::optional<PreconditionExpression> m_Preconditions{};
```

aronwk-aaron (approved):

dGame/dComponents/PetComponent.cpp
EmosewaMC: The codebase standard has been pretty consistent with utilizing operator bools on optional.  Not sure why specifically here you are drawing the line with readability.

EmosewaMC (approved):

PR #1556, opened 2024-05-01
chore: shallow checkout on fetcing git repos

aronwk-aaron:
for reducing ~~size on disk~~ cmake configuration time

### DIFF
--- a/cmake/FindGoogleTest.cmake
+++ b/cmake/FindGoogleTest.cmake
@@ -6,6 +6,7 @@ FetchContent_Declare(
 	googletest
 	GIT_REPOSITORY https://github.com/google/googletest.git
 	GIT_TAG release-1.12.1
+	GIT_SHALLOW 1
 )
 
 # For Windows: Prevent overriding the parent project's compiler/linker settings
--- a/cmake/FindJSON.cmake
+++ b/cmake/FindJSON.cmake
@@ -6,6 +6,7 @@ FetchContent_Declare(
 	json
 	GIT_REPOSITORY https://github.com/nlohmann/json
 	GIT_TAG v3.11.3
+	GIT_SHALLOW 1
 )
 
 FetchContent_GetProperties(json)


EmosewaMC:
How much is the reduction in size?

aronwk-aaron:
none.... good catch, but it does make the cmake configuration significantly faster
Looking into what the space is being used by, it's the git object pack for the json-src.
and that's not something we can fix. [They would need to repack the repo as suggested by our lord and savior Linus Torvalds.](https://gcc.gnu.org/legacy-ml/gcc/2007-12/msg00165.html)

This does take `cmake ..` from about 16s to 2s (I just switched branches and deleted the `build/_deps` folder a few times to get these numbers)

EmosewaMC:
take a look at these stats i tracked and logs from cmake
with shallow
real    1m33.516s
user    0m3.766s
sys     0m5.188s

real    1m14.400s
user    0m3.297s
sys     0m4.250s

real    3m0.019s
user    0m3.516s
sys     0m5.047s

without shallow
real    0m21.816s
user    0m8.438s
sys     0m5.234s

[ 11%] Creating directories for 'json-populate'
[ 22%] Performing download step (git clone) for 'json-populate'
Cloning into 'json-src'...
error: RPC failed; curl 92 HTTP/2 stream 0 was not closed cleanly: CANCEL (err 8)
error: 1888 bytes of body are still expected
fetch-pack: unexpected disconnect while reading sideband packet
fatal: early EOF
fatal: fetch-pack: invalid index-pack output
Cloning into 'json-src'...
error: 5053 bytes of body are still expected
fetch-pack: unexpected disconnect while reading sideband packet
fatal: early EOF
fatal: fetch-pack: invalid index-pack output
Cloning into 'json-src'...
error: RPC failed; curl 92 HTTP/2 stream 0 was not closed cleanly: CANCEL (err 8)
error: 4798 bytes of body are still expected
fetch-pack: unexpected disconnect while reading sideband packet
fatal: early EOF
fatal: fetch-pack: invalid index-pack output
-- Had to git clone more than once: 3 times.
CMake Error at json-subbuild/json-populate-prefix/tmp/json-populate-gitclone.cmake:39 (message):
  Failed to clone repository: 'https://github.com/nlohmann/json'


gmake[2]: *** [CMakeFiles/json-populate.dir/build.make:102: json-populate-prefix/src/json-populate-stamp/json-populate-download] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/json-populate.dir/all] Error 2
gmake: *** [Makefile:91: all] Error 2

CMake Error at /usr/share/cmake-3.29/Modules/FetchContent.cmake:1714 (message):
  Build step for json failed: 2
Call Stack (most recent call first):
  /usr/share/cmake-3.29/Modules/FetchContent.cmake:1854:EVAL:2 (__FetchContent_directPopulate)
  /usr/share/cmake-3.29/Modules/FetchContent.cmake:1854 (cmake_language)
  cmake/FindJSON.cmake:13 (FetchContent_Populate)
  CMakeLists.txt:99 (find_package)


-- Configuring incomplete, errors occurred!

real    3m51.237s
user    0m0.141s
sys     0m0.656s

aronwk-aaron:
given the need for json isn't viable due to the httplib not being viable for an API, just going to opt to remove json and maybe httplib too in a different PR

PR #1558, opened 2024-05-01
fix: out of order physics updates

EmosewaMC:
fixes an issue where physics entities were not given a chance to be marked as sleeping, causing a initial sleeping calls to be missed and causing objects that collided with one another to not register new collisions since they were sleeping at the time the new collision fired off.

Tested that Brick Fury now corectly aggros the _first_ spawn of enemies near by to him. Tested that the turrets in crux prime now correctly shoot the _first_ wave of enemies that spawn.


aronwk-aaron (approved):

EmosewaMC:
just for a reference, the client runs through these two in this order too, gameobj and then physics.

DarwinAnim8or (approved):

PR #1571, opened 2024-05-13
CppScripts Cleanup

### LINKED ISSUE #1511: ENH: Cleanup script loading

Script loading is quite clunky as it stands right now, given its form of a giant if else block, we would like to see it be cleaned up a bit.  

Instead of the if else chain [in the script loader](https://github.com/DarkflameUniverse/DarkflameServer/blob/main/dScripts/CppScripts.cpp#L339), something like the following  is wanted:
```cpp
map<string, std::function<CppScripts::Script*()>> ScriptLoader = {
        {"scripts/test.lua", [](){ return new Test(); } },
        {"scripts/foo.lua", [](){ return new Foo(); } },
        // and so it goes
 };
```
When creating the initial copy of a script, the following would be done:
```cpp
map<string, std::function<CppScripts::Script*()>> ScriptLoader = {
        {"scripts/test.lua", [](){ return new Test(); } },
        {"scripts/foo.lua", [](){ return new Foo(); } },
        // and so it goes
 };
std::map<string, CppScripts::Script*> cache;

CppScripts::Script* GetScript(string script_name) {
auto itr = cache.find(script_name);
if (itr != cache.end()) return itr->second;
auto itrLoad = ScriptLoader[script_name];
if (itrLoad == ScriptLoader.end()) {
    cache[script_name] = InvalidScript;
    return InvalidScript;    
}
auto toReturn = itrLoad->second();
cache[script_name] = toReturn;
return toReturn;
}
```

During the migration, its possible an error could cause a script to be missing accidentally due to a typo or another issue

modifying windows build to support if else chains longer than 128 blocks

### Additional context

Shoutouts windows for enforcing a limit that doesnt need to exist

TAHuntling:
Fixes #1511 
Desc: Changed the CppScripts file to use a lambda map to load scripts rather than the if else chain. This should make it easier to make future changes as well as reduce code size (especially with an issue with number of if-else statements for some operating machines).

Motivation: Received notice of the issue from a head developer

Type of Changes: Made changes to the CppScripts main file, which now use a lambda. Made a few comment changes but tried to keep all original comments. Overall file lines reduced drastically.

How Has This Been Tested?:
Emosewamc helped write a test file which we used to verify the scripts between the old scripts and new script changes. It returned successful tests when running `ctest` in the `/build` directory


EmosewaMC:
> Issue #1511
Change this to Fixes #1511 so the issue is automatically marked as closed

dScripts/CppScripts.cpp
```diff
@@ -327,668 +327,382 @@
...
+#include <functional>
+
 namespace {
-	// This is in the translation unit instead of the header to prevent wierd linker errors
-	InvalidScript* const InvalidToReturn = new InvalidScript();
-	std::map<std::string, CppScripts::Script*> m_Scripts;
+    // This is in the translation unit instead of the header to prevent weird linker errors
+    InvalidScript InvalidToReturn;
+    std::map<std::string, CppScripts::Script*> m_Scripts;
+    std::map<std::string, std::function<CppScripts::Script* ()>> test = {
```
EmosewaMC, line 338: g_Scripts would be a better name for this map

dScripts/CppScripts.cpp
```diff
@@ -327,668 +327,382 @@
...
-		script = new WildPants();
-	else if (scriptName == "scripts\\ai\\WILD\\L_WILD_NINJA_BRICKS.lua")
-		script = new WildNinjaBricks();
-	else if (scriptName == "scripts\\ai\\WILD\\L_WILD_NINJA_STUDENT.lua")
-		script = new WildNinjaStudent();
-	else if (scriptName == "scripts\\ai\\WILD\\L_WILD_NINJA_SENSEI.lua")
-		script = new WildNinjaSensei();
-	else if (scriptName == "scripts\\ai\\WILD\\L_LUP_generic_interact.lua")
-		script = new LupGenericInteract();
-	else if (scriptName.rfind("scripts\\zone\\LUPs\\RobotCity Intro\\WBL_RCIntro_RobotCitizen", 0) == 0)
```
EmosewaMC, line 974: Where did these go?

EmosewaMC (changes requested):

dScripts/CppScripts.cpp
EmosewaMC: there should be 4-5 with this style name we need to load, its the only missing script from the transfer

dScripts/CppScripts.cpp
```diff
@@ -327,668 +327,382 @@
...
+		{"scripts\\ai\\WILD\\L_WILD_GF_SNAIL.lua", [](){return new WildAndScared();}},
+		{"scripts\\ai\\WILD\\L_WILD_GF_GLOWBUG.lua", [](){return new WildGfGlowbug();}},
+		{"scripts\\ai\\WILD\\L_WILD_AMBIENT_CRAB.lua", [](){return new WildAmbientCrab();}},
+		{"scripts\\ai\\WILD\\L_WILD_PANTS.lua", [](){return new WildPants();}},
+		{"scripts\\ai\\WILD\\L_WILD_NINJA_BRICKS.lua", [](){return new WildNinjaBricks();}},
+		{"scripts\\ai\\WILD\\L_WILD_NINJA_STUDENT.lua", [](){return new WildNinjaStudent();}},
+		{"scripts\\ai\\WILD\\L_WILD_NINJA_SENSEI.lua", [](){return new WildNinjaSensei();}},
+		{"scripts\\ai\\WILD\\L_LUP_generic_interact.lua", [](){return new LupGenericInteract();}},
+
+		//And on, and on, and on, and on
```
EmosewaMC, line 680:
```suggestion
```

EmosewaMC:
Please also format your files before commiting with alt+shift+f in VSCode or your preferred editor

dScripts/CppScripts.cpp
TAHuntling: Made change

dScripts/CppScripts.cpp
TAHuntling: There are 5, one for each color. Currently at lines 680-684 in the CPPScripts.cpp file

If this is still a question let me know

aronwk-aaron:
You don't need to close this PR and open a new one, pushing to your branch will add changes to it

PR #1572, opened 2024-05-13
feat: help command sorting and pagination

TAHuntling:
feat: command organization and pagination

Desc: Modified SlashCommandHandler HelpCommand to use CommandInfos as a map, rather than a vector. This helps to alphabetize the commands, for easier location of certain commands. Additionally, pagination was added to display 10 total commands per page when running `/help`. You can run `/help 2` to access page 2, `/help 3` to access page 3 etc. I also made it so that it wont display pages that are not available, either due to GMlevel or number of total commands.

Motivation and Context: I was working on testing some script changes and used the `/help` command, only to be unable to find the commands i was looking for due to them being out of order and in a massive list.

Type of Changes: Changes mad to CommandInfos to be a map, as well as relocating the original code for the help command display into HelpCommand, which helped with some scope issues I was having.
   
How Has This Been Tested?: Building the server, and trying out the `/help` command as different GMLevels, as well as testing to ensure it returns `Invalid page number.` If the page number is too high or the user doesnt have access to said commands.

Help Improvements1
Help Improvements2
Help Improvements3
Help Improvements4

### DIFF
--- a/dGame/dUtilities/SlashCommandHandler.cpp
+++ b/dGame/dUtilities/SlashCommandHandler.cpp
@@ -19,10 +19,34 @@
 #include "dServer.h"
 
 namespace {
-	std::vector<Command> CommandInfos;
+	std::map<std::string, Command> CommandInfos;
+	//std::vector<Command> CommandInfos;
 	std::map<std::string, Command> RegisteredCommands;
 }
 
+void SlashCommandHandler::RegisterCommand(Command command) {
+    if (command.aliases.empty()) {
+        LOG("Command %s has no aliases! Skipping!", command.help.c_str());
+        return;
+    }
+
+    for (const auto& alias : command.aliases) {
+        LOG_DEBUG("Registering command %s", alias.c_str());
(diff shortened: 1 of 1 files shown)


dGame/dUtilities/SlashCommandHandler.cpp
```diff
@@ -19,10 +19,34 @@
...
+        LOG_DEBUG("Registering command %s", alias.c_str());
+        auto [_, success] = RegisteredCommands.try_emplace(alias, command);
+        // Don't allow duplicate commands
+        if (!success) {
+            LOG_DEBUG("Command alias %s is already registered! Skipping!", alias.c_str());
+            continue;
+        }
+    }
+
+    // Inserting into CommandInfos using the first alias as the key
```
EmosewaMC, line 43: Comment and if block can be removed.  Also remove all the commented out code

EmosewaMC (changes requested):
Please split this up into two PRs.  One for pages and one for organization

PR #1576, opened 2024-05-13
feat: Help command pagination

TAHuntling:
feat: Help Command Pagination

feat: Command Organization

Description: Changed HelpCommand to now be sorted into different pages for commands, which should help to further help users to find the correct commands they are looking for without having to scroll through a list of 92 commands. 

Motivation and Context: Recommendation from Aron, and my own experience reading through the /help command list.

How Has This Been Tested?: Running my server and using the `/help` command will return page 1 of the command list. Running `/help 1` will as well, but `/help 2` and so forth will return that respective page number of commands (currently 10 commands per page). If a user tries to enter a command higher than the number of pages, it will say it is an invalid page number. Additionally, commands will only show up given the user has the proper GMLevel to use that command.

### DIFF
--- a/dGame/dUtilities/SlashCommandHandler.cpp
+++ b/dGame/dUtilities/SlashCommandHandler.cpp
@@ -19,10 +19,34 @@
 #include "dServer.h"
 
 namespace {
-	std::vector<Command> CommandInfos;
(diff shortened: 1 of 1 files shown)


EmosewaMC (commented):
Will review when the other pr is merged

PR #1577, opened 2024-05-14
fix: undefined behavior in scripts and sirens issue

EmosewaMC:
fixes an issue where the sirens would not be destroyed correctly fixes undefined behavior in buff station

tested that the sirens now correctly stop stunning enemies when they die


aronwk-aaron (dismissed):

EmosewaMC:
check is smashable from db alongside death_behavior0

EmosewaMC:
Retested that pets, avant gardens survival, siren and property guard all now fully die with their correct death animations (no animation except the property guard)

PR #1579, opened 2024-05-14
chore: CppScripts refactor 

TAHuntling:
Fixes https://github.com/DarkflameUniverse/DarkflameServer/issues/1511
Desc: Changed the CppScripts file to use a lambda map to load scripts rather than the if else chain. This should make it easier to make future changes as well as reduce code size (especially with an issue with number of if-else statements for some operating machines).

Motivation: Received notice of the issue from a head developer

Type of Changes: Made changes to the CppScripts main file, which now use a lambda. Made a few comment changes but tried to keep all original comments. Overall file lines reduced drastically.

How Has This Been Tested?:
Emosewamc helped write a test file which we used to verify the scripts between the old scripts and new script changes. It returned successful tests when running ctest in the /build directory


EmosewaMC (approved):
Pending CI

aronwk-aaron (approved):

PR #1592, opened 2024-05-22
Client commands implementation

TAHuntling:
FEAT: Implementing client commands

Description: Implemented a list of commands which are used by the client rather than the server. Essentially made an empty command for the server which parses the command through to the client, which then does its thing (this mostly includes things like emotes and a few config options)

Motivation and Context: Adding the ability to use these commands to continue fleshing out our current command palette and help command.

Type of Changes: Changes made to SlashCommandHandler.cpp to include additional command attributes for client commands, added a declaration in GMZeroCommands.h and definition in GMZeroCommands.cpp

How Has This Been Tested: Running commands in game, which triggers their respective emotes and or other functionality. Additionally, these client commands are listed in the `/help` menu for users to have access to them and get information as necessary.

### DIFF
--- a/dGame/dUtilities/SlashCommandHandler.cpp
+++ b/dGame/dUtilities/SlashCommandHandler.cpp
@@ -1116,7 +1116,7 @@ void SlashCommandHandler::Startup() {
 	Command victoryCommand{
 		.help = "Victory!",
 		.info = "Victory!",
-		.aliases = {"Victory!"},
+		.aliases = {"victory!"},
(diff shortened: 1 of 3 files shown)


EmosewaMC (approved):

EmosewaMC:
as a note for the testing, these commands have always worked and were never broken.  This just prevents a "command not found" from being sent to the clients when one of these is used, and adds a help message for them.

aronwk-aaron (approved):

Issue #462, opened 2022-02-22
BUG: Attempting to put items in a full vault results in them being mailed to you


EmosewaMC:
### Make sure you've done the following:

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### Issue Description

When a vault is full, attempting to put an item into it will result in the item being mailed to you.  I have not tested this with items that can be stacked yet but items that don't stack will be mailed upon trying to place in the full vault.
The GameMessage that handles this is https://github.com/DarkflameUniverse/DarkflameServer/blob/ad52facef71e1349057512237ccbd0b157af19f9/dGame/dGameMessages/GameMessages.cpp#L5460-L5502

### Reproduction steps

1) Fill a vault
2) Place item in said full vault
3) Check player mail

### Expected Behavior

Item stays in previous inventory or something happens so that the item isnt mailed to you.  

### Environment

WSL

PR #1595, opened 2024-05-25
fix: prevent moving items between inventories under cetain circumsances

aronwk-aaron:
fixes: #462 
Tested moving items to a full inventory and moving the thinking hat, as well as added general error and success in the function where applicable

Tested that the item does not move or mailed and the error is displayed


dGame/dGameMessages/GameMessages.cpp
```diff
@@ -4564,16 +4565,31 @@ void GameMessages::HandleRequestMoveItemBetweenInventoryTypes(RakNet::BitStream&
...
+			if (item->GetLot() == 6086) { // Thinking hat
+				SendResponseMoveItemBetweenInventoryTypes(entity->GetObjectID(), sysAddr, invTypeDst, invTypeSrc, eReponseMoveItemBetweenInventoryTypeCode::FAIL_CANT_MOVE_THINKING_HAT);
+				return;
+			}
+			
+			auto* destInv = inventoryComponent->GetInventory(invTypeDst);
+			if (destInv && destInv->GetEmptySlots() == 0) {
+				SendResponseMoveItemBetweenInventoryTypes(entity->GetObjectID(), sysAddr, invTypeDst, invTypeSrc, eReponseMoveItemBetweenInventoryTypeCode::FAIL_INV_FULL);
+				return;
+			}
```
EmosewaMC, line 4592: Is this bothering to allow people to move a partial amount of objects or to fill in remaining not full stacks in the destination inventory?

EmosewaMC: What about if someone moves enough items such that the transfer is split up into 2 stacks?  what happens in this case?

aronwk-aaron: It does not combine stacks, and so it wont have an issue with having to split stacks.
I think it would be more complex to do those checks, And splitting stacks where some move and some don't isn't handled in the fail messages either... so I doubt the did all of that work. and you just had to have a slot empty to move stuff like that

dCommon/dEnums/eReponseMoveItemBetweenInventoryTypeCode.h
```diff
@@ -0,0 +1,21 @@
+#ifndef __EREPONSEMOVEITEMBETWEENINVENTORYTYPECODE__H__
+#define __EREPONSEMOVEITEMBETWEENINVENTORYTYPECODE__H__
+
+#include <cstdint>
+
+enum class eReponseMoveItemBetweenInventoryTypeCode : int32_t {
```
EmosewaMC, line 6: any particular reason this is signed and not unsigned?

aronwk-aaron: msvc defaults to int

Jettford (approved):

DarwinAnim8or (approved):

PR #1599, opened 2024-05-25
feat: Observers and deferred implementations

Wincent01:
This pull request seeks to add helper classes for observers as well as a way to optionally defer/override an implementation of a method or action.

These classes would in the future help with server-side modding. The observer class is very standard as observers go. Usually you would relay on something like a strategy pattern to switch between implementations — do not think that is quite the right solution here.

In the event of modding through library loading at runtime, they could set their observers and overrides through these classes.

In this pull request are examples of how these classes could be implemented into existing classes. These could be excluded if that would be preferred.

### DIFF
--- a/dCommon/Implementation.h
+++ b/dCommon/Implementation.h
@@ -1,0 +1,92 @@
+#ifndef __IMPLEMENTATION_H__
+#define __IMPLEMENTATION_H__
+
+#include <functional>
+#include <optional>
+
+/**
+ * @brief A way to defer the implementation of an action.
+ * 
+ * @tparam R The result of the action.
+ * @tparam T The types of the arguments that the implementation requires.
+ */
+template <typename R, typename... T>
+class Implementation {
+public:
+	typedef std::function<std::optional<R>(T...)> ImplementationFunction;
+
+	/**
+	 * @brief Sets the implementation of the action.
+	 * 
+	 * @param implementation The implementation of the action.
+	 */
+	void SetImplementation(const ImplementationFunction& implementation) {
+		this->implementation = implementation;
+	}
+
+	/**
+	 * @brief Clears the implementation of the action.
+	 */
+	void ClearImplementation() {
+		implementation.reset();
+	}
+
+	/**
+	 * @brief Checks if the implementation is set.
+	 * 
+	 * @return true If the implementation is set.
+	 * @return false If the implementation is not set.
+	 */
+	bool IsSet() const {
+		return implementation.has_value();
+	}
+
+	/**
+	 * @brief Executes the implementation if it is set.
+	 * 
+	 * @param args The arguments to pass to the implementation.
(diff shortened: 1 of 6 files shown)


dCommon/Observable.h
```diff
@@ -0,0 +1,73 @@
...
+		observers.push_back(observer);
+	}
+
+	/**
+	 * @brief Removes an observer from the event.
+	 * 
+	 * @param observer The observer to remove.
+	 */
+	void RemoveObserver(const Observer& observer) {
+		observers.erase(std::remove(observers.begin(), observers.end(), observer), observers.end());
```
EmosewaMC, line 32: Consider using ranges here for algorithm functions when you are going from start to end

dGame/Entity.h
```diff
@@ -299,6 +300,11 @@ class Entity {
 	// Scale will only be communicated to the client when the construction packet is sent
 	void SetScale(const float scale) { m_Scale = scale; };
 
+	/**
+	 * @brief The observable for player entity position updates.
+	 */
+	static Observable<Entity*, const PositionUpdate&> OnPlayerPositionUpdate;
```
EmosewaMC, line 306: Consider moving static variables into an unnamed namespace in the cpp file (prevents lazy initialization)

EmosewaMC (approved):
Two nits

aronwk-aaron (approved):

dCommon/Implementation.h
```diff
@@ -0,0 +1,92 @@
...
+	std::optional<R> operator()(T... args) {
+		return !IsSet() ? std::nullopt : implementation(args...);
+	}
+
+private:
+	std::optional<ImplementationFunction> implementation;
+
+};
+
+#endif  //!__IMPLEMENTATION_H__
```
aronwk-aaron, line 92: no newline at eof

aronwk-aaron (changes requested):

dGame/Entity.h
Wincent01: Then it wouldn't be accessible from the outside of that cpp file, which is something desired here. Alternative is having a getter for these, would that be preferable?

EmosewaMC: do you have examples of how this would be accessed?  hard to see the full picture here without how you would use it

EmosewaMC: the reason for asking about this is to prevent any future issue where you push stuff into this container, and then if the server crashes, all those observers are removed before the server does its processing of data during the crash.  if that is a non-concern then feel free to dismiss, though i try to verify that if a variable is static, it truly is 100% OK to be fully destructed before the crash handler does its work.

DarwinAnim8or (approved):

DarwinAnim8or:
Nice to see the gameprogrammingpatterns.com influence ;) Big fan of observers so this looks great

dCommon/Implementation.h
jadebenn: gonna toss this in another PR as per discord chat

PR #1602, opened 2024-05-26
fix: skill desync when picking up powerups or casting server side calculations

EmosewaMC:
From static analysis it appears that the client and server need to be keeping track of their own skillUIds as inside the client, there are GameMessages for CastLocalSkill which increment the client side skillUId counter `0x00d45300` alongside the same tracker being used for tracking the ones the client sends to the server `0x00d44980`.  With no way for the server to actually know the clients actual skill tracker value, I believe it is safe to assume that the two are ok to overlap and you just have to keep track of both.  This PR utilizes a multimap to accomplish this task.  

Tested that battling on Crux Prime using a wormholer, bat lord helm and picking up powerups constantly from boxes does not cause my character to become permanently slowed, nor do I permanently gain as money magnet.
Tested that loading into Avant Gardens and destroying the piles of debris at this location with the wormholer, a speedboost and a money magnet active results in
- wormholer doing damage
- speedboosts actually going away
- money magnet actually going away
instead of
- wormholer sometimes not doing damage, causing invisible objects
- speedboosts, both positive and negative going away
- money magnet staying permanently
image

### DIFF
--- a/dGame/dBehaviors/BehaviorContext.cpp
+++ b/dGame/dBehaviors/BehaviorContext.cpp
@@ -105,7 +105,7 @@ void BehaviorContext::ExecuteUpdates() {
 	this->scheduledUpdates.clear();
 }
 
-void BehaviorContext::SyncBehavior(const uint32_t syncId, RakNet::BitStream& bitStream) {
+bool BehaviorContext::SyncBehavior(const uint32_t syncId, RakNet::BitStream& bitStream) {
 	BehaviorSyncEntry entry;
 	auto found = false;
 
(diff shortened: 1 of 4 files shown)


aronwk-aaron (approved):

Jettford (approved):

Xiphoseer:
> there are GameMessages for CastLocalSkill which increment the client side skillUId counter 0x00d45300 alongside the same tracker being used for tracking the ones the client sends to the server 0x00d44980

Are those base offsets of some sort of global array? Are they hardcoded or could they change on reallocation?

EmosewaMC:
> > there are GameMessages for CastLocalSkill which increment the client side skillUId counter 0x00d45300 alongside the same tracker being used for tracking the ones the client sends to the server 0x00d44980
> 
> Are those base offsets of some sort of global array? Are they hardcoded or could they change on reallocation?

the addresses above are subroutines where the variable (which is located on the players' LWOSkillComponent) are incremented the same way and show that the variable on the client does not interact with one on the server.  these values reset upon a world change since the players' LWOSkillComponent is re-created and as such reset to 0.  Can share more over discord if you'd like :)

PR #1603, opened 2024-05-27
fix: use after free and uninitialized memory

EmosewaMC:
fixes errors spotted in valgrind when logging into the server

UnEquipItem works fine for sending a proxy, however PurgeProxies unfortunately deletes the item you are unequipping silently, causing a use after free when calling UnEquipScripts.  For now i have moved UnEquipScripts call up so there is no use after free.

Initialized some structures and variables that were used uninitialized.
Checked packet length for incoming packets to ensure we're not reading garbage memory.

Tested that
- I can login and load into Avant Gardens
- Valgrind no longer reports use after free for the purged items
- Valgrind no longer reports reading undefined memory when handling packets of too small of a size or when forwarding a packet to chat server
- Valgrind no longer reports using of uninitialized memory when accessing Character fields
- I can send receive (friend requests) and load my friends list (same for ignore lost)
- I can login and leave and chat server is still notified of my absense

### DIFF
--- a/dChatServer/ChatServer.cpp
+++ b/dChatServer/ChatServer.cpp
@@ -179,6 +179,7 @@ int main(int argc, char** argv) {
 }
 
 void HandlePacket(Packet* packet) {
+	if (packet->length < 1) return;
 	if (packet->data[0] == ID_DISCONNECTION_NOTIFICATION || packet->data[0] == ID_CONNECTION_LOST) {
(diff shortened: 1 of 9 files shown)


aronwk-aaron (dismissed):

EmosewaMC:
Tested that flying to a property no longer reports warnings

aronwk-aaron (approved):

Xiphoseer (approved):

PR #1604, opened 2024-05-27
fix: Dockerfile vanity COPY

SpeedyFireCyclone:
The current COPY directives in the Dockerfile that add the vanity files have an unintended trailing asterisk, causing only root.xml's contents to be in there under the filename '*'.
This is obviously unintendend as this causes the vanity to be non-functional in the default docker setup.


aronwk-aaron (approved):

EmosewaMC (approved):

EmosewaMC:
Thanks for spotting this and submitting a fix!

PR #1606, opened 2024-05-31
chore: optimize chat server reads and fix crash

EmosewaMC:
Tested that players with valid names up to the usual 33 character max are still added to the player container
Tested that you can still team with <= 4 players on a team
Tested that chat server no longer crashes with a bad memberSize variable
Verified that InsertPlayer is indeed much faster now and is no longer a slow point of ChatServer
image


aronwk-aaron (approved):

EmosewaMC:
Also fixes an issue where, due to race conditions, a players name would be twice as long when logging into a world.

DarwinAnim8or (approved):
What was making InsertPlayer so slow? I don't see it immediately; just the name reading?

PR #1607, opened 2024-05-31
Buyback inventory overflow glitch fix

TAHuntling:
Currently still a work in progress. I have something that may be very similar to what we want, and need to go through it and likely simplify things, remove logs, etc, but for the time being ive kept all those for debugging. 

There is an issue with leaving a single inventory slot open at all times, which may be acceptable for the fix or not, as either way i could see how it could behave that way. That being said, it shouldnt be too hard of a change to get it to not include that empty space, i just need to go through and figure out what causes that specifically. 

Tested using a variety of random items and stacks, and i know some modifications will definitely be requested and made, again this is just very rough unfinished code.

Thanks for the feedback

### DIFF
--- a/dGame/dComponents/InventoryComponent.cpp
+++ b/dGame/dComponents/InventoryComponent.cpp
@@ -161,7 +161,7 @@ const EquipmentMap& InventoryComponent::GetEquippedItems() const {
 	return m_Equipped;
 }
 
-void InventoryComponent::AddItem(
+std::vector<LWOOBJID> InventoryComponent::AddItem(
 	const LOT lot,
 	const uint32_t count,
 	eLootSourceType lootSourceType,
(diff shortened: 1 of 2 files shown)


dGame/dComponents/InventoryComponent.cpp
```diff
@@ -175,124 +175,153 @@ void InventoryComponent::AddItem(
...
-
-			return;
+			return toReturn;
 		}
 
 		auto* item = new Item(lot, inventory, slot, count, config, parent, showFlyingLoot, isModMoveAndEquip, subKey, bound, lootSourceType);
+		LOG("Created new item with ID %llu, LOT %i in slot %i", item->GetId(), lot, slot);
+
+		// Check if inventory type is VENDOR_BUYBACK and manage the inventory size
+		if (inventoryType == VENDOR_BUYBACK) {
```
EmosewaMC, line 217: please use full scoping on enums i.e. eInventoryType::VENDOR_BUYBACK

dGame/dComponents/InventoryComponent.h
```diff
@@ -379,7 +379,7 @@ class InventoryComponent final : public Component {
 	 * All the inventory this entity possesses
 	 */
 	std::map<eInventoryType, Inventory*> m_Inventories;
-
+	std::vector<LWOOBJID> invTransferred;
```
EmosewaMC, line 382: The way you designed this logic I would recommend just using a queue instead of a vector.  Since you are choosing to let stacking items fill in the previous slot without overriding its age, a vector no longer makes sense as a container to store buyback order in.  Also a better name here would be buybackItems.  There are more than 1 transfer inventory and this name could be confusing.

dGame/dComponents/InventoryComponent.h
```diff
@@ -477,6 +477,8 @@ class InventoryComponent final : public Component {
 	 * @param document the xml doc to load from
 	 */
 	void UpdatePetXml(tinyxml2::XMLDocument& document);
+	void ManageVendorBuybackInventory(std::vector<LWOOBJID>& itemVector, Item* newItem, Inventory* inventory, bool removeItem);
```
EmosewaMC, line 480: The first argument to this function can be removed.  the argument you set it as in every call is already a member of this.  You can already just access the container without passing it as a parameter.  newItem is never used as a pointer, only as an objectID.  Please convert this to an LWOOBJID instead.  removeitem as an argument is unused and should be removed.  inventory as a pointer is never used outside of the type.  Please convert this to just an eInventoryType instead.

EmosewaMC (commented):
Logic is fine overall, but where stuff is living has some interesting decisions based on how you decided to implement the logic for the buyback inventory.  We will convene over discord at some point and discuss how to design the API in a more robust way.

All logs in this PR should be removed when finished unless they are reporting an error,  alongside the comment about if there is a bug if thats ok, in this case no it is not because it is an easily fixed bug and this feature is not critical enough to replace a bug with another one

DarwinAnim8or:
Any updates on this? Is it ready for review?

jadebenn:
Seconding the question about merge readiness. I'm coming back from a period of long inactivity, and I wouldn't mind helping bringing this back to a ready-to-merge state.

EmosewaMC:
there was some odd behavior with how items were popped from the inventory that I was working with them to resolve (would result in a funky deletion order)

PR #1611, opened 2024-06-07
feat: deletion restrictions

aronwk-aaron:
WIP

### DIFF
--- a/dDatabase/CDClientDatabase/CDClientTables/CDDeletionRestrictionsTable.cpp
+++ b/dDatabase/CDClientDatabase/CDClientTables/CDDeletionRestrictionsTable.cpp
@@ -1,10 +1,16 @@
 #include "CDDeletionRestrictionsTable.h"
 #include "GeneralUtils.h"
+#include "eDeletionRestrictionsCheckType.h"
 
-//! Constructor
+CDDeletionRestriction CDDeletionRestrictionsTable::Default = {
+	.id = 0,
+	.restricted = false,
+	.ids = {},
+	.checkType = eDeletionRestrictionsCheckType::MAX
+};
+
 void CDDeletionRestrictionsTable::LoadValuesFromDatabase() {
 
-	// First, get the size of the table
 	uint32_t size = 0;
 	auto tableSize = CDClientDatabase::ExecuteQuery("SELECT COUNT(*) FROM CurrencyTable");
 	while (!tableSize.eof()) {
@@ -15,14 +21,11 @@ void CDDeletionRestrictionsTable::LoadValuesFromDatabase() {
 
 	tableSize.finalize();
 
-	// Reserve the size
 	auto& entries = GetEntriesMutable();
-	entries.reserve(size);
 
-	// Now get the data
 	auto tableData = CDClientDatabase::ExecuteQuery("SELECT * FROM DeletionRestrictions");
 	while (!tableData.eof()) {
-		CDDeletionRestrictions entry;
+		CDDeletionRestriction entry;
 		entry.id = tableData.getIntField("id", -1);
 		if (entry.id == -1) continue;
 		entry.restricted = tableData.getIntField("restricted", -1);
@@ -37,17 +40,18 @@ void CDDeletionRestrictionsTable::LoadValuesFromDatabase() {
 		}
 		entry.checkType = static_cast<eDeletionRestrictionsCheckType>(tableData.getIntField("checkType", 6)); // MAX
 
-		entries.push_back(entry);
+		entries.insert(std::make_pair(entry.id, entry));
 		tableData.nextRow();
 	}
 
 	tableData.finalize();
 }
 
-std::vector<CDDeletionRestrictions> CDDeletionRestrictionsTable::Query(std::function<bool(CDDeletionRestrictions)> predicate) {
-	std::vector<CDDeletionRestrictions> data = cpplinq::from(GetEntries())
-		>> cpplinq::where(predicate)
-		>> cpplinq::to_vector();
-
-	return data;
+const CDDeletionRestriction& CDDeletionRestrictionsTable::GetByID(uint32_t id) {
+	auto& entries = GetEntries();
+	const auto& it = entries.find(id);
+	if (it != entries.end()) {
+		return it->second;
+	}
+	return Default;
 }
--- a/dDatabase/CDClientDatabase/CDClientTables/CDDeletionRestrictionsTable.cpp
+++ b/dDatabase/CDClientDatabase/CDClientTables/CDDeletionRestrictionsTable.cpp
@@ -35,7 +35,7 @@ void CDDeletionRestrictionsTable::LoadValuesFromDatabase() {
 				} 
 			}
 		}
-		entry.checkType = static_cast<eDeletionRestrictionsCheckType>(tableData.getIntField("checkType", 6));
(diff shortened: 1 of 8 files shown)


dDatabase/CDClientDatabase/CDClientTables/CDDeletionRestrictionsTable.cpp
```diff
@@ -0,0 +1,57 @@
...
+				} 
+			}
+		}
+		entry.checkType = static_cast<eDeletionRestrictionsCheckType>(tableData.getIntField("checkType", 6)); // MAX
+
+		entries.insert(std::make_pair(entry.id, entry));
+		tableData.nextRow();
+	}
+
+	tableData.finalize();
```
EmosewaMC, line 47:
```suggestion
```
RAII takes care of this call :)

dDatabase/CDClientDatabase/CDClientTables/CDDeletionRestrictionsTable.cpp
```diff
@@ -0,0 +1,57 @@
...
+		const std::string raw_ids = tableData.getStringField("ids", "");
+		if (!raw_ids.empty()) {
+			for (const auto& idstr : GeneralUtils::SplitString(raw_ids, ',')) {
+				if (!idstr.empty()) {
+					const auto id = GeneralUtils::TryParse<int32_t>(idstr).value_or(-1);
+					if (id != -1) entry.ids.push_back(id);
+				} 
+			}
+		}
+		entry.checkType = static_cast<eDeletionRestrictionsCheckType>(tableData.getIntField("checkType", 6)); // MAX
```
EmosewaMC, line 41: I would like it if this was the enum value instead of just the 6 and a comment as if we added our own restriction, we would need to remember to update this instead of just having it be updated for us

dDatabase/CDClientDatabase/CDClientTables/CDDeletionRestrictionsTable.cpp
```diff
@@ -0,0 +1,57 @@
...
+		CDDeletionRestriction entry;
+		entry.id = tableData.getIntField("id", -1);
+		if (entry.id == -1) continue;
+		entry.restricted = tableData.getIntField("restricted", -1);
+		const std::string raw_ids = tableData.getStringField("ids", "");
+		if (!raw_ids.empty()) {
+			for (const auto& idstr : GeneralUtils::SplitString(raw_ids, ',')) {
+				if (!idstr.empty()) {
+					const auto id = GeneralUtils::TryParse<int32_t>(idstr).value_or(-1);
+					if (id != -1) entry.ids.push_back(id);
```
EmosewaMC, line 37:
```suggestion
					const auto id = GeneralUtils::TryParse<int32_t>(idstr);
					if (id) entry.ids.push_back(id);
```
implicitly covertible to bool.  also .has_value() works too for the check without the 2 comparisons :)

dGame/dInventory/Item.cpp
```diff
@@ -568,3 +570,43 @@ void Item::LoadConfigXml(const tinyxml2::XMLElement& i) {
...
+			return true;
+		case eDeletionRestrictionsCheckType::WHILE_IN_ZONE:
+			if (std::ranges::find(restriction.ids, Game::zoneManager->GetZoneID().GetMapID()) != restriction.ids.end()) return false;
+			else return true;
+		case eDeletionRestrictionsCheckType::ALWAYS_RESTRICTED:
+			return false;
+		case eDeletionRestrictionsCheckType::MAX:
+		default:
+			return true;
+	}
```
EmosewaMC, line 611: I would consider instead a bool toReturn before the switch and a single return at the bottom of the function instead.  This would also allow you to greatly simplify the logic in the loop

dDatabase/CDClientDatabase/CDClientTables/CDDeletionRestrictionsTable.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+	uint32_t size = 0;
+	auto tableSize = CDClientDatabase::ExecuteQuery("SELECT COUNT(*) FROM CurrencyTable");
+	while (!tableSize.eof()) {
+		size = tableSize.getIntField(0, 0);
+
+		tableSize.nextRow();
+	}
+
+	tableSize.finalize();
```
EmosewaMC, line 22:
```suggestion
```
None of this work is used.

dDatabase/CDClientDatabase/CDClientTables/CDDeletionRestrictionsTable.h
```diff
@@ -0,0 +1,20 @@
...
+	std::vector<uint32_t> ids;
+	eDeletionRestrictionsCheckType checkType;
+};
+
+class CDDeletionRestrictionsTable : public CDTable<CDDeletionRestrictionsTable, std::map<uint32_t, CDDeletionRestriction>> {
+public:
+	void LoadValuesFromDatabase();
+	const CDDeletionRestriction& GetByID(uint32_t id);
+
+	static CDDeletionRestriction Default;
```
EmosewaMC, line 19: It would be great if this could be not public, and instead use a getter which returns a const reference to a default struct so this cannot be modified accidentally, alongside removing a static variable in place for a static function and an unnamed namespace member.

EmosewaMC:
thank you for adding the new table!  Let's try and clean it up a bit since I know our table reading in other places is quite messy and tends to spider web to other tables.  Some extra feedback was left on the logic for the check as well :)

dDatabase/CDClientDatabase/CDClientTables/CDDeletionRestrictionsTable.cpp
aronwk-aaron: makes sense! I didn't have the enum in this file when I wrote this part

dGame/dInventory/Item.cpp
aronwk-aaron: no, that would make too much sense /s

dDatabase/CDClientDatabase/CDClientTables/CDDeletionRestrictionsTable.h
aronwk-aaron: Tell that to itemComponentTable

dDatabase/CDClientDatabase/CDClientTables/CDDeletionRestrictionsTable.cpp
aronwk-aaron: actually, can't use the enum since it's a class.
this is the behavior in the client, and I can't really see modding needing other restrictions than the ones listed, but maybe I'm not that imaginative lol

dGame/dInventory/Item.cpp
aronwk-aaron: One of reasons for the verbose-ness of this currently is that I wanted to make sure I got the logic correct. now that I have, I will simplify it

dDatabase/CDClientDatabase/CDClientTables/CDDeletionRestrictionsTable.cpp
EmosewaMC: Is a static cast not doable?

aronwk-aaron: ah true, just would look ugly :P

PR #1615, opened 2024-06-12
feat: add millisecond tracking for racing

EmosewaMC:
Tested that times are now saved, compared and shown on the client in millisecond format.  Tested that better lap times correctly update.


EmosewaMC:
unbelievable that you cant do copy assignment with two numbers.  thanks c++ committee.

DarwinAnim8or (approved):

aronwk-aaron (approved):

PR #1620, opened 2024-07-02
feat: add admin account creation options from cli

EmosewaMC:
Tested that creating accounts through CLI shows the option for setting admin access.
only `y` or `yes` allows for an account to be an admin, anything else results in a civilian account
updating an account now prompts for an updated gm level and updates it accordingly.

### DIFF
--- a/dMasterServer/MasterServer.cpp
+++ b/dMasterServer/MasterServer.cpp
@@ -41,6 +41,7 @@
 #include "Start.h"
 #include "Server.h"
 #include "CDZoneTableTable.h"
+#include "eGameMasterLevel.h"
 
 namespace Game {
 	Logger* logger = nullptr;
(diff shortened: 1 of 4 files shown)


dMasterServer/MasterServer.cpp
```diff
@@ -187,15 +188,20 @@ int main(int argc, char** argv) {
 		std::cout << "Enter a username: ";
 		std::cin >> username;
 
+		const auto checkIsAdmin = []() {
+			std::string admin;
+			std::cout << "What level of privilege should this account have? Please enter a number between 0 and 9 inclusive. No entry will default to 0." << std::endl;
```
Xiphoseer, line 193:
```suggestion
			std::cout << "What level of privilege should this account have? Please enter a number between 0 and 9 inclusive. No entry will default to 0." << std::endl;
```
```suggestion
			std::cout << "What level of privilege should this account have? Please enter a number between 0 (Player) and 9 (Admin) inclusive. No entry will default to 0." << std::endl;
```

Xiphoseer (approved):

dMasterServer/MasterServer.cpp
Xiphoseer: Fwiw, enum calls them `Civilian` / "Normal Player" and `Operator`  "can shut down server"

EmosewaMC: I had minecraft on the brain 🙈

Xiphoseer (approved):

aronwk-aaron (approved):

PR #1623, opened 2024-07-06
feat: Add some save data tests

EmosewaMC:
Added a test to ensure save data can be read from, written to, read from again and is in the expected updated state.  Will be used to allow for safe editing of character xml saving code.

Tests pass.


aronwk-aaron:
Error

EmosewaMC:
> Error

ran tests all fine locally, only when pushed they started failing.

aronwk-aaron (approved):

dDatabase/GameDatabase/TestSQL/TestSQLDatabase.cpp
DarwinAnim8or, line 1: Empty function?

DarwinAnim8or (approved):

dDatabase/GameDatabase/TestSQL/TestSQLDatabase.cpp
EmosewaMC: yeah, they are overridden abstract functions (to ensure that any alternate database has the required functions for the server to function) and since they dont return anything, they will just be an empty implementation.  I'll see if I can remove this file and define all those functions as ` = default;` later.

DarwinAnim8or: alr, ill mark as resolved so you can merge in the meantime

PR #1634, opened 2024-09-10
Fix mission related preconditions

Wincent01:
The current logic registered missions that had not been accepted as passing preconditions they should not. This PR seeks to use the GetMissionState function which takes care of this case.


DarwinAnim8or (approved):

aronwk-aaron (approved):

EmosewaMC:
The logic was incorrect before, however if there are any tests you can do to make sure this change has the expected results, those would be appreciated.  Noticed this 2 years ago and didnt fix because it wasnt broken lol.

EmosewaMC (approved):

PR #1636, opened 2024-10-13
feat: TCP Transport Layer

Wincent01:
# TCP Transport Layer

### Scope
This PR seeks to add an alternative TCP transport protocol. It works under the specifications of lcdr's tcpudp shim dll. The protocol initially specified the use of UDP for unreliable packets, but this is disabled in the dll due to port issues. This feature PR will assume this disabled state is permanent; UDP for unreliable packets will not be enabled.

### Current status
* Optionally compiled additional TCP transport layer.
* Config to enable it.
* Tested and functional with 2 clients.
* Removed unused RakNet replica manager and id manager. We've got our own replica manager since pre-open-source.
* Utilizes async boost behavior.

### Refactor structure
* dServer is not an adaptor for the transport layer.
* Multiple transport layers that implement a interface.
* No significant change to the dServer interface.

### Todo
* Figure out how to do ping calculations.
* Fix crashes on universe shutdown.
* Test TLS on a VPS.
* Remove unnecessary logging.
* Test with lots of clients.
* Maybe evaluate performance vs raknet.
* Finish "master" to "manager" naming refactor (I'm going to be that guy, this is as good of a time as any).

### DIFF
--- a/dNet/TcpSession.cpp
+++ b/dNet/TcpSession.cpp
@@ -1,0 +1,307 @@
+#include "TcpSession.h"
+
+#include "TcpTransportLayer.h"
+#include "Logger.h"
+#include "Game.h"
+#include "MessageIdentifiers.h"
+
+#include <boost/endian/conversion.hpp>
+
+TcpSession::TcpSession(boost::asio::io_service& io_service, boost::asio::ssl::context* sslContext, TcpTransportLayer* server):
+	m_PingTimerObject(io_service) {
+	m_Server = server;
+	
+	m_UseTls = server->GetIsEncrypted();
+	
+	if (m_UseTls) {
+		m_SSLStream = std::make_unique<boost::asio::ssl::stream<boost::asio::ip::tcp::socket>>(io_service, *sslContext);
+	}
+	else {
+		m_Socket = std::make_unique<boost::asio::ip::tcp::socket>(io_service);
+	}
+
+	m_IsConnected = false;
+	m_ReadyForUserPackets = false;
+	m_PingCount = 0;
+	m_CumulativePing = 0;
+	m_LastPing = 0;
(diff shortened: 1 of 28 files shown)


dNet/TcpSession.cpp
```diff
@@ -0,0 +1,307 @@
...
+	}
+}
+
+void TcpSession::HandlePing(Packet* packet) {
+	RakNet::BitStream bitStream;
+	bitStream.Write<uint8_t>(ID_CONNECTED_PONG);
+	bitStream.Write<uint8_t>(packet->data[1]);
+	bitStream.Write<uint8_t>(packet->data[2]);
+	bitStream.Write<uint8_t>(packet->data[3]);
+	bitStream.Write<uint8_t>(packet->data[4]);
```
EmosewaMC, line 205: This could access undefined heap memory if data is < 5 in size, i would consider bounds checks here on a packet

dNet/TcpTransportLayer.h
```diff
@@ -0,0 +1,158 @@
...
+
+namespace std {
+	// SystemAddress hash specialization
+	template <>
+	struct hash<SystemAddress> {
+		std::size_t operator()(const SystemAddress& sysAddr) const {
+			const std::size_t hash1 = sysAddr.binaryAddress;
+			const std::size_t hash2 = sysAddr.port;
+			return (hash1 | (hash2 << 32));
+		}
```
EmosewaMC, line 32: on platforms where size_t is 32 bits, this will result in just being hash1 always.  a 32 bit shift on a 4 byte value also is not a valid operation which is worth considering for a different hash value

EmosewaMC (commented):
cool stuff, some comments about bounds checking but overall looks cool

PR #1638, opened 2024-10-22
move command resetmission to gmlevel 0

bgstack15:
While it takes knowledge of opening up the client-side locale.xml, removing impossible missions from a civilian user's active mission list is incredibly useful.


EmosewaMC:
This command is not meant to be run on civilians and is only for debugging purposes.

EmosewaMC:
Can you please elaborate on which missions are impossible to complete?  all missions in LEGO Universe are currently able to be completed without issue on the master branch.

bgstack15:
On a small server I don't have enough users to ever populate 5 public worlds, so I cannot complete the "Visit 5 Block Yard or Avant Gardens properties."

PR #1641, opened 2024-10-27
feat: add debug builds for CI

EmosewaMC:
alongside that, fixes the issues caused by adding said pipelines
(copying over the shared libraries to directories where they are expected to exist on windows and mac)

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -5,124 +5,163 @@
       "minor": 14,
       "patch": 0
     },
-    "configurePresets": [
-      {
-        "name": "default",
-        "displayName": "Default configure step",
-        "description": "Use 'build' dir and Unix makefiles",
-        "binaryDir": "${sourceDir}/build",
-        "generator": "Unix Makefiles"
-      },
-      {
-        "name": "ci-ubuntu-22.04",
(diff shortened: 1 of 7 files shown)


CMakePresets.json
```diff
@@ -5,124 +5,163 @@
...
-			"jobs": 2
-		},
-        "output": {
-          "outputOnFailure": true
-        },
-		"filter": {
-			"exclude": {
-				"name": "((example)|(minigzip))+"
-			}
+  "configurePresets": [
```
EmosewaMC, line 8: We had to do the whitespace changes at somepoint sorry 🙈

aronwk-aaron (approved):

jadebenn:
This isn't quite configured correctly, I don't think. MSVC builds are handled a bit differently and (generally) shouldn't require separate configurations for debug and release because the MSVC build system is smart enough to be able to read the build preset instead.

jadebenn:
Screenshot 2024-11-16 18 44 46

jadebenn:
Since we merged #1439, one thing I think that might be worth further research on your end is whether we need to keep some of these MSVC presets. On my end, both 'Windows Only Configure Settings' and 'Default' don't work properly.
image

PR #1644, opened 2024-10-30
fix: chat server crash

EmosewaMC:
tested that sending an empty packet with this messageID no longer crashes the chat server


aronwk-aaron (approved):

jadebenn (approved):
lgtm

PR #1646, opened 2024-11-10
fix: laggy property models (and probably more)

EmosewaMC:
global fix correcting the initial physics motion state from 0 to 5 (confirmed in client).  packet captures from a few worlds (didnt scan more than 5 files) show that the value for simple physics was either 5 as the default, or 4 for property models, or 1 for property models with behaviors.

properties with pre-built models no longer lag and values of physics types should be correct across the board. Not sure what else this fixes but probably more stuff related to performance

### DIFF
--- a/dGame/dComponents/SimplePhysicsComponent.cpp
+++ b/dGame/dComponents/SimplePhysicsComponent.cpp
@@ -27,6 +27,7 @@ SimplePhysicsComponent::SimplePhysicsComponent(Entity* parent, uint32_t componen
 	} else {
 		SetClimbableType(eClimbableType::CLIMBABLE_TYPE_NOT);
 	}
+	m_PhysicsMotionState = m_Parent->GetVarAs<uint32_t>(u"motionType");
 }
 
 SimplePhysicsComponent::~SimplePhysicsComponent() {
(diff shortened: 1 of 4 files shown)


aronwk-aaron (dismissed):
Amazing find

dGame/dComponents/SimplePhysicsComponent.h
```diff
@@ -102,7 +102,9 @@ class SimplePhysicsComponent : public PhysicsComponent {
 	/**
 	 * The current physics motion state
 	 */
-	uint32_t m_PhysicsMotionState = 0;
+	uint32_t m_PhysicsMotionState = 5;
+
+	bool m_DirtyPhysicsMotionState = true;
```
aronwk-aaron, line 107: I don't see you set this to false after serialization

dGame/dComponents/SimplePhysicsComponent.cpp
```diff
@@ -51,6 +51,7 @@ void SimplePhysicsComponent::Serialize(RakNet::BitStream& outBitStream, bool bIs
 	outBitStream.Write(m_DirtyPhysicsMotionState || bIsInitialUpdate);
 	if (m_DirtyPhysicsMotionState || bIsInitialUpdate) {
 		outBitStream.Write<uint32_t>(m_PhysicsMotionState);
+		m_DirtyPhysicsMotionState = false;
```
aronwk-aaron, line 54: Did you test this? :P

EmosewaMC: nope!

jadebenn: image
lgtm!

jadebenn (approved):
lgtm!

aronwk-aaron (approved):

PR #1647, opened 2024-11-18
consolidate the messagetype enums into a single namespace

jadebenn:
Organization improvement: Move the messagetype enums into a subdirectory and namespace.


EmosewaMC (approved):
cool beans

PR #1650, opened 2024-11-18
fix: Remove database requirements for Property Entrance Component and greatly simplify logic

EmosewaMC:
Tested that the interface is identical in behavior to before
best friends get a special icon
alt accounts get a special icon
friends get a different special icon
arrows to go forward only show up if there are more properties to view
search looks in all 3 string fields
having an exact number of properties to fill a screen (12) does not show an arrow


aronwk-aaron (approved):

dCommon/dEnums/ePropertySortType.h
```diff
@@ -0,0 +1,11 @@
+#ifndef EPROPERTYSORTTYPE_H
+#define EPROPERTYSORTTYPE_H
```
jadebenn, line 2: I think you told me this earlier, but why aren't we just using `#pragma once` here?

EmosewaMC: doesn't work with pre-compiled headers on gcc

dGame/dComponents/PropertyEntranceComponent.h
```diff
@@ -59,10 +59,6 @@ class PropertyEntranceComponent final : public Component {
 	 */
 	[[nodiscard]] LWOMAPID GetMapID() const { return m_MapID; };
```
jadebenn, line 60:
```suggestion
	[[nodiscard]] LWOMAPID GetMapID() const noexcept { return m_MapID; };
```
the nitpickiest of nitpicks: a simple getter for a basic integer type can be noexcept

EmosewaMC: but what if

dGame/User.h
```diff
@@ -25,7 +25,7 @@ class User {
 	User& operator=(const User& other);
 	bool operator==(const User& other) const;
 
-	uint32_t GetAccountID() { return m_AccountID; }
+	uint32_t GetAccountID() const { return m_AccountID; }
```
jadebenn, line 28:
```suggestion
	uint32_t GetAccountID() const noexcept { return m_AccountID; }
```
ditto

jadebenn (approved):
nitpicks aside, built and ran and it looks good to me!

jadebenn (approved):

aronwk-aaron (approved):

PR #1651, opened 2024-11-21
feat: test using FetchContent for dependencies

jadebenn:
Playing around with CMake a bit. Not sure I'm going to go forward with this.


aronwk-aaron:
looks interesting

PR #1652, opened 2024-11-21
fix: Remove instances of undefined behavior

jadebenn:
Hooked up the sanitizer to try and track down any UB in our codebase

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,10 @@ project(Darkflame
 	LANGUAGES C CXX
 )
 
+#add_compile_options("-fsanitize=address,undefined")
+add_compile_options("-fsanitize=undefined" "-fvisibility=default")
+add_link_options("-fsanitize=undefined" "-static-libsan")
+
 # check if the path to the source directory contains a space
 if("${CMAKE_SOURCE_DIR}" MATCHES " ")
 	message(FATAL_ERROR "The server cannot build in the path (" ${CMAKE_SOURCE_DIR} ") because it contains a space. Please move the server to a path without spaces.")
@@ -17,7 +21,7 @@ set(CMAKE_C_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON) 	# Export the compile commands for debugging
 set(CMAKE_POLICY_DEFAULT_CMP0063 NEW) 	# Set CMAKE visibility policy to NEW on project and subprojects
-set(CMAKE_VISIBILITY_INLINES_HIDDEN ON) # Set C and C++ symbol visibility to hide inlined functions
+set(CMAKE_VISIBILITY_INLINES_HIDDEN OFF) # Set C and C++ symbol visibility to hide inlined functions
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 
 # Read variables from file
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,9 +4,10 @@ project(Darkflame
 	LANGUAGES C CXX
 )
 
-#add_compile_options("-fsanitize=address,undefined")
+# TEMP - Sanitizer flags
+# add_compile_options("-fsanitize=address,undefined" "-fvisibility=default")
 add_compile_options("-fsanitize=undefined" "-fvisibility=default")
-add_link_options("-fsanitize=undefined" "-static-libsan")
+add_link_options("-fsanitize=undefined" "-static-libasan")
 
 # check if the path to the source directory contains a space
 if("${CMAKE_SOURCE_DIR}" MATCHES " ")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,9 +5,17 @@ project(Darkflame
 )
 
 # TEMP - Sanitizer flags
+if (UNIX)
 # add_compile_options("-fsanitize=address,undefined" "-fvisibility=default")
-add_compile_options("-fsanitize=undefined" "-fvisibility=default")
-add_link_options("-fsanitize=undefined" "-static-libasan")
+	add_compile_options("-fsanitize=undefined" "-fvisibility=default")
+	add_link_options("-fsanitize=undefined")
+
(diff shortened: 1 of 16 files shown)


EmosewaMC:
make sure to check foot races in AG show the correct times at game end 👍

CMakeLists.txt
```diff
@@ -4,6 +4,19 @@ project(Darkflame
...
+	# add_compile_options("-fsanitize=address,undefined" "-fvisibility=default")
+	add_compile_options("-fsanitize=undefined" "-fvisibility=default")
+	add_link_options("-fsanitize=undefined")
+
+	if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+		add_link_options("-static-libsan")
+	else()
+		add_link_options("-static-libasan")
+	endif()
+endif()
```
EmosewaMC, line 18: we should keep these and lock them under a cmake variable flag

jadebenn: Agreed. the 'TEMP' bit is just a reminder to myself that I need to create a CMake config to enable/disable these

dCommon/Amf3.h
```diff
@@ -106,7 +94,7 @@ using AMFDoubleValue = AMFValue<double>;
  */
 class AMFArrayValue : public AMFBaseValue {
 	using AMFAssociative =
-		std::unordered_map<std::string, std::unique_ptr<AMFBaseValue>, GeneralUtils::transparent_string_hash, std::equal_to<>>;
+		std::unordered_map<std::string, std::unique_ptr<AMFBaseValue>, GeneralUtils::transparent_string_hash, std::equal_to<void>>;
```
EmosewaMC, line 97: lets get this in sooner rather than later in a separate PR because its breaking on gcc10, a supported compiler

dCommon/BrickByBrickFix.cpp
```diff
@@ -146,7 +146,7 @@ void WriteSd0Magic(char* input, uint32_t chunkSize) {
 	input[2] = '0';
 	input[3] = 0x01;
 	input[4] = 0xFF;
-	*reinterpret_cast<uint32_t*>(input + 5) = chunkSize; // Write the integer to the character array
+	std::memcpy(&input[5], &chunkSize, sizeof(uint32_t)); // Write the integer to the character array
```
EmosewaMC, line 149: this one is a bit harder to test, but ill send you a test case for it

dCommon/dClient/AssetManager.h
```diff
@@ -54,6 +54,7 @@ struct AssetStream : std::istream {
 	}
 
 	operator bool() {
+		// NEED TO FIX THIS
```
EmosewaMC, line 57: if this is changed make sure to check files existing and not existing with both packed and unpacked clients

dCommon/BrickByBrickFix.cpp
jadebenn: Unfortunately, this still isn't supported with gcc 10. I know I said earlier in the Discord, but I was mistaken. We either need to up the compiler requirement or strip out the offending code.

dCommon/Amf3.h
jadebenn: Unfortunately, this still isn't supported with gcc 10. I know I said earlier in the Discord, but I was mistaken. We either need to up the compiler requirement or strip out the offending code.

jadebenn:
> make sure to check foot races in AG show the correct times at game end 👍

Tested it, and it looks good.

dCommon/Amf3.h
EmosewaMC: I believe there are two ways we can go forward with this fix:
option A is to remove the transparent string hashing and use references to strings for comparisons (I dont think there is much copying if any going on when looking elements up and these strings are generally not big enough to cause large slowdowns?).
option B is to try and use some older stuff and make it compliant with gcc.

Why did you implement the string hasher initially? Does it make that much of a performance impact we'll notice at runtime?

jadebenn: I doubt it'll make a noticeable difference, but it was good practice and I thought it might come in useful for other unordered_maps.

One way we could tackle this is by gating it behind an #ifdef
```suggestion
#if defined(__cpp_lib_generic_unordered_lookup) && __cpp_lib_generic_unordered_lookup >= 201811L
		std::unordered_map<std::string, std::unique_ptr<AMFBaseValue>, GeneralUtils::transparent_string_hash, std::equal_to<void>>;
#else
		std::unordered_map<std::string, std::unique_ptr<AMFBaseValue>, std::equal_to<void>>;
#endif		
```

EmosewaMC: I'm not particularly a fan of changing a container definition based on the os you are on, however a solution is a solution.  If you're happy with this feel free to make a pr and we'll merge it :).  I do believe your set for the equal_to template is one argument early (the standard docs show it as argument 4 and not argument 3) so i would make sure this works.

jadebenn: I was doubly wrong. :(
The issue at hand is that std::unordered_map::find() was changed to use a template argument with C++20. For whatever reason, this particular feature of C++20 was not implemented in gcc 10. Thus, we can either rollback the change, or we can up the version requirement to gcc 11. Unfortunately, therre doesn't seem to be any other option.

EmosewaMC: I still leave it to your discretion of what to do.  which ever method you choose i'll support it.

jadebenn: Undoing the changes would require either changing all the string_views to strings or creating temporary strings at the call sites to handle the hashing. I think it makes more sense to just say we require versions of gcc > 10

EmosewaMC: gcc 11 is fine on my end.  again, im not concerned about copies for amf as these strings are generally small enough for small string optimization and speed is a far greater factor in other spots of the codebase.

jadebenn: It's not the creation of the strings that bothers me as much as the fact we'd need to put constructors in every call of find(). 😅

jadebenn: Doesn't seem to. If you try and compile this code on gcc 10, find complains about the argument being the wrong type. It doesn't seem to do any implicit conversion.

EmosewaMC: odd

PR #1653, opened 2024-11-23
fix: stack traces work again

EmosewaMC:
idk man stack traces work again
-rdynamic is === -export-dynamic 
the former is the compiler passing the latter to the linker, dont know why the order matters for this one but this works


aronwk-aaron (approved):
thanks emmo

jadebenn (approved):

PR #1656, opened 2024-11-27
fix: Old level files not loading

EmosewaMC:
tested that old level files correctly parse
tested that newer level files also still correctly parse


aronwk-aaron (approved):

jadebenn (approved):
lgtm

PR #1659, opened 2024-12-04
feat: Enable VS builds using Clang

jadebenn:
Add presets for using the Clang toolchain with VS


Jettford:
I took over this PR and managed to ruin the branch, it has moved to windows-clang-fixes

PR #1660, opened 2024-12-04
fix: abstract ugc rockets and cars from GameMessages

EmosewaMC:
Tested that trashing, placing for part swap, and crafting a new race car all insert and remove from the database the same as before

### DIFF
--- a/dGame/dGameMessages/GameMessages.cpp
+++ b/dGame/dGameMessages/GameMessages.cpp
@@ -5066,9 +5066,7 @@ void GameMessages::HandleModularBuildConvertModel(RakNet::BitStream& inStream, E
 
 	item->Disassemble(TEMP_MODELS);
 
-	std::unique_ptr<sql::PreparedStatement> stmt(Database::Get()->CreatePreppedStmt("DELETE FROM ugc_modular_build where ugc_id = ?"));
-	stmt->setUInt64(1, item->GetSubKey());
-	stmt->execute();
+	Database::Get()->DeleteUgcBuild(item->GetSubKey());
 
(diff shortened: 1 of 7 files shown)


dGame/dGameMessages/GameMessages.cpp
```diff
@@ -5394,6 +5392,8 @@ void GameMessages::HandleRemoveItemFromInventory(RakNet::BitStream& inStream, En
 		const auto itemType = static_cast<eItemType>(item->GetInfo().itemType);
 		if (itemType == eItemType::MODEL || itemType == eItemType::LOOT_MODEL) {
 			item->DisassembleModel(iStackCount);
+		} else if (itemType == eItemType::VEHICLE) {
+			Database::Get()->DeleteUgcBuild(item->GetSubKey());
```
Xiphoseer, line 5396: Since this is "RemoveItem", what happens if you put a modular build in a vault? Is `bConfirmed` only for delete?

Xiphoseer (approved):

dGame/dGameMessages/GameMessages.cpp
EmosewaMC: i will check, good question

aronwk-aaron (approved):

dGame/dGameMessages/GameMessages.cpp
EmosewaMC: given the line above this disassembles models into bricks, i dont think bConfirmed is used for vault transfers and this does destroy the item.  is this an apt conclusion?

EmosewaMC: this sends RequestMoveItemBetweenInventoryTypes and not this message.  no concerns in this case.

PR #1662, opened 2024-12-04
feat: re-write leaderboards again and fully remove mysql dependency outside of database

EmosewaMC:
to be merged after #1660 and #1661 

tested that all leaderboards show summaries in the same way as prior
tested that new scores update the values in the table or add one if one did not exist

the pages stuff was removed because the client has no functionality to actually request pages of the leaderboard.  it can put it in the GM but nothing sets it to anything aside from 0 and 10

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -312,7 +312,7 @@ add_subdirectory(dPhysics)
 add_subdirectory(dServer)
 
 # Create a list of common libraries shared between all binaries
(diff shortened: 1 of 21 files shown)


DarwinAnim8or (commented):
"remove mysql" is music to my ears

code also looks good so far, haven't run it locally though. 
but great job so far emo!

Xiphoseer (dismissed):

aronwk-aaron (dismissed):

aronwk-aaron (approved):

Xiphoseer (approved):

PR #1664, opened 2024-12-05
fix: use a default for build type if one is not specified

EmosewaMC:
tested that flags are set appropriately for the root folder if no cmake build type is specified, and that linux-gnu-release has its -03 flags set as per usual


aronwk-aaron (approved):

jadebenn:
I think this may cause problems with multiple configuration generators. Emphasis on 'may' - I need to investigate when I have a chance.

PR #1669, opened 2024-12-07
fix: use current binary directory mariadb shared object and dont override env variable

EmosewaMC:
we were overwriting the users environment variable with the build directory always, resulting in inis always being placed in the incorrect spot if a preset was specified, or if the user had an environment variable where they stored the inis.

Also, assuming it passes ci, made it so we use the shared library in the binary directory as opposed to the one

mac please

### DIFF
--- a/dDatabase/CMakeLists.txt
+++ b/dDatabase/CMakeLists.txt
@@ -2,6 +2,12 @@ add_subdirectory(CDClientDatabase)
 add_subdirectory(GameDatabase)
 
 add_library(dDatabase STATIC "MigrationRunner.cpp")
+
+add_custom_target(conncpp_dylib
+	${CMAKE_COMMAND} -E copy $<TARGET_FILE:MariaDB::ConnCpp> ${PROJECT_BINARY_DIR})
+
+add_dependencies(dDatabase conncpp_dylib)
+
 target_include_directories(dDatabase PUBLIC ".")
 target_link_libraries(dDatabase
 	PUBLIC dDatabaseCDClient dDatabaseGame)
--- a/tests/dCommonTests/CMakeLists.txt
+++ b/tests/dCommonTests/CMakeLists.txt
@@ -17,18 +17,13 @@ list(APPEND DCOMMONTEST_SOURCES ${DENUMS_TESTS})
 
 # Set our executable
 add_executable(dCommonTests ${DCOMMONTEST_SOURCES})
-add_dependencies(dCommonTests conncpp_tests)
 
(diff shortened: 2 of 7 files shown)


aronwk-aaron (approved):

Xiphoseer (approved):
Thank you!

dDatabase/CMakeLists.txt
```diff
@@ -2,6 +2,12 @@ add_subdirectory(CDClientDatabase)
 add_subdirectory(GameDatabase)
 
 add_library(dDatabase STATIC "MigrationRunner.cpp")
+
+add_custom_target(conncpp_dylib
+	${CMAKE_COMMAND} -E copy $<TARGET_FILE:MariaDB::ConnCpp> ${PROJECT_BINARY_DIR})
```
jadebenn, line 7: This is a bit of kludge. It's being placed in the correct directory for a multi-config generator. The problem is our scripts aren't set up to handle that correctly. We shouldn't need this.

tests/dCommonTests/CMakeLists.txt
```diff
@@ -17,18 +17,13 @@ list(APPEND DCOMMONTEST_SOURCES ${DENUMS_TESTS})
...
-add_custom_command(TARGET dCommonTests POST_BUILD
-	COMMAND otool ARGS -l dCommonTests
-	COMMAND otool ARGS -L dCommonTests
-	COMMAND ls
-	COMMAND otool ARGS -D libmariadbcpp.dylib
-	COMMAND install_name_tool ARGS -change libmariadbcpp.dylib @rpath/libmariadbcpp.dylib dCommonTests
-	COMMAND otool ARGS -L dCommonTests
-	WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
+	add_custom_target(dCommonTestsLink
+		${CMAKE_COMMAND} -E copy $<TARGET_FILE:MariaDB::ConnCpp> ${CMAKE_CURRENT_BINARY_DIR})
```
jadebenn, line 24: Ditto. This is a kludge.

jadebenn (commented):
I can't exactly do a 'changes required' since this has already been merged, but I think we need  to revisit this PR.

dDatabase/CMakeLists.txt
EmosewaMC: It is not being placed in the correct directory for milti-config generators as for mac CI and development, the file is requires to be in the linking directory and the binary directory at these times, and in neither was this true before.

PR #1670, opened 2024-12-08
fix: Make multi-config presets work again

jadebenn:
Reverted breaking change for multiple-configuration presets that was previously included in #1669

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -11,6 +11,9 @@
 			"displayName": "Default configure step",
 			"description": "Use 'build' dir and Unix makefiles",
 			"binaryDir": "${sourceDir}/build",
+			"environment": {
+				"DLU_CONFIG_DIR": "${sourceDir}/build"
+			},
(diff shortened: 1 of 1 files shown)


jadebenn:
On further review, I'm not sure #1669 should have been merged at all.

Jettford:
I believe this PR is counter-productive to changes I am trying to make on the origin/windows-clang branch. Not closing as I am awaiting further feedback.

Xiphoseer (approved):

Xiphoseer:
Fwiw: I'm generally in favor of not tying things down to a `build` directory, because it doesn't even have to be called that in cmake.

At the same time, this is our own env var which shouldn't confuse tooling too much, so if it restores stuff that was working it may be fine to merge even if replaced by a better way later.

EmosewaMC:
This makes it so CI no longer contains ini files.  I would rather prioritize CI and release over development environments.

jadebenn:
closing - there is a better approach

PR #1678, opened 2024-12-12
fix: implement enemy clear threat script

EmosewaMC:
this was supposed to be a small pr

tested that enemies that collide with this phantom collider tether to their spawn point and others tether as before.

### DIFF
--- a/dGame/dComponents/ControllablePhysicsComponent.cpp
+++ b/dGame/dComponents/ControllablePhysicsComponent.cpp
@@ -15,7 +15,7 @@
 #include "LevelProgressionComponent.h"
 #include "eStateChangeType.h"
 
-ControllablePhysicsComponent::ControllablePhysicsComponent(Entity* entity) : PhysicsComponent(entity) {
+ControllablePhysicsComponent::ControllablePhysicsComponent(Entity* entity, int32_t componentId) : PhysicsComponent(entity, componentId) {
 	m_Velocity = {};
 	m_AngularVelocity = {};
 	m_InJetpackMode = false;
--- a/dGame/Entity.cpp
+++ b/dGame/Entity.cpp
@@ -97,6 +97,8 @@
 #include "CDSkillBehaviorTable.h"
 #include "CDZoneTableTable.h"
 
+#include <ranges>
+
 Observable<Entity*, const PositionUpdate&> Entity::OnPlayerPositionUpdate;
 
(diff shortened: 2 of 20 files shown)


aronwk-aaron (approved):

jadebenn:
Can you explain the intent of the changes in a bit more detail?

EmosewaMC:
> Can you explain the intent of the changes in a bit more detail?

i wanted to implement the last 2 scripts for avant gardens.
the zone script is identical for ns and gf, so i added those too.
the if chain was getting ugly so i made it a set.
i added the enemy clear threat script.
this required implementing collision group on physics component.
this also required the component id making its way down to the base component class, editing every physics component constructor.
this is a base class with no identifier, so it needs to loop over the component list to find a physics component.
we dont have forced tether logic, so that had to be made.
this duplicated existing code so i made the common stuff a function.
the logic forces an object to tether for 3 seconds which is enough for an enemy to reset to their start position

dGame/dComponents/ControllablePhysicsComponent.cpp
```diff
@@ -15,7 +15,7 @@
 #include "LevelProgressionComponent.h"
 #include "eStateChangeType.h"
 
-ControllablePhysicsComponent::ControllablePhysicsComponent(Entity* entity) : PhysicsComponent(entity) {
+ControllablePhysicsComponent::ControllablePhysicsComponent(Entity* entity, int32_t componentId) : PhysicsComponent(entity, componentId) {
```
jadebenn, line 18: If this is a component ID, shouldn't it use the corresponding enum type?
```suggestion
ControllablePhysicsComponent::ControllablePhysicsComponent(Entity* entity, eReplicaComponentType componentId) : PhysicsComponent(entity, componentId) {
```

aronwk-aaron: It's the id of that component entry, not the type of component

jadebenn: What's the difference? Sorry if this is a dumb question.

EmosewaMC: component type is like the class type, component id is the row in the db to query

jadebenn (approved):
lgtm

aronwk-aaron (approved):

PR #1680, opened 2024-12-15
fix: warnings

EmosewaMC:
need to find something that uses that precondition

### DIFF
--- a/dGame/dUtilities/Preconditions.cpp
+++ b/dGame/dUtilities/Preconditions.cpp
@@ -137,7 +137,7 @@ bool Precondition::CheckValue(Entity* player, const uint32_t value, bool evaluat
 
 		return inventoryComponent->GetLotCount(value) >= count;
 	case PreconditionType::DoesNotHaveItem:
-		return inventoryComponent->IsEquipped(value) < count;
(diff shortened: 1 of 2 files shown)


EmosewaMC:
i cant test the precondition change because nothing uses it

dGame/dUtilities/Preconditions.cpp
```diff
@@ -137,7 +137,7 @@ bool Precondition::CheckValue(Entity* player, const uint32_t value, bool evaluat
 
 		return inventoryComponent->GetLotCount(value) >= count;
 	case PreconditionType::DoesNotHaveItem:
-		return inventoryComponent->IsEquipped(value) < count;
+		return inventoryComponent->IsEquipped(value) < static_cast<bool>(count);
```
jadebenn, line 140: The comparison being performed here is really unclear to me. Is it equivalent to this?
```suggestion
		return inventoryComponent->IsEquipped(value) && count > 0;
```
If so, maybe it should be rewritten like that to make the intent clearer.

EmosewaMC: sure

aronwk-aaron (approved):

jadebenn (approved):

PR #1679, opened 2024-12-13
fix: Implement missing survival tooltip script

EmosewaMC:
#1678 

Tetsed that a tooltip is now shown which has this message in it

PR #1681, opened 2024-12-15
feat: start car races if you "equip" the car near the car pad; add more old ns scripts to ignore list

EmosewaMC:
#1679 

Tested that dragging my car near the car pad and double clicking it in the inventory brings up the ui to start the race

### DIFF
--- a/dGame/dGameMessages/GameMessages.cpp
+++ b/dGame/dGameMessages/GameMessages.cpp
@@ -6338,6 +6338,40 @@ void GameMessages::SendUpdateInventoryUi(LWOOBJID objectId, const SystemAddress&
 
 	bitStream.Write(objectId);
 	bitStream.Write(MessageType::Game::UPDATE_INVENTORY_UI);
-	
+
+	SEND_PACKET;
+}
+
+void GameMessages::DisplayTooltip::Send() const {
+	CBITSTREAM;
+	CMSGHEADER;
+
+	bitStream.Write(target);
+	bitStream.Write(msgId);
+
+	bitStream.Write(doOrDie);
+	bitStream.Write(noRepeat);
+	bitStream.Write(noRevive);
+	bitStream.Write(isPropertyTooltip);
+	bitStream.Write(show);
+	bitStream.Write(translate);
+	bitStream.Write(time);
+	bitStream.Write<int32_t>(id.size());
+	bitStream.Write(id);
+
+	std::string toWrite;
+	for (const auto* item : localizeParams) {
+		toWrite += item->GetString() + "\n";
+	}
+	if (!toWrite.empty()) toWrite.pop_back();
+	bitStream.Write<int32_t>(toWrite.size());
+	bitStream.Write(GeneralUtils::ASCIIToUTF16(toWrite));
+	if (!toWrite.empty()) bitStream.Write<uint16_t>(0x00); // Null Terminator
+
+	bitStream.Write<int32_t>(imageName.size());
+	bitStream.Write(imageName);
+	bitStream.Write<int32_t>(text.size());
+	bitStream.Write(text);
+
 	SEND_PACKET;
 }
--- a/dGame/dGameMessages/GameMessages.cpp
+++ b/dGame/dGameMessages/GameMessages.cpp
@@ -6342,36 +6342,57 @@ void GameMessages::SendUpdateInventoryUi(LWOOBJID objectId, const SystemAddress&
 	SEND_PACKET;
 }
 
-void GameMessages::DisplayTooltip::Send() const {
-	CBITSTREAM;
-	CMSGHEADER;
+namespace GameMessages {
+	void GameMsg::Send(const SystemAddress& sysAddr) const {
+		CBITSTREAM;
+		CMSGHEADER;
 
-	bitStream.Write(target);
-	bitStream.Write(msgId);
+		bitStream.Write(target); // Who this message will be sent to on the (a) client
+		bitStream.Write(msgId); // the ID of this message
 
-	bitStream.Write(doOrDie);
-	bitStream.Write(noRepeat);
-	bitStream.Write(noRevive);
-	bitStream.Write(isPropertyTooltip);
-	bitStream.Write(show);
-	bitStream.Write(translate);
-	bitStream.Write(time);
-	bitStream.Write<int32_t>(id.size());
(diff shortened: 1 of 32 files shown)


EmosewaMC:
this is the PR to critique the game msg struct and the related components since i settled on a design here

aronwk-aaron (approved):

dGame/dGameMessages/GameMessages.cpp
```diff
@@ -6342,36 +6342,57 @@ void GameMessages::SendUpdateInventoryUi(LWOOBJID objectId, const SystemAddress&
...
+		bitStream.Write(translate);
+		bitStream.Write(time);
+		bitStream.Write<int32_t>(id.size());
+		bitStream.Write(id);
+
+		std::string toWrite;
+		for (const auto* item : localizeParams) {
+			toWrite += item->GetString() + "\n";
+		}
+		if (!toWrite.empty()) toWrite.pop_back();
```
jadebenn, line 6378: Is this being used to axe the null terminator?

dGame/dGameMessages/GameMessages.cpp
```diff
@@ -6342,36 +6342,57 @@ void GameMessages::SendUpdateInventoryUi(LWOOBJID objectId, const SystemAddress&
...
+	}
+
+	void UseItemOnClient::Serialize(RakNet::BitStream& bitStream) const {
+		bitStream.Write(itemLOT);
+		bitStream.Write(itemToUse);
+		bitStream.Write(itemType);
+		bitStream.Write(playerId);
+		bitStream.Write(targetPosition.x);
+		bitStream.Write(targetPosition.y);
+		bitStream.Write(targetPosition.z);
```
jadebenn, line 6396: Since you asked for some feedback on the design: Doesn't Raknet provide a way to serialize all the fields of a struct using one function call?

dGame/dGameMessages/GameMessages.h
```diff
@@ -705,7 +705,17 @@ namespace GameMessages {
...
+	};
+
+	struct UseItemOnClient : public GameMsg {
+		UseItemOnClient() : GameMsg(MessageType::Game::USE_ITEM_ON_CLIENT) {}
+		LWOOBJID playerId{};
+		LWOOBJID itemToUse{};
+		uint32_t itemType{};
+		LOT itemLOT{};
+		NiPoint3 targetPosition{};
+		void Serialize(RakNet::BitStream& bitStream) const override;
```
jadebenn, line 718: I like the idea of using structs for this a lot, especially if there's a way to make the process less verbose. 👍

dGame/dGameMessages/GameMessages.cpp
EmosewaMC: the new line

dGame/dGameMessages/GameMessages.cpp
EmosewaMC: it does allow this, however this relies on the structure never changing and we need it to match something in not our code, hence why I personally manually write each field of a struct (you can just call .Write(targetPosition) however this will include packing bytes, which we dont have and since we cant control the structure well, i'd rather not risk it and set a pattern that may be broken in some contexts).  Do you still think its worth it to write the position struct as is?

dGame/dGameMessages/GameMessages.h
EmosewaMC: SendSetStun(true, true, true, true, true, true, true, true, true, true, false, true, true, true, true, false);

dGame/dGameMessages/GameMessages.cpp
EmosewaMC: an alternative to make sure its correct while allowing the direct write of a struct would be to override this template in our bitstream utils like we do with string.  I could do that here if its wanted

dGame/dGameMessages/GameMessages.h
aronwk-aaron: can't we use . (dot) oporators with this as well?
So if we need....... the 5th bool in (EX) set to a non-deafult value, let's say `true` (idk what it is just as an example) we can just do `SetStun(.fifthvalue=true).send()` or something like that?

jadebenn: You could maybe create something with a similar effect, but that syntax wouldn't work

dGame/dGameMessages/GameMessages.cpp
jadebenn: What about a library to automatically deserialize the struct without packing bytes? Maybe something like this? https://github.com/p-ranav/alpaca

It would be so very nice to just pass a struct to a function and have it automatically be written as a bitstream.

dGame/dGameMessages/GameMessages.h
jadebenn: I.e. it might be possible to do something like
`SetStun.GetField<5>(true).send()`
but that's significantly less nice

dGame/dGameMessages/GameMessages.cpp
jadebenn: My understanding is the "behind the scenes" there is you're basically just using structured unbinding and template specialization to generate the appropriate function calls for each field

dGame/dGameMessages/GameMessages.h
EmosewaMC: im fairly certain it would also have to be `SetStun{.fifthvalue=true}.send()`.  i think it would work like that as is if you're setting a couple fields, itll be tomato tomato at the end of the day as the compiler will likely optimize it away at Release

jadebenn: That would only work if you were constructing a new value, right?

dGame/dGameMessages/GameMessages.cpp
EmosewaMC: Yes it would be nice, however alpaca supports from a glance _byte_ arrays, not bit arrays, which is what we would need here.

dGame/dGameMessages/GameMessages.h
EmosewaMC: i believe so, the whole thing would be in-line and would be destroyed at the end of the expression, which if you only need to change 1 field may be more clear, in some maybe not.

jadebenn (approved):
lgtm

PR #1688, opened 2024-12-18
remove unneeded Component.cpp, forward declare dependencies, and make Component definition header-only

jadebenn:
The Component.cpp file consists of only empty functions and isn't needed. Tinyxml2.h and raknet can also be forward-declared


aronwk-aaron (approved):

EmosewaMC:
imagine using labels couldnt be me

EmosewaMC (approved):

PR #1689, opened 2024-12-18
chore: switch cppscripts to references and unique_ptrs

jadebenn:


EmosewaMC:
i have some feedback points however mac ci beat me to them. I'd also like to suggest instead making the storage in g_Scripts the unique_ptr that way the only thing that needs to be changed is the body of GetScript with how the new'd script is stored (.reset on a unique_ptr) and how an existing one is returned (.get)

EmosewaMC:
alongside this, please make sure each and every script is actually _able_ to be deleted without issue and have its destructor called.  These have never had to have this happen and as such those code paths have not been tested ever.

EmosewaMC:
ignore my second feedback about removing all the make_unique, see #1682 for why.

PR #1692, opened 2024-12-24
fix: folder and file checks

EmosewaMC:
Tested that running MasterServer.exe from its own binary dir and 1 folder up create a resServer folder without issue and do not allow the program to continue if migrations, vanity or navmeshes are not found.


aronwk-aaron (dismissed):

aronwk-aaron (approved):

EmosewaMC:
I have also tested that the windows download behaves functionally identical to the built version, and shows the error box if folders dont exist the binary directory

DarwinAnim8or (approved):

PR #1694, opened 2024-12-24
fix: use encoding on latin1 strings from cdclient

EmosewaMC:
Tested that cafe corner no longer shows up with the incorrect text and now shows up as cafe corner.

@Xiphoseer said to "just hook up an iterator to the utf8 encoder" so I did just that by reading the string from binary as u8 (because if you try and convert a std::string it uses signed characters which utf8 did not like) and then run that through the exact same function that handled the utf16 to wtf8 conversion.

### DIFF
--- a/dCommon/GeneralUtils.cpp
+++ b/dCommon/GeneralUtils.cpp
@@ -167,17 +167,19 @@ std::u16string GeneralUtils::ASCIIToUTF16(const std::string_view string, const s
 	return ret;
 }
 
-//! Converts a (potentially-ill-formed) UTF-16 string to UTF-8
+
+//! Converts a (potentially-ill-formed) Latin1 string to UTF-8
 //! See: <http://simonsapin.github.io/wtf-8/#decoding-ill-formed-utf-16>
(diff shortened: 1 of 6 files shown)


aronwk-aaron (approved):

DarwinAnim8or (approved):

dCommon/GeneralUtils.cpp
```diff
@@ -194,6 +196,13 @@ std::string GeneralUtils::UTF16ToWTF8(const std::u16string_view string, const si
 
 	return ret;
 }
+std::string GeneralUtils::Latin1ToWTF8(const std::u8string_view string, const size_t size) {
+	return ToWTF8(string, size);
```
Xiphoseer, line 200: This is way too complicated... what I meant is just call `PushUtf8CodePoint` on all bytes, will get you a latin1 decoder. It needs none of the surrgate handling.

Xiphoseer (changes requested):

dCommon/GeneralUtils.cpp
EmosewaMC: #1695

PR #1695, opened 2024-12-25
fix: use simpler converter for latin1

EmosewaMC:
Still converts the cafe name

xipho is required to review this before it is merged.


Xiphoseer (approved):
Thank you!

jadebenn (approved):

PR #1699, opened 2024-12-28
fix: use correct res path for fdb check

### LINKED ISSUE #1698: BUG: Fdb checksum calculation gets stuck in infinite loop with packed client

### DarkflameServer Version

0c948a8d

### Issue Description

When check_fdb=0, the server behaves normally. When it is set to 1, a client attempting to connect will get stuck on the "sending username" step of the login process.

### Reproduction Steps

Have a MacBook Air with Apple Silicon and host a server with check_fdb=1

### Expected Behavior

Login attempts should either succeed or fail FDB authentication, not hang endlessly

### Environment

MacOS Sequoia 15

EmosewaMC:
fixes #1698 
hide whitespace changes

Tested that a packed client no longer gets stuck in an infinite loop trying to calculate the checksum.
Additionally we dont try to process the file unless we actually _found_ one now and exit if one was not found and the option was enabled.

### DIFF
--- a/dWorldServer/WorldServer.cpp
+++ b/dWorldServer/WorldServer.cpp
@@ -267,8 +267,6 @@ int main(int argc, char** argv) {
 
 	// pre calculate the FDB checksum
 	if (Game::config->GetValue("check_fdb") == "1") {
-		std::ifstream fileStream;
-
 		static const std::vector<std::string> aliases = {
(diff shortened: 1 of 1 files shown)


jadebenn (approved):
Thanks for fixing this

aronwk-aaron (approved):

PR #1700, opened 2024-12-28
fix: remove dead loop

EmosewaMC:
This loop is redundant since we're looking for the file in the client res directory

### DIFF
--- a/dWorldServer/WorldServer.cpp
+++ b/dWorldServer/WorldServer.cpp
@@ -267,36 +267,26 @@ int main(int argc, char** argv) {
 
 	// pre calculate the FDB checksum
 	if (Game::config->GetValue("check_fdb") == "1") {
-		static const std::vector<std::string> aliases = {
-			"CDServers.fdb",
-			"cdserver.fdb",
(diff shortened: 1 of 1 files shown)


jadebenn (approved):

dWorldServer/WorldServer.cpp
```diff
@@ -267,36 +267,26 @@ int main(int argc, char** argv) {
 
 	// pre calculate the FDB checksum
 	if (Game::config->GetValue("check_fdb") == "1") {
-		static const std::vector<std::string> aliases = {
```
aronwk-aaron, line 270: This removes looking via aliases?

EmosewaMC: cdserver.fdb is not used by anyone afaik and CDClient.fdb is covered by case insensitive searching of file paths.  I dont see a reason to keep the aliases anymore

aronwk-aaron: Makes sense

aronwk-aaron (approved):

PR #1707, opened 2025-01-01
fix: property entrance component optional

EmosewaMC:
Tested that the property entrance menu no longer crashes under windows debug


aronwk-aaron (approved):

jadebenn:
Can you explain what's happening here in a bit more detail?

EmosewaMC:
> Can you explain what's happening here in a bit more detail?

msvc crashes accessing an optional that has no value, since the default of an optional is nullopt in msvc, the assignment causes a crash if not initialized.  initializing to not nullopt fixes the issue

jadebenn (approved):

PR #1721, opened 2025-01-04
fix: flashing rebuilds not flashing and bricks not appearing from templates without destroyable components but with is_smashable set.

EmosewaMC:
Tested that rebuilds now correctly flash when getting close to smashing
Tested that lots that have is_smashable set, but dont have a destroyable component set get their isSmashable flag applied correctly.
Tested that SendDieNoImplCode is actually correct now

thanks dave

### DIFF
--- a/dGame/Entity.cpp
+++ b/dGame/Entity.cpp
@@ -386,6 +386,9 @@ void Entity::Initialize() {
 		if (m_Character) {
 			comp->LoadFromXml(m_Character->GetXMLDoc());
 		} else {
+			// extraInfo overrides. Client ORs the database smashable and the luz smashable.
+			comp->SetIsSmashable(comp->GetIsSmashable() | isSmashable);
+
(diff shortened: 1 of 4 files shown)


aronwk-aaron (approved):

dGame/Entity.cpp
```diff
@@ -1565,7 +1565,7 @@ void Entity::Kill(Entity* murderer, const eKillType killType) {
 
 	m_DieCallbacks.clear();
 
-	//OMAI WA MOU, SHINDERIU
+	//お前はもう死んでいる
```
jadebenn, line 1568: すべて計画通り

jadebenn (approved):
lgtm

dGame/Entity.cpp
EmosewaMC: やっぱりですよ

PR #1735, opened 2025-01-19
feat: Mail Re-write and packet/bitstream handler POC

aronwk-aaron:
using handlers and a new bitstream paradigm that we want to use moving forward
this is mostly for learning and fun since this is a small thing to re-write 
Tested all mail handlers, (de)serailize-s, etc
Everything works!


dGame/dUtilities/Mail.h
```diff
@@ -1,43 +1,217 @@
-#pragma once
+#ifndef __MAIL_H__
```
jadebenn, line 1:
```suggestion
#ifndef MAIL_H
```
Leading underscore defines are reserved

dNet/MailInfo.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+	VALIDATE_READ(bitStream.Read(itemID));
+	VALIDATE_READ(bitStream.Read(itemCount));
+	VALIDATE_READ(bitStream.Read(languageCode));
+	bitStream.IgnoreBytes(4); // padding
+
+	DluAssert(bitStream.GetNumberOfUnreadBits() == 0);
+
+	return true;
+}
```
jadebenn, line 63: No linebreak at EOF
```suggestion
}

```

dGame/dUtilities/Mail.h
aronwk-aaron: This is not consistent across the codebase, that should so be addressed in a separate PR if we want to enforce that

jadebenn: Fair enough

dNet/MailInfo.cpp
EmosewaMC, line 63:
```suggestion
}

```

EmosewaMC (approved):

DarwinAnim8or (approved):

PR #1737, opened 2025-01-20
feat: Entity message handling and a usage of said message handling

EmosewaMC:
Tested that the message is handled and has the expected result.  Tested that being gmlevel 0 rejected the message and that shooting gallery still tracks misses


Jettford (approved):
fuck, yeah

aronwk-aaron (approved):
seconded

PR #1739, opened 2025-01-20
feat: Add default client_location

EmosewaMC:
Since we by default expect the client to be here, may as well have it in the config by default to reduce setup steps

### DIFF
--- a/resources/sharedconfig.ini
+++ b/resources/sharedconfig.ini
@@ -24,7 +24,7 @@ dump_folder=
 
 # The location of the client
 # Either the folder with /res or with /client and /versions
-client_location=
+client_location=..
 
 # The maximum outgoing bandwidth in bits.  If your clients are having
 # issues with enemies taking a while to catch up to them, increse this value.


aronwk-aaron (approved):

jadebenn:
I have mixed feelings on this PR. What's the rationale behind putting the compiled server in the client directory, and why are we promoting that as the "standard" install?

aronwk-aaron:
> I have mixed feelings on this PR. What's the rationale behind putting the compiled server in the client directory, and why are we promoting that as the "standard" install?

It's simple and what most people will do

jadebenn:
Yeah, but by default a "proper" install of LU is going to be in the protected Program Files directory on Windows. While these days most people are not going to have "proper" installations, I'm still not sure it's something we should support.

aronwk-aaron:
Throwing the server in the client folder and hitting run had always been a goal on this project, especially since it makes single player instances self contained

EmosewaMC:
I did it like this because the other server emulators are also setup to run like this in a self contained folder in the client directory.  The current setup instructions already advise this so the goal was to simplify it to skip that step by default.  We dont exactly have an installer or a way of running the server and shutting it down when its not needed.  People are still able to and welcome to edit the sharedconfig.ini file with the client_location and we will continue to support that, however the goal of the default configuration is the least amount of steps possible to setup a server (which if this is implemented, is download the server, put the folder in the client directory, launch master twice and start the game).  I have added this back however it is listed as an optional step as I believe the default endorsed setup it putting it in the client directory so the server and client are in one place.

EmosewaMC:
After sleeping on it I believe doing an installer approach in Program Files is not the direction we are going for here.  If someone reinstalls windows, their whole save data will be wiped and finding/uninstalling the server will become much harder.  The people have always wanted a download and play with the fewest steps blocking them as possible, and dropping the server folder in the client folder with the default value of it being there reduces the steps by 2.  The only thing that I would maybe change here is where we say you should put the folder as I am fine with specifying instructions to put the folder 1 directory up from the client and then having them input the name of the folder in sharedconfig.ini.  Would this be acceptable?

aronwk-aaron:
would be neat at server startup if it can't find the client, then ask for it and it would save that to the setting. though that would be more complex

Xiphoseer:
Since the client folder should be called `client` anyway for reasons we've discussed elsewhere, which classically has `patcher` next to it:

1. I generally want to put that in a folder with a name of my choice anyway
2. that would naturally extend to a `server` with DLU executables, using `client_location=../client`

Xiphoseer:
Software defaulting to look outside of their install / working directory without really knowing what is there and outside a generic "walk up to find config" is a bit weird but I guess it'd work in this case

EmosewaMC:
> Since the client folder should be called `client` anyway for reasons we've discussed elsewhere, which classically has `patcher` next to it:
> 
> 1. I generally want to put that in a folder with a name of my choice anyway
> 2. that would naturally extend to a `server` with DLU executables, using `client_location=../client`

easier than editing a file because knowing how to rename something is required to use a computer

NealSpellman:
I'm personally ok with it in the client folder or a folder next to the client folder like @Xiphoseer's suggestion. I see what you were going for with just dropping it into the client folder though, as that is definitely something I've seen with some MMO repacks.

Definitely think it's right to put it _somewhere_ near the client like this.

jadebenn:
After further thought, I'm fine with @Xiphoseer's suggestion. I do like the idea of having a folder structure of 'client' 'server' 'patcher' etc.

aronwk-aaron (approved):

aronwk-aaron:
waiting on changes to be implemented before merging

PR #1740, opened 2025-01-20
chore: limit API to only listen on localhost

aronwk-aaron:


EmosewaMC (approved):

jadebenn (approved):
Don't quite understand but lgtm

EmosewaMC:
> Don't quite understand but lgtm

There is zero security around the webapi and i think its a good idea to not endorse setting up a non-secured api on a public server if someone fails to secure it on their own

PR #1751, opened 2025-03-17
fix: avery npc not having animations anymore

EmosewaMC:
Checked that the npc now jumps and teleports away instead of disappearing


aronwk-aaron (approved):

aronwk-aaron (approved):

codeshaunted (approved):
lgtm, thanks!

PR #1757, opened 2025-04-02
fix: pin actions to SHA commits and downgrade cmake to ~3.25

EmosewaMC:


EmosewaMC:
pinning to SHA commits makes it so if a bad actor commits to a release, our stuff wont be affected.  CMake 4.0 removes all backwards compatibility so i'm pinning us down to 3.x while our dependencies upgrade.

EmosewaMC:
マジ？？？？

Xiphoseer:
I'd just pin the non first-party actions, i.e. not the `actions/` ones

aronwk-aaron (approved):

aronwk-aaron (approved):

Xiphoseer (approved):

PR #1759, opened 2025-04-03
feat: Property behaviors partially functional

EmosewaMC:
Have tested that the following behaviors function:
On Interact
Drop Life Powerup
Drop Imagination Powerup
Drop Armor Powerup
Spawn Stroming
Spawn Pirate
Spawn Ronin
Play Sound
Smash
Wait
UnSmash

The model resets its state correctly when entering build mode and ceases playing further Actions
The model is not able to be interacted with if there are zero strips which have an OnInteract next
The model is able to be interacted with when one or more strips have OnInteract next
The model executes only its own actions and does not start other strips actions
The models behaviors do not interfere with each other and operate independently from one another

### DIFF
--- a/dGame/dPropertyBehaviors/Strip.cpp
+++ b/dGame/dPropertyBehaviors/Strip.cpp
@@ -7,6 +7,8 @@
 #include "ModelComponent.h"
 #include "PlayerManager.h"
 
+#include "DluAssert.h"
+
 template <>
 void Strip::HandleMsg(AddStripMessage& msg) {
 	m_Actions = msg.GetActionsToAdd();
@@ -105,7 +107,7 @@ void Strip::IncrementAction() {
 
 void Strip::Spawn(LOT lot, Entity& entity) {
 	EntityInfo info{};
-	info.lot = lot; // Dark Ronin property
+	info.lot = lot;
 	info.pos = entity.GetPosition();
 	info.rot = NiQuaternionConstant::IDENTITY;
 	info.spawnerID = entity.GetObjectID();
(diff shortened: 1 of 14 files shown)


dGame/dPropertyBehaviors/Strip.cpp
```diff
@@ -75,7 +80,138 @@ void Strip::HandleMsg(MigrateActionsMessage& msg) {
...
+	auto number = nextAction.GetValueParameterDouble();
+	auto numberAsInt = static_cast<int32_t>(number);
+	auto nextActionType = GetNextAction().GetType();
+	if (nextActionType == "SpawnStromling") {
+		Spawn(10495, entity); // Stromling property
+	} else if (nextActionType == "SpawnPirate") {
+		Spawn(10497, entity); // Maelstrom Pirate property
+	} else if (nextActionType == "SpawnRonin") {
+		Spawn(10498, entity); // Dark Ronin property
+	} else if (nextActionType == "DropImagination") {
```
aronwk-aaron, line 136: What about the commented out items in the block defs?

aronwk-aaron (approved):

dGame/dPropertyBehaviors/Strip.cpp
EmosewaMC: what about them?

aronwk-aaron: Iirc you said you were going to add them

EmosewaMC: will add them later

aronwk-aaron: sounds good

DarwinAnim8or: re: any traction on this?

aronwk-aaron: Later as in, another PR

PR #1762, opened 2025-04-11
fix: Add properties to landing anim excluded zones

EmosewaMC:
Check this video for footage of no animation playing on landing in a property. https://www.youtube.com/watch?v=FYqjZBnuBIg


aronwk-aaron (approved):

jadebenn:
Is there third person footage of someone entering a property? I'm wondering if the character might have just spawned in while the assets were being downloaded in the footage you shared.

PR #1771, opened 2025-04-17
feat: broadcast achievements in chat as in live

### LINKED ISSUE #1764: BUG: Achievement completion is not broadcast in Local Chat

- [x] I have checked that this issue has not already been reported.

- [x] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [x] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### DarkflameServer Version

current year

### Issue Description

Okay so. Had to comb through live footage for this one - following up for @EmosewaMC 

There is an individual locale entry for each unique achievement which is Extremely Funny
Image

But that's actually seemingly what they use to determine if an achievement should be broadcast or not. Maybe it's also defined somewhere in CDClient?

This video confirms two things: Achievements for others are posted in local chat (as seen for a half second), while achievements for others are **not** posted in team chat (as seen a half second earlier)

Image

Image
https://youtu.be/Yvou0wskQ7E?si=y36xbWR-piGbyCIt&t=5602

--

Here's where it gets fun. Right here, a player clearly levels up from smashing Butterscorch, but their achievement is not posted in chat. Mission 1754 (Smash Butterscorch!) does not have an accompanying MissionEmail entry.

Image
https://youtu.be/5G7ERUxNpOc?si=98DWmbM9pKcWUZW_&t=2645

--

We can confirm that your own achievements are not visible in your own chat. Here, the player earns both Mission 40 (A Creative Opportunity) and Mission 254 (Hidden Track). While 40 has a MissionEmail entry and 254 does not, neither are printed to chat.
https://youtu.be/mZk8w1Ru6FM?si=jv-xAzPILk9c_6LD&t=2374 

--

In conclusion, Achievements should be broadcast to local chat only, and only when they have an accompanying MissionEmail entry in locale. This actually goes against my memories, as I could have sworn they were posted in team chat but I guess not.

### Reproduction Steps

1. Have two players visit Red Blocks
2. Collect the Blue Imagination Brick with one player
3. Note that the other player does not see "%(PlayerName) just Completed 'A Creative Opportunity'!" in chat.
4. [Bingo bango bongo bish bash bosh](https://www.youtube.com/watch?v=Sq1oO8kxyuE)

### Expected Behavior

It should print the thing in chat

### Environment

EVGA RTX 9080 XTX (2GB) VII with dedicated Ageia PhysX card

EmosewaMC:
Tested that everyone on the receiving players' friends list receives the announcement as it went out in live.  Only works for achievements that have an entry in the MissionEmail table.  This may have been sent out to everyone in your zone as well however we don't really have a way to verify this aside from questioning why the client checks for the receiver being in the ignore list.  This is the only hint to me that this may have been broadcast to more than friends but again, no proof.

fixes #1764

### DIFF
--- a/dChatServer/ChatPacketHandler.cpp
+++ b/dChatServer/ChatPacketHandler.cpp
@@ -602,6 +602,19 @@ void ChatPacketHandler::HandleTeamInvite(Packet* packet) {
 	SendTeamInvite(other, player);
 
 	LOG("Got team invite: %llu -> %s", playerID, invitedPlayer.GetAsString().c_str());
+
+	bool failed = false;
+	for (const auto& ignore : other.ignoredPlayers) {
+		if (ignore.playerId == player.playerID) {
+			failed = true;
+			break;
+		}
+	}
+	
+	ChatPackets::TeamInviteInitialResponse response{};
+	response.inviteFailedToSend = failed;
+	response.playerName = invitedPlayer.string;
+	ChatPackets::SendRoutedMsg(response, playerID, player.worldServerSysAddr);
 }
 
 void ChatPacketHandler::HandleTeamInviteResponse(Packet* packet) {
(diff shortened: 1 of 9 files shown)


aronwk-aaron (dismissed):

EmosewaMC:
ignore the push meant for different pr.

Wincent01 (approved):

aronwk-aaron (approved):

PR #1777, opened 2025-04-21
chore: move all teams logic to its own namespace to consolidate logic

EmosewaMC:
I did not change any logic here, just moved function definitions and rename things accordingly.


EmosewaMC:
built this off of #1775, wait for that to merge before looking at this

DarwinAnim8or:
This seems interesting, but it appears it has some conflicts with the code now

EmosewaMC:
I manually checked that all function bodies that were moved matched what they were before (adjusted for namespace changes of course)

aronwk-aaron (approved):

PR #1794, opened 2025-05-07
feat: auto reject empty properties config option

### LINKED ISSUE #1752: ENH: Config option to auto reject empty properties

Players are currently able to submit empty properties for moderation. This creates clutter in the moderation queue.

An option in one of the .ini config files that, when enabled, the server will automatically reject a property that has submitted to moderation if has a performance cost of 0 (aka is empty). Should also send the typical rejection mail stating why the property was rejected (ie its empty).

_No response_

_No response_

### Additional context

_No response_

EmosewaMC:
Tested that having the config option set to 1 and having an empty property auto-rejected it.  Tested that having a model on the property or having the new config option set to 0 auto approved the property (as per live)
closes #1752

### DIFF
--- a/resources/worldconfig.ini
+++ b/resources/worldconfig.ini
@@ -77,3 +77,6 @@ allow_players_to_skip_cinematics=0
 # Customizable message for what to say when there is a cdclient fdb mismatch
 cdclient_mismatch_title=Version out of date
 cdclient_mismatch_message=We detected that your client is out of date. Please update your client to the latest version.
+
+# Auto reject properties which contain no models | must be 1 in order to auto reject.
+auto_reject_empty_properties=0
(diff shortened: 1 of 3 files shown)


resources/worldconfig.ini
```diff
@@ -77,3 +77,6 @@ allow_players_to_skip_cinematics=0
 # Customizable message for what to say when there is a cdclient fdb mismatch
 cdclient_mismatch_title=Version out of date
 cdclient_mismatch_message=We detected that your client is out of date. Please update your client to the latest version.
+
+# Auto reject properties which contain no models | must be 1 in order to auto reject.
+auto_reject_empty_properties=0
```
aronwk-aaron, line 82: Imo, default this to 1

aronwk-aaron (approved):
Nit, non-blocking.

resources/worldconfig.ini
EmosewaMC: I set this to 0 to mimic live behavior since if the server communications to the client were implemented properly, the client would auto-display the property as approved and not bother displaying a moderation status whether or not it would have had one.

PR #1799, opened 2025-05-16
refactor: Convert WorldPacket and ClientPackets to use LUBitStreams

aronwk-aaron:
doing both at once since they are swapped and not all that large


dNet/ClientPackets.cpp
```diff
@@ -5,119 +5,136 @@
...
+		for (auto it : rejectedWords) {
+			bitStream.Write<uint8_t>(it.first); // start index
+			bitStream.Write<uint8_t>(it.second); // length
 		}
+
+		// Pad out the rest of the packet
+		// The client expects 64 items, so we need to write 64 - rejectedWords.size() empty items
+		for (int i = rejectedWords.size(); 64 > i; i++) {
+			bitStream.Write<uint16_t>(0);
+		}
```
EmosewaMC: The first loop should make sure only 64 items are being written in this case, writing more will not be read or acknowledged.  If we're going to be making some minor logic changes let's get a few of them that are minor while we're here

dNet/ClientPackets.cpp
```diff
@@ -5,119 +5,136 @@
...
-	return language;
+	void HTTPMonitorInfoResponse::Serialize(RakNet::BitStream &bitStream) const {
+		bitStream.Write<uint16_t>(port);
+		bitStream.Write<uint8_t>(openWeb);
+		bitStream.Write<uint8_t>(supportsSum);
+		bitStream.Write<uint8_t>(supportsDetail);
+		bitStream.Write<uint8_t>(supportsWho);
+		bitStream.Write<uint8_t>(supportsObjects);
+	}
 }
```
EmosewaMC, line 127: This file matches the previous bitstream logic

dNet/ClientPackets.h
```diff
@@ -8,29 +8,122 @@
...
+
+	struct ServerState : public LUBitStream {
+		bool serverReady = false;
+
+		ServerState() : LUBitStream(eConnectionType::CLIENT, MessageType::Client::SERVER_STATES) {};
+		void Serialize(RakNet::BitStream& bitStream) const override;
+	};
+
+	struct CreateCharacter : public LUBitStream {
+		LWOOBJID objid = 0;
```
EmosewaMC, line 77:
```suggestion
		LWOOBJID objid = LWOOBJID_EMPTY;
```

dNet/WorldPackets.cpp
```diff
@@ -1,186 +1,192 @@
...
+	}
 
-void WorldPackets::SendHTTPMonitorInfo(const SystemAddress& sysAddr, const HTTPMonitorInfo& info) {
-	CBITSTREAM;
-	BitStreamUtils::WriteHeader(bitStream, eConnectionType::CLIENT, MessageType::Client::HTTP_MONITOR_INFO_RESPONSE);
-	info.Serialize(bitStream);
-	SEND_PACKET;
-}
+	void StringCheck::Handle() {
+	}
```
EmosewaMC, line 190: Forget something?

dNet/WorldPackets.h
```diff
@@ -2,40 +2,75 @@
...
+		struct LocalSpaceInfo {
+			LWOOBJID objectId = LWOOBJID_EMPTY;
+			NiPoint3 position = NiPoint3Constant::ZERO;
+			NiPoint3 linearVelocity = NiPoint3Constant::ZERO;
+		};
+		LocalSpaceInfo localSpaceInfo;
+		struct RemoteInputInfo {
+			bool operator==(const RemoteInputInfo& other) {
+				return m_RemoteInputX == other.m_RemoteInputX && m_RemoteInputY == other.m_RemoteInputY && m_IsPowersliding == other.m_IsPowersliding && m_IsModified == other.m_IsModified;
+			}
```
EmosewaMC, line 51:
```suggestion
			bool operator==(const RemoteInputInfo&) = default;
```

dNet/ClientPackets.h
```diff
@@ -8,29 +8,122 @@
...
+		LOT templateID = 1;
+		std::u16string name;
+		eGameMasterLevel gmLevel = eGameMasterLevel::CIVILIAN;
+		int32_t chatMode = 0;
+		std::string xmlData;
```
EmosewaMC, line 82: This string is often times massive.  Let's make this a string_view instead.

dWorldServer/WorldServer.cpp
```diff
@@ -1040,8 +1037,17 @@ void HandlePacket(Packet* packet) {
...
 				if (!characterComponent) return;
 
-				WorldPackets::SendCreateCharacter(packet->systemAddress, player->GetComponent<CharacterComponent>()->GetReputation(), player->GetObjectID(), c->GetXMLData(), username, c->GetGMLevel());
-				WorldPackets::SendServerState(packet->systemAddress);
+				ClientPackets::CreateCharacter createCharacter;
+				createCharacter.objid = player->GetObjectID();
+				createCharacter.name = username;
+				createCharacter.gmLevel = c->GetGMLevel();
+				createCharacter.xmlData = c->GetXMLData();
+				createCharacter.reputation = player->GetComponent<CharacterComponent>()->GetReputation();
```
EmosewaMC, line 1046:
```suggestion
				createCharacter.reputation = characterComponent->GetReputation();
```
toctou

dWorldServer/WorldServer.cpp
```diff
@@ -1216,16 +1222,14 @@ void HandlePacket(Packet* packet) {
...
+		positionUpdate.Deserialize(inStream);
 		User* user = UserManager::Instance()->GetUser(packet->systemAddress);
 		if (!user) {
 			LOG("Unable to get user to parse position update");
 			return;
 		}
-
-		Entity* entity = Game::entityManager->GetEntity(user->GetLastUsedChar()->GetObjectID());
-		if (entity) entity->ProcessPositionUpdate(positionUpdate);
+		positionUpdate.objectID = user->GetLastUsedChar()->GetObjectID();
```
EmosewaMC, line 1233: Handle is not called

EmosewaMC (changes requested):
Please make sure all packets have a test in-game alongside a read test to check for no missed logic

dNet/WorldPackets.cpp
aronwk-aaron: d;ksmdlkfmsdl

dNet/WorldPackets.h
aronwk-aaron: Does this just work then? or does it need to be defined somewhere else?

EmosewaMC: this works as is - =default means "use a compiler generated == comparison" which will just compare all members for equality with other

dNet/ClientPackets.cpp
```diff
@@ -91,14 +89,16 @@ namespace ClientPackets {
...
+			if (toWrite <= 0) return;
 			bitStream.Write<uint8_t>(it.first); // start index
 			bitStream.Write<uint8_t>(it.second); // length
+			toWrite--;
 		}
-
-		// Pad out the rest of the packet
-		// The client expects 64 items, so we need to write 64 - rejectedWords.size() empty items
-		for (int i = rejectedWords.size(); 64 > i; i++) {
+		for (toWrite; toWrite <= 0; toWrite--) {
```
EmosewaMC, line 101: you dont need the `toWrite;`.  a `;` is sufficient, just evaluates to an empty expression

dNet/WorldPackets.h
aronwk-aaron: image
It does not like that

aronwk-aaron: (dicord) It's const so that' doesn't work

PR #1800, opened 2025-05-16
fix: playing an emote not showing on all clients

### LINKED ISSUE #1342: BUG: Messages shown by emotes are only shown client-side

### DarkflameServer Version

e239166

### Issue Description

When playing an emote, the animation is shown to other players, but emotes that play messages do not show any messages to other players.

### Reproduction Steps

1. Get two players
2. Have one play an emote that should show a message
3. Observe the other player cannot see any message

### Expected Behavior

Other players should see a message above other players' heads when they play an emote that displays one.

### Environment

WSL

ElectScholar:
(Fix #1342)

Thanks to @EmosewaMC 's suggestions,

I Created new structure for EmotePlayed with the `Game::MessageType::EMOTEPLAYED` that already exists in `DGame.h`
With the targetID for whom the emote is directed to and the emoteID for what type of emote is played which client handles with locale.xml (I believe atleast).

Then a serialize function was made in the GameMessage class to take advantage of boiler plate headers for serializing and sending.

implementation requires identifying the target through a 'LWOOBJID' of entity that started emote, the emoteID for emote type, and targetID if there is a target given.

For [PetComponent::Command] I forced targetID as 0 since I don't yet know if there are any emotes directed to other entities or if they are all just broadcast. If wanted I can keep it to be Owner entity.(https://github.com/DarkflameUniverse/DarkflameServer/blob/main/dGame/dComponents/PetComponent.cpp#L944)

### DIFF
--- a/dGame/dGameMessages/GameMessages.cpp
+++ b/dGame/dGameMessages/GameMessages.cpp
@@ -4958,7 +4958,7 @@ void GameMessages::HandleQuickBuildCancel(RakNet::BitStream& inStream, Entity* e
 }
 
 void GameMessages::HandlePlayEmote(RakNet::BitStream& inStream, Entity* entity) {
-	int emoteID;
+	int32_t emoteID;
 	LWOOBJID targetID;
 
 	inStream.Read(emoteID);
@@ -4979,7 +4979,7 @@ void GameMessages::HandlePlayEmote(RakNet::BitStream& inStream, Entity* entity) 
 
 	GameMessages::EmotePlayed msg;
 	msg.target = entity->GetObjectID();
-	msg.emoteID = static_cast<int32_t>(emoteID);
+	msg.emoteID = emoteID;
 	msg.targetID = targetID;      // The emote’s target entity or 0 if none
 	msg.Send(UNASSIGNED_SYSTEM_ADDRESS);  // Broadcast to all clients
 
--- a/dGame/dGameMessages/GameMessages.cpp
+++ b/dGame/dGameMessages/GameMessages.cpp
@@ -4977,7 +4977,6 @@ void GameMessages::HandlePlayEmote(RakNet::BitStream& inStream, Entity* entity) 
 		if (emote) sAnimationName = emote->animationName;
 	}
 
-	//RenderComponent::PlayAnimation(entity, sAnimationName);
 	GameMessages::EmotePlayed msg;
 	msg.target = entity->GetObjectID();
 	msg.emoteID = static_cast<int32_t>(emoteID);
--- a/dGame/dGameMessages/GameMessages.cpp
+++ b/dGame/dGameMessages/GameMessages.cpp
@@ -4977,7 +4977,12 @@ void GameMessages::HandlePlayEmote(RakNet::BitStream& inStream, Entity* entity) 
 		if (emote) sAnimationName = emote->animationName;
 	}
 
-	RenderComponent::PlayAnimation(entity, sAnimationName);
+	//RenderComponent::PlayAnimation(entity, sAnimationName);
+	GameMessages::EmotePlayed msg;
+	msg.target = entity->GetObjectID();
+	msg.emoteID = static_cast<int32_t>(emoteID);
+	msg.targetID = targetID;      // The emote’s target entity or 0 if none
+	msg.Send(UNASSIGNED_SYSTEM_ADDRESS);  // Broadcast to all clients
 
 	MissionComponent* missionComponent = entity->GetComponent<MissionComponent>();
 	if (!missionComponent) return;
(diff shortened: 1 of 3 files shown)


dGame/dGameMessages/GameMessages.cpp
```diff
@@ -4977,7 +4977,12 @@ void GameMessages::HandlePlayEmote(RakNet::BitStream& inStream, Entity* entity)
 		if (emote) sAnimationName = emote->animationName;
 	}
 
-	RenderComponent::PlayAnimation(entity, sAnimationName);
+	//RenderComponent::PlayAnimation(entity, sAnimationName);
```
EmosewaMC, line 4980:
```suggestion
```

dGame/dGameMessages/GameMessages.cpp
```diff
@@ -4977,7 +4977,12 @@ void GameMessages::HandlePlayEmote(RakNet::BitStream& inStream, Entity* entity)
...
+	GameMessages::EmotePlayed msg;
+	msg.target = entity->GetObjectID();
+	msg.emoteID = static_cast<int32_t>(emoteID);
```
EmosewaMC, line 4983: instead of a static cast here, the `int` type of emoteID should be changed out with `int32_t`.  on most platforms `int` does refer to a 32 bit int, however theres always the odd chance we find a platform with 16 bit ints.  For DLU's case since we are a cross platform server for a windows compiled game, we always use specific bit widths for bitstreams.  

As an example, heres a godbolt snippet of why we do this https://godbolt.org/z/svosssbvf

dGame/dGameMessages/GameMessages.h
```diff
@@ -833,6 +833,15 @@ namespace GameMessages {
 	struct ResetModelToDefaults : public GameMsg {
 		ResetModelToDefaults() : GameMsg(MessageType::Game::RESET_MODEL_TO_DEFAULTS) {}
 	};
-};
 
+	struct EmotePlayed : public GameMsg {
+		EmotePlayed() : GameMsg(MessageType::Game::EMOTE_PLAYED), emoteID(emoteID), targetID(targetID) {}
```
EmosewaMC, line 838:
```suggestion
		EmotePlayed() : GameMsg(MessageType::Game::EMOTE_PLAYED) {}
```

EmosewaMC (changes requested):

EmosewaMC:
I'll check what pets used as a target in their emotes

dGame/dGameMessages/GameMessages.h
ElectScholar: Would we prefer to default ID's to a set value for later debugging purposes like emoteID = 175 ('Well Done!' emote) and a targetID of zero for general broadcast? Or do you think we should just remove it all together like you originally suggested?

ElectScholar:
I have committed most of your suggestions and would like to just check on my build if removing the PlayAnimation function would work out because I do remember during my testing that PlayAnimation did successfully run the animation and don't know yet if EmotePlayed applies to animation as well as the text bubble.

ElectScholar:
> I have committed most of your suggestions and would like to just check on my build if removing the PlayAnimation function would work out because I do remember during my testing that PlayAnimation did successfully run the animation and don't know yet if EmotePlayed applies to animation as well as the text bubble.

Ignore this

dGame/dGameMessages/GameMessages.h
EmosewaMC: zero initialized is fine.  we want to make sure the value is zero so if someone doesnt set a field, undefined behavior does not occour, but we don't need to provide a default aside from zero here

ElectScholar: Commit made!

ElectScholar:
All desired changes have been made :sunglasses:

EmosewaMC:
> All desired changes have been made 😎

waiting on CI

dGame/dGameMessages/GameMessages.h
```diff
@@ -833,6 +833,15 @@ namespace GameMessages {
 	struct ResetModelToDefaults : public GameMsg {
 		ResetModelToDefaults() : GameMsg(MessageType::Game::RESET_MODEL_TO_DEFAULTS) {}
 	};
-};
 
+	struct EmotePlayed : public GameMsg {
+		EmotePlayed() : GameMsg(MessageType::Game::EMOTE_PLAYED), emoteID(0), targetID(0) {}
```
EmosewaMC, line 838: In the future don't chain constructors for a zero initialization, instead prefer using `{}` on the member itself like int32_t emoteID{};

EmosewaMC (approved):

PR #1807, opened 2025-06-07
Added spawn management commands, along with some minor bugfixes

NinjaOfLU:
Added commands to print to chat, despawn, and save to a bug reports entry, lists of entities, based on entity groups, or who/what spawned them.

Added a command to despawn an object by object id, and one to despawn based on the 'nearest to player' check done by the inspect command.

Added a command to print a player's object id to chat.

Also replaced (splitArgs.empty()) checks with (args == ""), because the argument parser never returns an empty vector, only a vector containing an empty string.

Updated Commands.md to reflect the added commands.

### DIFF
--- a/dGame/EntityManager.cpp
+++ b/dGame/EntityManager.cpp
@@ -271,6 +271,18 @@ std::vector<Entity*> EntityManager::GetEntitiesInGroup(const std::string& group)
 	return entitiesInGroup;
 }
 
+std::vector<Entity*> EntityManager::GetEntitiesBySpawnerID(const LWOOBJID& objectID) const {
+	std::vector<Entity*> entitiesSpawnedBy;
+	if (!GetEntity(objectID)) return entitiesSpawnedBy;
+	for (auto* entity : m_Entities | std::views::values) {
+		if (!(entity->GetSpawnerID() == objectID)) continue;
+
+		entitiesSpawnedBy.push_back(entity);
+	}
+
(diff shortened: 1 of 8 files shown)


EmosewaMC:
will review today sometime.  Please revert the changes to ConstructEntity in the meantime as that is old code now.

EmosewaMC:
This is quite a large change list, if it could be broken up at all that would be nice but if not its fine

NinjaOfLU:
> > Also replaced (splitArgs.empty()) checks with (args == ""), because the argument parser never returns an empty vector, only a vector containing an empty string.
> 
> instead lets fix the issue at the root and not push back into the vector if the input is empty

Works for me! I was kind of just scared of touching it and breaking more things. I guess I should have checked where else it was used, but I kind of ran with 'maybe it's like that for a reason; I don't want to touch more than I need to'. All in all, sounds good!

EmosewaMC (changes requested):
On second thought, this needs to be split up into multiple prs, theres a lot going on here and it would be easier to follow whats being added in smaller chunks.  The saving code also makes no sense being stored/read from where it does and should be re-tooled to either have its own table or to write to disk.  there are a lot of changes i'd like to suggest however i want to put that time in in smaller chunks.  
i would think this should be 3-4 prs in total, one modifying any existing functions, and 2-3 more adding more commands (ideally 3).

PR #1821, opened 2025-06-22
feat: change behavior ID to LWOOBJID length

EmosewaMC:
Not sure why I didn't start the behavior ID as this.  Cherry-picked from a larger feature PR, so i'll do testing in the morning to check that all prior features work as expected and that migrations apply well.

### DIFF
--- a/dGame/dComponents/ModelComponent.cpp
+++ b/dGame/dComponents/ModelComponent.cpp
@@ -67,7 +67,7 @@ void ModelComponent::LoadBehaviors() {
 	for (const auto& behavior : behaviors) {
 		if (behavior.empty()) continue;
 
-		const auto behaviorId = GeneralUtils::TryParse<int32_t>(behavior);
+		const auto behaviorId = GeneralUtils::TryParse<LWOOBJID>(behavior);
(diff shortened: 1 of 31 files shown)


dGame/dComponents/ModelComponent.cpp
```diff
@@ -116,8 +116,11 @@ void ModelComponent::Serialize(RakNet::BitStream& outBitStream, bool bIsInitialU
 	if (bIsInitialUpdate) outBitStream.Write0(); // We are not writing model editing info
 }
 
-void ModelComponent::UpdatePendingBehaviorId(const int32_t newId) {
-	for (auto& behavior : m_Behaviors) if (behavior.GetBehaviorId() == -1) behavior.SetBehaviorId(newId);
+void ModelComponent::UpdatePendingBehaviorId(const LWOOBJID newId, const LWOOBJID oldId) {
+	for (auto& behavior : m_Behaviors) {
+		if (behavior.GetBehaviorId() != oldId) continue;
+		behavior.SetBehaviorId(newId);
+	}
```
EmosewaMC, line 123: This is the only functional change in this pr. From the more recent one and the corresponding call changes are at the top of ControlBehaviors.cpp

PR #1855, opened 2025-08-17
Consolidate serviceID enums into one enum

jadebenn:
There are currently 3 different enum types to represent the ID of a service, whereas only one is really necessary. This PR consolidates them down into 1.


dMasterServer/MasterServer.cpp
```diff
@@ -601,26 +601,26 @@ void HandlePacket(Packet* packet) {
 			inStream.Read(theirServerType);
 			inStream.Read(theirIP);
 
-			if (theirServerType == ServerType::World) {
+			switch (theirServerType) {
```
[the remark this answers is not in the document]
EmosewaMC: I think a break would be fine

dNet/AuthPackets.cpp
```diff
@@ -64,24 +65,22 @@ void AuthPackets::HandleHandshake(dServer* server, Packet* packet) {
...
 	const uint32_t clientNetVersion = GeneralUtils::TryParse<uint32_t>(clientNetVersionString).value_or(171022);
 
 	bitStream.Write<uint32_t>(clientNetVersion);
 	bitStream.Write<uint32_t>(861228100);
-
-	if (serverType == ServerType::Auth) bitStream.Write(ServiceId::Auth);
-	else if (serverType == ServerType::World) bitStream.Write(ServiceId::World);
-	else bitStream.Write(ServiceId::General);
+	bitStream.Write(serverType);
+    bitStream.Write<uint16_t>(0); // padding
```
EmosewaMC, line 83:
```suggestion
```

EmosewaMC, line 83:
```suggestion
	bitStream.Write(static_cast<uint32_t>(serverType));
```
makes more sense than a secondary padding (client also does refer to this as a uint and not a short). This does differ from the original code though in that now General won't be replied if serverType is not Auth or World, though im not sure those servers even handle this packet.

tests/dGameTests/dGameMessagesTests/GameMessageTests.cpp
EmosewaMC, line 1: The test changes should be able to be reverted (except the naming, good to fix those up)

dNet/AuthPackets.cpp
```diff
@@ -51,9 +51,10 @@ void AuthPackets::HandleHandshake(dServer* server, Packet* packet) {
 	inStream.Read(clientVersion);
 	inStream.IgnoreBytes(4);
 
-	ServiceId serviceId;
-	inStream.Read(serviceId);
-	if (serviceId != ServiceId::Client) LOG("WARNING: Service ID is not a Client!");
+	ServiceType serviceType;
+	inStream.Read(serviceType);
+	if (serviceType != ServiceType::CLIENT) LOG("WARNING: Service is not a Client!");
+    inStream.IgnoreBytes(2);
```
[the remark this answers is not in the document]
jadebenn: Nope, looks like a mistake to me. Good catch. 👍

dNet/AuthPackets.cpp
jadebenn: Change looks good to me, and I've committed it. Any worries about the potential logic/behavior change? I've playtested the server and haven't noticed any ill effects.

EmosewaMC (approved):
Thanks again for this!

PR #1856, opened 2025-08-21
Revert uint8_t in game message tests

jadebenn:
This is a remnant of a reverted change that slipped into a previous PR and no longer makes sense since ServiceType is a uint16_t, not a uint8_t


jadebenn:
Ah, crap. The git history is borked. This needs to look like it was forked from main, not the original commit branch. I'll fix it later, I guess...

PR #1877, opened 2025-09-14
feat: lxfml splitting for bbb 

aronwk-aaron:
TODO: write up comprehensive doc for it


dCommon/Lxfml.cpp
```diff
@@ -128,3 +152,258 @@ Lxfml::Result Lxfml::NormalizePosition(const std::string_view data, const NiPoin
...
+		} else if (const tinyxml2::XMLComment* c = child->ToComment()) {
+			auto* n = dstDoc.NewComment(c->Value());
+			dst->InsertEndChild(n);
+		}
+	}
+
+	return dst;
+}
+
+std::vector<Lxfml::Result> Lxfml::Split(const std::string_view data, const NiPoint3& curPosition) {
```
EmosewaMC, line 182: How well does this function handle bad lxfml data?  it would be trivial to send a payload with bad data to the server, please check for error handling with either malformed xml, malformed _lxfml_ (i.e. an invalid lxfml format) and lxfml with extra data in it.  We know the client will likely only send good lxfml, but we can never guarantee user input is good.

aronwk-aaron:
> Please ensure there is no issue with lots of splitting (try getting as many splits as you can from a vanilla player) to ensure stability.

good call, will test

dCommon/Lxfml.cpp
aronwk-aaron: It did not handle it well, neither did parts of the normalization function, I have addressed issues in both functions, added tests for theses cases, and re-tested everything in-game

aronwk-aaron:
> Please ensure there is no issue with lots of splitting (try getting as many splits as you can from a vanilla player) to ensure stability.

there was no server issue, it did take a hot second to split 15k bricks into their own models and insert them into the DB, but the server worked as expected. The client, however, did not like loading that many models and crashed lol, but it did attempt to load them in until it gave up

dCommon/Lxfml.cpp
```diff
@@ -32,10 +32,18 @@ namespace {
...
 Lxfml::Result Lxfml::NormalizePosition(const std::string_view data, const NiPoint3& curPosition) {
 	Result toReturn;
+	
+	// Handle empty or invalid input
+	if (data.empty()) {
+		return toReturn;
+	}
+	
+	// Ensure null-terminated string for tinyxml2::Parse
+	std::string nullTerminatedData(data);
```
EmosewaMC, line 42: Please explore other options of checking this.  This is an expensive copy.

EmosewaMC (changes requested):
I've run the tests locally and it creates a bunch of extra files after a test is done.  Please re-write the test logic to properly copy over the text file to the build folder.  Writing to another file should also not be necessary at all.  The file should also be read from the build folder, not from the source tree.

Issue #1870, opened 2025-09-07
ENH: Enhance Mute functionality


HailStorm32:
### Is your feature request related to a problem?

Current Mute only restricts mail and chat. The muted player could still interact with players via trade and still submit names. 

### Describe the solution you'd like

Add two new config options. One to enable trade to be restricted on a mute, and the other to restrict the ability to submit names (both pet and character).

Names submitted by the muted account would be auto rejected, and the needs_rename column set to 0 (so that the  user can't resubmit).

### Repository breaking implications

None

PR #1887, opened 2025-09-28
Add configurable restrictions for muted accounts

HailStorm32:
Addresses https://github.com/DarkflameUniverse/DarkflameServer/issues/1870 plus extra

Adds three new config options that change what an account mute does.

### mute_restrict_mail
Enables/disables mail being restricted on an account with an active mute

### mute_restrict_trade
Enables/disables trade being restricted on an account with an active mute

### mute_auto_reject_names
Enables/disables an account with an active mute account from giving custom names

-  Pets
     - Pet name will default to `Pet` and will be automatically approved 
- Characters
     - If a character is created with a custom name, custom name will be ignored and will use the predefined name instead
     - If a character is forced to rename, any name given will be ignored, and a random predefined name will be generated and used instead
     
 Tested that functionality works as intended.


EmosewaMC (approved):

EmosewaMC:
- [ ] ci

PR #1894, opened 2025-10-04
fix: live accurate player flag missions and flag debugging 

EmosewaMC:
Tested that the client reflects the correct server progression after a test map and manually setting a flag off.
tested that session flags correctly pick up on progression updates


aronwk-aaron (approved):
I witnessed it working

aronwk-aaron (approved):

PR #1917, opened 2025-10-19
feat: banana banana banana

EmosewaMC:
the banana drops from the banana tree and then the banana rolls to a banana destination on the banana client
when banana is hit, banana will drop the loot at the base of the banana tree for some reason
dont know why live bananas did this but i hacked it in so banana can do banana things and banana roll


Terrev:
sounds like a plan-tain : )

PR #1920, opened 2025-10-22
feat:  proper gminvs with ghosting

aronwk-aaron:
tested that this works

there is an issue with loading in as invisible between the server and the client.
You will be invisible, but your character will not be transparent due to some obscure load ordering that we have not figured out yet

### DIFF
--- a/dChatServer/TeamContainer.cpp
+++ b/dChatServer/TeamContainer.cpp
@@ -477,7 +477,7 @@ TeamData* TeamContainer::CreateLocalTeam(std::vector<LWOOBJID> members) {
 		}
 	}
 
-	newTeam->lootFlag = 1;
+	newTeam->lootFlag = 0;
 
 	TeamStatusUpdate(newTeam);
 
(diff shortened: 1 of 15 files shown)


dGame/dComponents/GhostComponent.cpp
```diff
@@ -61,6 +86,34 @@ void GhostComponent::GhostEntity(LWOOBJID id) {
...
+			controllableComp->SetDirtyPosition(true);
+		}
+	}
+	Game::entityManager->SerializeEntity(m_Parent);
+
+	return true;
+}
+
+bool GhostComponent::OnGetGMInvis(GameMessages::GameMsg& msg) {
+	LOG("GM Invisibility requested: %s", m_IsGMInvisible ? "true" : "false");
```
EmosewaMC:
```suggestion
	LOG_DEBUG("GM Invisibility requested: %s", m_IsGMInvisible ? "true" : "false");
```

EmosewaMC (approved):
pending ci

PR #1933, opened 2025-11-16
fix: chest loot not working

EmosewaMC:
tested that team and solo loot now function correctly


aronwk-aaron (approved):

EmosewaMC:
DropActivityLoot does the team logic internally

Issue #1941, opened 2025-11-19
BUG: Undefined Behavior due to Achievement Propagation


dawnreprime:
### Issue Description
Due to the way the achievement system is coded there exists a possibility that undefined behavior modifies the tracking incorrectly and adds additional progress to future meta achievements (probably the best-case scenario) when completing the one before it.

This is due to https://github.com/DarkflameUniverse/DarkflameServer/blob/a713216540f833b6e2a2dbeccb8aa40da1e9d86a/dGame/dComponents/MissionComponent.cpp#L144 being recursive in a depth-first manner alongside the fact that https://github.com/DarkflameUniverse/DarkflameServer/blob/a713216540f833b6e2a2dbeccb8aa40da1e9d86a/dGame/dComponents/MissionComponent.h#L180 is being modified whilst being iterated over during the tree execution.

This (can) produce an execution as follows:
1. mission _A_ completes meta mission _M_
2. meta mission _N_ gets 1 from meta mission _M_'s completion
3. meta mission _N_ gets 1 from mission _A_'s completion

leading to the extra mission contributing to a tier it should not.

In C++, it’s typically not a great idea to mutate over certain collections having iterator invalidation without taking precautions. Even if being careful about **unordered_map**’s specific cases about never calling **erase()** (specific element invalidated) and even adding a band-aid **reserve()** to avoid rehashes on **insert()** (complete iterator invalidation). Also, the behavior can differ depending on standard library implementations, binaries, and even different runs.

### Reproduction Steps
- If you are blessed enough to test your software on a single OS (GCC + Linux/Docker), the bug can be seen happening when unlocking the final tier of meta achievements.

- If you are running on the MSVC Windows version, you’ll find that it happens on every subsequent meta achievement.

I really wish Clang Windows worked with this project… but compilers don’t fix UB!

### Expected Behavior
Upon progressing to the next meta achievement, the value should start at 1/N instead of 2/N (Ensure to not trigger multiple achievements on the last step).

### Environment
- GCC Linux/Docker, MSVC Windows

- SQLite or MariaDB

### Workaround
This is what I came up with to stop UB happening during the recursive tree execution:
```cpp
#include <utility>
```
https://github.com/DarkflameUniverse/DarkflameServer/blob/a713216540f833b6e2a2dbeccb8aa40da1e9d86a/dGame/dComponents/MissionComponent.cpp#L144-L150
[...]

PR #1943, opened 2025-11-24
Fix Achievement Propagation Undefined Behavior

dawnreprime:
Address primary issues in #1941:
- Fix undefined behavior during achievement progress algorithm
- Avoid iterator invalidation if unordered_map inserts lead to rehash

Other changes:
- Leverage mission list iteration for functions iterating all missions
- Fix const-correctness in various places
- Add missing delete in MissionComponent


dawnreprime:
### Extra Notes:
The removal case is not triggered by the code section.

### Testing:
Newly added meta achievements no longer gain incorrectly tracked progress on accepting, including edge cases before the next rank with multiple triggered achievements cascading.

EmosewaMC:
I have already resolved this issue on a local branch in a more efficient manner.

Issue #1926, opened 2025-11-12
BUG: scriptedactivity loot apparently using LOT instead of specified activity


Terrev:
### Make sure you've done the following:

- [x] I have checked that this issue has not already been reported.

- [x] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [x] I have downloaded/pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

- [x] I have verified that my boot.cfg is configured as per the [README](https://github.com/DarkflameUniverse/DarkflameServer?tab=readme-ov-file#allowing-a-user-to-connect-to-your-server).

### DarkflameServer Version

headache

### Issue Description

https://explorer.lu-dev.net/objects/12199/39

aura mar flowers currently reward loot as expected because their LOT and activity ID are the same (12199)

other flowers in the game currently do not reward loot when they should

generic flower found throughout NS, pet cove, etc:
https://explorer.lu-dev.net/objects/3646/39
is LOT 3646 with activity id 102 (NS Watered Flower)

frostburgh flower
https://explorer.lu-dev.net/objects/12302/39
is LOT 12302 with activity id 109 (Winter Watered Flower)

footage of a non-AM flower rolling its loot as defined by its scriptedactivity component https://youtu.be/JRM5RNI_yXg?list=PL684D6C6B270136D1&t=295

i am too tired to find vanilla footage of me watering a NS flower but the FB one is basically the same deal anyway

### Reproduction Steps

water ns or frostburgh flower

### Expected Behavior

imagination powerup drops 100% of time and rare chance of faction token

### Environment

swimming in the small container of arbys bronco berry sauce on my desk

PR #1950, opened 2025-12-21
fix: flowers using wrong/nonexistent activities

Terrev:
fixes www dot github dot com slash DarkflameUniverse slash DarkflameServer slash issues slash 1926

yeah i tested it all behaves as expected


aronwk-aaron:
fixes #1926

aronwk-aaron (approved):

PR #1990, opened 2026-06-10
fix: handling of the same skill on multiple items

### LINKED ISSUE #1339: BUG: server doesn't allow same skill in multiple slots

### DarkflameServer Version

ours. thats all i got rn

### Issue Description

if two items equipped in different slots share a skill, you only get the skill in one slot upon equipping both at once

https://explorer.lu-dev.net/objects/12601 Padded Toque of Healing
https://explorer.lu-dev.net/objects/12742 Medic Shield
https://explorer.lu-dev.net/objects/12733 Heavy Staff (via subitem 13804 Heavy Staff Left Hand Proxy)
all use skill 1006 Healing Pulse
but equipping the toque (headgear) at the same time as either the shield or staff (hand items) will not give you the skill of the second item equipped. things also get funky upon unequipping/re-equipping depending on the exact order you do things in

another example:
https://explorer.lu-dev.net/objects/13582 Mole Whacker (via subitem 13934 Molewhacker Left Hand Proxy)
https://explorer.lu-dev.net/objects/5995 Speedy Cap
both use skill 335 Speed Boost and conflict with each other in the same way

image

90% sure this is a server quirk rather than a client limitation given these are all vanilla items, but lu works in mysterious ways

### Reproduction Steps

equip 12601 Padded Toque of Healing and 12742 Medic Shield at the same time, look at your skills

### Expected Behavior

you have skill 1006 Healing Pulse in both slots 2 and 3 at the same time and things also behave properly when unequipping and re-equipping the items in various orders and combinations

### Environment

floor

oh and this is in fact untested on mainline dlu cause i cant rn and its 3 am and im not jumping through hoops to verify that at this hour lol goodnight

### LINKED ISSUE #1557: BUG: server does not accurately track skill references for added skills

### DarkflameServer Version

All

### Issue Description

currently the server allows the same skill to be equipped in two different slots when in live how this would work is the skill would only be in the first slot that it was equipped to, and then if all items that used that skill were unequipped, or if an equipping item overrides the original skills slot, the other slot would assume the skill.

### Reproduction Steps

equip these two items
https://explorer.lu-dev.net/objects/13582 Mole Whacker (via subitem 13934 Molewhacker Left Hand Proxy)
https://explorer.lu-dev.net/objects/5995 Speedy Cap
observe that unequipping the items removes the skill from the wrong slot instead of keeping it there or moving it if another items takes over that slot.

### Expected Behavior

2 items with the same skill:
- first equipped item slot is where the skill will reside.  unequipping either item will not remove the skill but instead decrement a reference counter by 1.
- equipping another item will remove the skill as usual, but only by decrementing a reference counter.  the item will do either of the following based on what happened
  - a new item overrides the slot 2 skills are in means the other slot will take ownership and add skill will need to be sent again for the new slot.
  - a new item overriding the slot not owned by the skill but owned by a reference will result in the new item skill being added and the reference counter going down by 1.
  - 

### Environment

日本語

aronwk-aaron:
fixes #1339 
fixes #1557


dGame/dComponents/InventoryComponent.cpp
```diff
@@ -1327,23 +1337,17 @@ void InventoryComponent::RemoveDatabasePet(LWOOBJID id) {
...
-	switch (type) {
-	case eItemType::HAT:
-		return BehaviorSlot::Head;
-	case eItemType::NECK:
-		return BehaviorSlot::Neck;
-	case eItemType::LEFT_HAND:
-		return BehaviorSlot::Offhand;
-	case eItemType::RIGHT_HAND:
-		return BehaviorSlot::Primary;
-	case eItemType::CONSUMABLE:
```
aronwk-aaron, line 1335: check this

aronwk-aaron: checked, works correctly and saves correctly

dGame/dComponents/InventoryComponent.cpp
```diff
@@ -723,10 +723,6 @@ void InventoryComponent::Serialize(RakNet::BitStream& outBitStream, const bool b
 		for (const auto& pair : m_Equipped) {
 			const auto item = pair.second;
 
-			if (bIsInitialUpdate) {
-				AddItemSkills(item.lot);
-			}
-
```
EmosewaMC, line 729: why is this removed?

EmosewaMC (approved):
approved aside from the one question

dGame/dComponents/InventoryComponent.cpp
aronwk-aaron: that AddSkill message is sent during entity construction, but the client's SkillComponent isn't initialized yet at that point — so the packets get dropped silently. In the PLAYER_LOADED handler in GameMessageHandler.cpp, where ClearSkills() is called first and then items are equiped with AddItemSkills. That's when the client is actually ready to receive them, from what I could tell from packet caps and RE, so basically, it didn't make sens to have them here and nothing really changes since it would have been really doing anything in the first place

PR #2011, opened 2026-06-21
feat: add the lightning orb script*

EmosewaMC:
Doesn't do anything because MPC don't work...

fixed a bug where multiple of the same name could be pushed into a dense debug array


aronwk-aaron (approved):
fu

PR #2025, opened 2026-08-24
fix(scripts): resolve player lock-in and skill execution for Personal Fortress (Issue #1609)

### LINKED ISSUE #1609: BUG: Getting stuck using faction skills

### DarkflameServer Version

v2.3.0

### Issue Description

Using the Assembly Defense Fortress for Rank 3 Engineer can cause the player to get stuck inside it, which forces the player to use the `/die` command to get unstuck. The turret's sweeping animation continues to play, but no damage is dealt. This has been noted at least on Forbidden Valley and Crux Prime, but likely affects all worlds. 

This also occurs with the wormholer mech.

### Reproduction Steps

Use a faction skill kit for the first time on a world. The two specific ones that have been pointed out are the Wormholer mech not despawning properly and assembly defense fortress locking the player inside.

### Expected Behavior

Wormholer mech should despawn, player should not get stuck in assembly fortress, wormholer mech should despawn

### Environment

Windows

ghost:
Fixes #1609

### Problem
When casting the Assembly Defense Fortress (Rank 3 Engineer faction kit, skill 649 / LOT 9479), the player could become permanently locked inside the fortress unable to move/attack, requiring \/die\ to escape. Furthermore, the fortress turret sweeping skill 650 did not consistently target or cast with the player as the originator.

### Root Cause
1. \PersonalFortress::OnStartup\ and \OnDie\ lacked fallback parent entity checks when \self->GetOwner()\ resolved to \self\ or null.
2. If the spawned fortress entity was destroyed or cleaned up without triggering a traditional \Kill\ sequence, \OnDie\ was bypassed, leaving the player's stun and status immunities indefinitely pushed on the client and server.
3. \OnTimerDone\ used \CalculateBehavior\ directly with magic parameters and no originator target, preventing proper skill routing.

### Solution
- Added safe owner and parent resolution in \PersonalFortress::OnStartup\ and \PersonalFortress::OnDie\.
- Registered an \AddDieCallback\ on the fortress entity to ensure the player's stun and status immunities are always popped upon entity destruction regardless of the death path.
- Updated \OnTimerDone\ to invoke \CastSkill\ with the originator correctly set to the owner entity.


ghost:
Closing for local testing and quality verification.

Issue #596, opened 2022-06-22
BUG: Crux Prime Computer does not respawn if it is killed by an Area of Effect attack


EmosewaMC:
### Make sure you've done the following:

- [X] I have checked that this issue has not already been reported.

- [X] I have validated that this issue is not a syntax error of either MySQL or SQLite.

- [X] I have pulled the latest version of the main branch of DarkflameServer and have confirmed that the issue exists there.

### DarkflameServer Version

c89f420

### Issue Description

The computer on Crux Prime does not respawn if it is killed with an indirect or direct Area of Effect attack.

### Reproduction Steps

1) Build the crux prime computer all the way until the end.
2) Smash it with the Doom hammer (easy repro) or the space ranger offhand (need an enemy nearby to actually use the ability)
3) Observe the computer does not respawn.

### Expected Behavior

Computer respawns and is unable to be destroyed.

### Environment

All

aronwk-aaron:
Possibly fixed, need to test

PR #2029, opened 2026-08-24
fix(scripts): prevent null mission dereference and out-of-range crash in dropship computer (Issue #596)

ghost:
### Overview
Resolves Issue #596 by preventing a fatal `nullptr` dereference segfault when players interact with the dropship computer without mission 979, and guarding against an uncaught `std::out_of_range` string slicing exception in `OnDie`.

---

### Root Cause & Code Analysis

1. **Unchecked `GetMission(979)` Pointer Dereference:**
   ```cpp
   if (inventoryComponent->GetLotCount(m_NexusTalonDataCard) != 0 || 
       missionComponent->GetMission(979)->GetMissionState() == eMissionState::COMPLETE) {
       return;
   }
   ```
   * `MissionComponent::GetMission(missionId)` returns `nullptr` whenever the given mission is not currently accepted or completed by the character.
   * Interacting with the dropship computer without mission 979 dereferences `nullptr->GetMissionState()`, causing an immediate server crash (`SIGSEGV`).
   * **Fix:** Safely check `mission979 != nullptr` before inspecting its state.

2. **String Slicing Out-of-Bounds:**
   * `myGroup.substr(10, 1)` unconditionally indexed offset 10 of `spawner_name`.
   * If `spawner_name` has fewer than 11 characters, `std::string::substr` throws an uncaught `std::out_of_range` exception.
   * **Fix:** Guard with `if (myGroup.length() < 11) return;`.

---

### Verification
- Tested simulated player interactions with:
  1. No mission 979 active/accepted $\to$ safely passes without null-pointer exception.
  2. Mission 979 active but incomplete $\to$ adds datacard as intended.
  3. Mission 979 complete $\to$ returns without re-adding datacard.
  4. Empty / malformed `spawner_name` $\to$ safely returns without throwing `std::out_of_range`.


ghost:
Closing to conduct full local simulation, in-game verification, and gather QA evidence before resubmission.

PR #2034, opened 2026-09-10
feat: Add cheat detection knowledge

EmosewaMC:
tested that getting kicked correctly displays the reasons why and their corresponding values
tested that a user with developer permissions is not kicked for said funness
tested that logs are correct for those funness values which i could actually trigger


aronwk-aaron (dismissed):

aronwk-aaron:
isn't there a cheat detection table that this should be pushed to? or did we get rid of that?

aronwk-aaron (approved):
lgtm otherwise

EmosewaMC:
> isn't there a cheat detection table that this should be pushed to? or did we get rid of that?

we still have that, i'll have a wack at adding that as well

